PR #14, opened 2014-07-09
WIP: Parsing infastructure

dcjones:
Not ready for review yet, but follow along and comment if interested.

Notable features:
  * Parser are generated with ragel (specifically, [my fork of ragel](https://github.com/dcjones/ragel)).
  * `ragel.jl` defines some types and macros to factor out the tricky parts: which is mainly parsing a buffered stream as a opposed to reading everything into memory before parsing, which we obviously want to avoid.
  * Regular files are memory mapped by default. Potentially that can lead to speed up in some cases, but I need to do some benchmarking to see.

You can see the example FASTA parser I have in `fasta.rl`. It's pretty simple to define once you wrap your head around ragel, and pretty fast (about 9 seconds to parse the human genome from a fasta file with 40% cpu utilization, indicating that it's mostly I/O bound). All it does now is print all the sequence identifiers it finds, so it's not yet useful.

Next up:
  1. [ ] Define some sequence types. (In a separate branch/PR)
  2. [ ] Make the fasta parser actually fully functional.
  3. [ ] Revaluate the design to see if there's anything else that can be factored out into macros, etc.
  4. [ ] More parsers!
  5. [ ] Document how to build parsers.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.3-
 Compat
-FactCheck
+Switch
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,4 @@
 julia 0.3-
 Compat
+Docile
 Switch
--- a/src/Bio.jl
+++ b/src/Bio.jl
@@ -1,6 +1,6 @@
 module Bio
 
+include("ragel.jl")
 include("seq/seq.jl")
-
 
 end # module
--- a/src/ragel.jl
+++ b/src/ragel.jl
@@ -1,0 +1,360 @@
+
+module Ragel
+
+using Switch
+import Base: push!, pop!, endof, append!, empty!, isempty, length, getindex,
+             setindex!, (==)
+
+
+# A simple buffer type, similar to IOBuffer but faster and with fewer features.
+# It's essentially just a vector that can grow but is never allowed to
+# reallocate to smaller size.
+
+const INITIAL_BUF_SIZE = 1000000
+
+type Buffer{T}
+    data::Vector{T}
+    pos::Int
+
+    function Buffer(initial_size=10000)
+        new(Array(T, initial_size), 1)
+    end
+end
+
+
+function isempty(buf::Buffer)
+    return buf.pos == 1
+end
+
+
+function length(buf::Buffer)
+    return buf.pos - 1
+end
+
+
+function empty!(buf::Buffer)
+    buf.pos = 1
+end
+
+
+function getindex(buf::Buffer, i::Integer)
+    if !(1 <= i < buf.pos)
+        throw(BoundsError())
+    end
+    @inbounds x = buf.data[i]
+    return x
+end
+
+
+function setindex!{T}(buf::Buffer{T}, value::T, i::Integer)
+    if !(1 <= i < buf.pos)
+        throw(BoundsError())
+    end
+    @inbounds buf.data[i] = value
+end
+
+
+function ensureroom!(buf::Buffer, n::Int)
+    if buf.pos + n > length(buf.data)
+        resize!(buf.data, 2 * length(buf.data))
+    end
+end
+
+
+function push!{T}(buf::Buffer{T}, c::T)
+    ensureroom!(buf, 1)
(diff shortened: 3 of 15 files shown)


dcjones:
We have a complete FASTQ parser now with all 51 test cases passing. :fireworks:

The last thing I want to do before merging this is improve the error messages. Right now it will just say "error parsing" if the input is bad, without saying what or where the problem is.

Ward9250:
Yaaay! :dancers: Good work! I've been trying to push forward with Phylo too, so it is as complete as I can get it before June - not sure if I can go to JuliaCon yet but it's something I'd love to do.

dcjones:
That's great! I'm thinking we should merge this branch and yours soon, then put this package in METADATA so it can be installed easily. Let me know when you feel phylo is relatively feature complete and I'll be happy to review and merge it.

dcjones:
I'm ready to merge this. There's still the question of documentation, but nothing is documented now, so that's going to be my next priority, and it'll be easier for me to tackle in one fell swoop in another PR.

This is a difficult PR to review, since it's partially written in ragel state machine specification. Most import is to not look at `fasta.jl` or `fastq.jl`. Those are generated by ragel from `fasta.rl` and `fastq.rl`, respectively, and will melt your face if you try to read them.

If you'd like, you can checkout this branch and kick the tires. The main interface it provides is a `read` function that works like: `read(open(filename), FASTA)` or `read(open(filename), FASTQ)` and returns an iterator over sequences.

Ward9250:
I see so you open a file handle and pass it to read() with a type
specification as the second argument for the correct method to be
dispatched?

Ward9250:
And presumably to just get all the sequences use collect(), the iteration
is a nice feature for keeping only what is required in memory in doing
operations on individual sequences.

dcjones:
> I see so you open a file handle and pass it to read() with a type
specification as the second argument for the correct method to be
dispatched?

Right, and I didn't mention but you can also just pass a filename like `read(filename, FASTA)`.

> And presumably to just get all the sequences use collect(), the iteration
is a nice feature for keeping only what is required in memory in doing
operations on individual sequences.

Eventually we may even want to define `read!(seq, input)` functions that update `seq` in-place and don't allocate anything. Especially for FASTQ and SAM/BAM this will be important if we want to be as fast as possible. Doing that will require making sequences mutable though, which will require a little work.

dcjones:
I want to merge this soon, but it's going to tank our coverage statistics. The use of macros in this code utterly perplexes coverage measurement, so there are thousands of lines that are executed but marked as irrelevant, and a bunch of blank lines and comments that get counted at no-coverage. :confused:

This may not be ideal, but I'll try to figure out a way for coveralls to ignore the generated parsers.

dcjones:
On the topic of testing: the tests now just check that there are no errors an well formed input, and are errors on malformed input. So it's not checking that what's output by the parser, its just checking (quite vigorously) that its definition of FASTA and FASTQ is correct.

I'm not sure the best way to go about testing the output from the parser. I could write separate hand-written parsers and check that the results are the same, but that's a lot of effort and hard to maintain. I could also check it against a parser in another language, e.g. parse the file in biopython, serialize it to json, read that in julia and check that it matches. That would also be complicated to automate though.

Blahah:
@dcjones I think we should eventually expand BioFmtSpecimens to allow result tests as well, e.g. by including metadata from the FASTA sequence headers in `index.yaml`, plus some stats like the sequence length, defline length, etc. Perhaps the simplest way to do this is to use some other languages to produce the metadata and populate `index.yaml`, then we can check that the parsed objects in Julia match what we expect.

For now though, let's just review it and merge ASAP. I'll have a run through this afternoon.

dcjones:
>  I think we should eventually expand BioFmtSpecimens to allow result tests as well

That's a good idea. The files aren't huge so I think we could just accompany each test case with the same data serialized to JSON or YAML. If we can read the data in two formats and have them agree, then it's probably correct. Those files should certainly be populated by something other than Bio.jl. Who's parsers do we trust the most?

PR #17, opened 2014-07-25
Phylo Module

Ward9250:
A lot to add but I'll add to the branch piecemeal from my other Phylogenetics repo to make sure things are acceptable bit by bit. To begin with, we have the basic types and methods for building phylogenies with Phylo. It was important to go back and think harder about this than in the previous version where phylogenies were heavily borrowed from ape and R. These types and methods after all are going to be building blocks for the input / output and creation / manipulation of trees as well as all manner of other possible methods.

The structure of a phylogeny is now as follows:

A Phylogeny type contains a name, whether it is rooted and reportable, and the root. The root is a reference to an instance of PhyNode. A PhyNode has a name, a field for branch length (length of the edge between it and it's parent). An array for extensions / annotations, an array of references to it's children (other PhyNodes), and a reference to it's parent - if it is a single node created on it's own or is the root of a tree or subtree, this parent field point to itself i.e. the node is self referential at that field.

We have several basic methods for getting and setting information, as well as several functions that mania the references to parents and children to edit the structure of the tree. However, currently only two are exported - prune! and graft! - these are the "safe" tree manipulators. You can graft! a subtree or single child to another node of a tree, or prune! and node off a tree, returning the node and any other children it has - yielding a subtree cut away from the rest of the tree. The graft! methods sure that when grafting the parent and child references are made properly, and the prune! method likewise makes sure these links are broken properly so there are no situations where a parent refers to a child, yet the child doesn't refer back to that parent and so on.

Let's say you wanted to build a basic tree out of some nodes in a Julia session (tree building algorithms may well have to do this) Well make a basic cladogram without branch lengths.

a = PhyNode("A.")
b = PhyNode("B.")
c = PhyNode("C.")
d = PhyNode("D.")
e = PhyNode("E.")

Let's say you want a, b and c to be tips and e the root and d an internal node connecting a and b.
We can build the structure like so.
graft!(e, d)
graft!(e, c)
graft!(d, a)
graft!(d, b)

Calling any of a, b, c, d, e. in the console allows you to see the structure.

So we have our PhyNodes linked together, now we can finish the tree.

myTree = Phylogeny("My Simple Tree", e, true, false)

We can also cal methods on the nodes:

isLeaf(a)
isLeaf(b)
isLeaf(c)
isLeaf(d)
isLeaf(e)

isRoot(a)
isRoot(b)
isRoot(c)
isRoot(d)
isRoot(e)

isNode(a)
isNode(b)
isNode(c)
isNode(d)
isNode(e)

Hope you like! Comments of course welcome!


Ward9250:
A heads up guys, I'm thinking of changing how annotations to nodes of trees is achieved, because currently, if you have some extra data on a few nodes it's really hard to see how optimised code could be developed. So I think I'm going to change the approach - rather than have tree nodes point to data, make a structure that pairs up a reference to the node, with a reference to the data - like a dict, and then you can have a dict pairing nodes with DNA sequences, or comments, or so on, and so having several parameterised dicts seems more sensible than expecting nodes to be super flexible about what they contain - possibly loosing performance in the process.

Blahah:
I'm working on tests in a branch off this PR: https://github.com/Blahah/Bio.jl/tree/Phylov1tests.

This is how I'm planning to structure the tests, which roughly maps to the structure of the code itself:

- [ ] Phylogeny
  - [ ] Creation
  - [ ] Rooting
  - [ ] Searching
  - [ ] Traversing
  - [ ] Grafting
- [ ] TreeIO
  - [ ] Newick
    - [ ] Reading
    - [ ] Writing
  - [ ] PhyloXML
    - [ ] Reading
    - [ ] Writing
- [ ] Iteration
  - [ ] Breadth first
  - [ ] Depth first
  - [ ] Searching
- [ ] Annotation

If you have any suggestions for a better structure, please jump in.

binarybana:
A suggestion for TreeIO->Newick->Reading tests: I just tried reading in the [green genes](http://greengenes.lbl.gov/Download/Trees/16S_all_gg_2011_1.tree.gz) tree using `parsenewick` and wasn't successful (looks like leftpcount = rightpcount = 0 for some reason). So a test to parse that might be useful (with a `jldownload` to pull it down before testing).

Blahah:
thanks @binarybana, I'll test on that

dcjones:
@Ward9250, I'm going to go through and make more detailed comments, but here are some high level things that should be addressed before merging:

  * Bring the stye in line with the rest of the code. 4-space indentation, in particular.
  * I don't think generated html (everything under `site`) should be checked into `master`. I'm afraid this will end up creating a lot of noise. It would be ok to have an orphan `gh-pages` branch for this.
  * `alignment.jl` doesn't seem to be used by anything. That should probably be a separate PR, or maybe that's encompassed by #36.
  * This is a lot of commits, so we may want squash this before merging. Don't worry about that yet, though.

REQUIRE
```diff
@@ -1,4 +1,5 @@
-julia 0.3-
-Compat
-Docile
+julia 0.4-
```
dcjones: Is there anything in particular preventing this from working on julia 0.3?

src/phylo/phylotest.jl
```diff
@@ -0,0 +1,24 @@
+# Testscript for Bio.jl phylo.
```
dcjones: I'm guessing this file was just for debugging and can be removed?

src/phylo/nodes.jl
```diff
@@ -0,0 +1,655 @@
...
+      parent = one)
+
+**Parameters:**
+
+* `name`:         The name of the node (optional). Defaults to an empty string, indicating
+                  the node has no name.
+
+* `branchlength`: The branch length of the node from its parent (optional).
+                  Defaults to `-1.0`, indicating an unknown branch length.
+
```
dcjones: Document `confidence` parameter.

src/phylo/nodes.jl
```diff
@@ -0,0 +1,655 @@
...
+* `ext`:          An array of zero or more PhyExtensions (optional). Defaults to an empty
+                  array, i.e. `[]`, indicating there are no extensions.
+
+* `parent`:       The parent node (optional). Defaults to a self-reference, indicating
+                  the node has no parent.
+"""
+  function PhyNode(name::String = "", branchlength::Float64 = -1.0, confidence::Float64 = -1.0, children::Vector{PhyNode} = PhyNode[], parent = nothing)
```
dcjones: I worry a little about using `-1` to represent no value. Negative branch lengths or confidence values may not make sense, but I could imagine they arise anyway in some computation. What about using `Nullable{Float64}`?

src/phylo/nodes.jl
```diff
@@ -0,0 +1,655 @@
...
+@Docile.doc """
+Get the parent of a node.
+
+**Parameters:**
+
+* `x`: The PhyNode to get the parent of.
+""" ->
+function parent(x::PhyNode)
+  if parentisself(x)
+    println("Node does not have a parent. It is self referential.")
```
dcjones: Should be an error, or at least a warning.

src/phylo/nodes.jl
```diff
@@ -0,0 +1,655 @@
...
+end
+
+@Docile.doc """
+Test whether a node is the root node.
+
+**Parameters:**
+
+* `x`: The PhyNode to test.
+""" ->
+function isroot(x::PhyNode)
```
dcjones: If `x` has no parent and no children, isn't it still technically a root node?

src/phylo/nodes.jl
```diff
@@ -0,0 +1,655 @@
...
+If all branch lengths are unknown, the tree is a cladogram, and it is still useful to be able to compare relative distances. If individual branch lengths are unknown, they should not affect the calculation of path distances. To satisfy both of these cases, we use machine epsilon as the minimal distance.
+
+**Parameters:**
+
+* `x`: A PhyNode.
+
+""" ->
+function distanceof(x::PhyNode)
+  bl = branchlength(x)
+  return bl == -1.0 ? eps() : bl
```
dcjones: I can see how returning `eps()` on unknown branch length could be useful, but it seems like it could also be very surprising. You could have two versions, one that errors by default, and another that replaces unknown values with some specified value, like:
```julia
function distanceof(x::PhyNode)
    b1 = branchlength(x)
    if b1 == -1.0
        error("Branch length is unknown.")
    end
    return b1
end

function distanceof(x::PhyNode, replace_unknown::Float64)
    b1 = branchlength(x)
    if b1 == -1.0
        return replace_unknown
    end
    return b1
end
```

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,502 @@
...
+
+@Docile.doc """
+Find the distance of each node from the root.
+
+**Parameters:**
+* `tree`: The Phylogeny to measure.
+""" ->
+function distance(tree::Phylogeny)
+  distances::TreeAnnotations{Float64} = TreeAnnotations{Float64}(tree)
+  function updatedistances(node::PhyNode, currentdist::Float64)
```
dcjones: This code can be inline. It doesn't seem like the function is helping anything, since it's only used once right after it's declared.

src/phylo/iteration.jl
```diff
@@ -0,0 +1,277 @@
...
+#==================================#
+
+# Ben J. Ward, 2014.
+
+@Docile.doc """
+`PhylogenyIterator` is an abstract type that defines a family of iterators
+for traversing trees in various ways, including `BreadthFirst`, and `DepthFirst`,
+or from a tip to a root.
+""" ->
+abstract PhylogenyIterator
```
dcjones: I don't know that this abstract type is needed. Abstract types are useful to groups types that share some common interface, but there's not much of a specialized interface here. These are all just regular iterators.

src/phylo/iteration.jl
```diff
@@ -0,0 +1,277 @@
...
+`Phylogeny`.
+
+A `DepthFirst` phylogeny iterator contains:
+
+- A `PhyNode` that the iterator will `start` at.
+""" ->
+immutable DepthFirst <: PhylogenyIterator
+  start::PhyNode
+  function DepthFirst(x::PhyNode)
+    if isempty(x)
```
dcjones: Should it be valid to iterate over an empty tree? Iterating over empty arrays or dicts doesn't result in an error, is this different?

src/phylo/iteration.jl
```diff
@@ -0,0 +1,277 @@
...
+
+@Docile.doc """
+Iterate over a `Phylogeny` and return a reference to the first `PhyNode` 
+that matches a given condition.
+
+**Parameters:** 
+* `it`: One of any type of `PhylogenyIterator`. This dictates the order in which nodes will be searched.
+* `condition`: A `Function` that accepts a `PhyNode` as the first and only argument, and returns a Bool.
+""" ->
+function Base.search{T <: PhylogenyIterator}(it::T, condition::Function)
```
dcjones: I don't think this function is needed, because you can already do this idiomatically on any iterator with `first(filter(condition, it))`.

src/phylo/iteration.jl
```diff
@@ -0,0 +1,277 @@
...
+
+@Docile.doc """
+Iterate over a `Phylogeny` and return an array of references to all `PhyNode`s 
+that match a given condition.
+
+**Parameters:** 
+* `it`: One of any type of `PhylogenyIterator`. This dictates the order in which nodes will be searched.
+* `condition`: A `Function` that accepts a `PhyNode` as the first and only argument, and returns a Bool.
+""" ->
+function searchall{T <: PhylogenyIterator}(it::T, condition::Function)
```
dcjones: Similarly, `filter(condition, it)` is pretty much the same thing but general purpose.

src/phylo/annotation.jl
```diff
@@ -0,0 +1,39 @@
...
+A type-alias: Dictionaries with keys of type `PhyNode` and values of type `T`
+are aliased as NodeAnnotations{T}.  
+""" ->
+typealias NodeAnnotations{T} Dict{PhyNode, T}
+
+@Docile.doc """
+TreeAnnotations: A type to allow the assignment of data to clades of a phylogeny.
+""" ->
+type TreeAnnotations{T}
+  x::Phylogeny
```
dcjones: Maybe a more descriptive name than `x`? `phylogeny`?

dcjones:
Should Newick names be allowed to have spaces? If so, the name here should be "A B C", and if not, this should be an error, rather than silently disappearing the `A B`. Some peculiar cases:
```
julia> Phylo.parsenewick("A B C")[1].root
C

julia> Phylo.parsenewick("\"A B C\"")[1].root
C"

julia> Phylo.parsenewick("'A B C'")[1].root
RootA B C'

julia> Phylo.parsenewick("A:1.2.3")[1].root
.3
```

You tokenizer implementation is good, except that it skips over anything it doesn't recognize as a token (I think that's what's happening in the cases above). This would be a little more work, but rather than calling `matchall`, you can call `eachmatch` and make sure there's nothing meaningful between tokens. Something like:

```julia

function tokenize(pat, str)
    lastpos = 0
    tokens = String[]
    for mat in eachmatch(pat, str)
        if mat.offset > lastpos + 1
            skipped = str[lastpos+1:mat.offset-1]
            if !isspace(skipped)
                error("Could not tokenize '$(skipped)'")
            end
        end

        push!(tokens, mat.match)

        lastpos = mat.offset + length(mat.match) - 1
    end
    return tokens
end
```

As far as interface: what I've done with the other parsers is to define a `read` function, and empty type to dispatch on, so you can say `read(open(filename), Newick)`.

Also, the XML stuff should either be completed or removed and saved for another PR. The latter would be fine, we don't have to merge everything once.

dcjones:
Oh, and we should merge @Blahah's work on tests into this PR, right?

Ward9250:
@dcjones Do you think a Ragel machine is possible for Newick strings? With a view of replacing the current parser. I've been looking at the documentation you wrote and I'm not completely sure, but I wonder if we could use the Newick Grammar rules on the wiki page here:

https://en.wikipedia.org/wiki/Newick_format

To generate a state machine for Newick that would eliminate current shortcomings of the tokenizer parser.

src/phylo/iteration.jl
Ward9250: The main reason it is used is for Parameteric definition of the Base.search and searchall functions. But maybe there's something better we can do.

Ward9250:
Build works on 0.4 now.

Ward9250:
So If I recall from @dcjones specific comments, I've corrected tab style and so forth. Now I think you made comment on the branch lengths on PhyNodes - can they be Nullable, instead of a -1 to represent a missing branch length.

src/phylo/nodes.jl
Ward9250: That's something we need to decide. Because if a node has no children, and no parent, then it is not part of any phylogeny. 

It is a single, unconnected node and cannot be called a Phylogeny in any meaningful way: Even the most meaningless phylogenetic structure (totally unresolved phylogenies) have a central root node with every tip (every organism or species) branching off it. This is (called an unresolved star phylogeny). If that is as simple as phylogenies can be, then a single unconnected node should not be called one, and so isn't a root insofar as it isn't even part of a phylogeny of which to be a root. This is why they instead of returning true for isroot, they return false. But I believe there is another method that checks is a node is part of a tree / is a single unconnected node.

src/phylo/iteration.jl
Ward9250: Good point, maybe it should be a one-liner syntactic sugar, so writing it in full is not needed.

PR #30, opened 2014-12-24
Fix show methods for nucleotide input

julian-gehring:
Has referred to the non-existing variable 'aa', should refer to the input 'nt'


dcjones:
Thanks!

PR #31, opened 2015-03-17
[Seq] Reduced the number of files

brk00:
As discussed on Gitter, I reverted the file split.


Blahah:
Approve

Ward9250:
Looks good - I concur with Richard.

dcjones:
Looks good to me also. Merging.

PR #32, opened 2015-03-18
[RFC] Intervals

dcjones:
I'm starting in on this. I have an `IntervalSet` type that you can insert into and iterate over. Speed seems reasonable: on julia 0.4, inserting 1.2 million intervals (exons from Ensembl's human annotations) takes about 16 seconds. A lot of that time is spent on the dumb pseudo-parsing of the GTF file I did for the benchmark. So this is a pretty promising start, and I think it will get faster.

Some fun bikeshedding :bike: :house::

After some hand-wringing, I think I prefer the term "interval" to "range", hence the names in this PR. The relevant OED definitions of "range": 
> A row, line, or series of things.

> A set of points on a straight line, esp. as determining a pencil (pencil n. 7) of lines joining each point to some non-collinear point.

And of "interval".
> An open space lying between two things or two parts of the same thing; a gap, opening. Also, an intervening portion of something.

"An intervening portion of something" is pretty precise to our use, and "interval" has fewer definitions than "range", so it's sort of a tighter fit.

I called the collection of intervals `IntervalSet`, but that's not a very good name. It's not a set, because it can contain the same interval multiple times. A more accurate name would be `IntervalBag` or `IntervalMultiset`, but those might make it sound weirder than it is, which is just an indexed collection of annotated intervals. Open to suggestions here.


Blahah:
Excellent!

I agree that the `Set` nomenclature is misleading - to the point where I think people will assume it acts as a set. Perhaps `IntervalCollection`, or, if it's just an array of intervals, `IntervalArray`?

Blahah:
OK, having actually looked at the code, I can see it's not that simple - `IntervalSet` is a hash mapping a name to an `IntervalTree`. So it can only have each name occurring once, but the same interval could appear under multiple names. With that in mind I vote for a name that implies a generic collection of things, without implying the rules that govern its methods. So, `IntervalCollection`, `IntervalStore`, or something like that?

Ward9250:
This looks very cool - haven't chance to read thoroughly until the weekend, but seen the "Files Changed". I was gonna say what about IntervalCollection, but see Richard already suggested it.

dcjones:
Ok, I like `IntervalCollection`. A good name for something intended as a general purpose collection of intervals.

brk00:
Nice work! 

`IntervalTree.jl` isn't on Julia Packages, so we can't download them with REQUIRE. That seems to be one of the problems with Travis.

dcjones:
I should probably just add IntervalTrees to METADATA, since that's pretty complete and stand-alone.

dcjones:
This is now rebased on my io branch (PR #40). It doesn't strictly depend on that, but the work I'm going to do on BED, GTF, GFF parsers needs io and intervals and this seems like the cleanest way. So reminder: don't merge until #40 is merged.

dcjones:
This is well tested now and ready to review and merge.

src/intervals/interval.jl
```diff
@@ -0,0 +1,175 @@
...
+function isless(a::Strand, b::Strand)
+    return convert(Uint8, a) < convert(Uint8, b)
+end
+
+
+# Note, just to be clear: this shadows IntervalTrees.Interval
+@doc """
+A genomic interval specifies interval with some associated metadata.
+""" ->
+immutable Interval{T} <: IntervalTrees.AbstractInterval{Int64}
```
brk00: Since `IntervalTrees` export `AbstractInterval`, we don't need to write explicitly the module name here, do we?

src/intervals/interval.jl
```diff
@@ -0,0 +1,175 @@
...
+    seqname::String
+    first::Int64
+    last::Int64
+    strand::Strand
+    metadata::T
+end
+
+
+function IntervalTrees.first(i::Interval)
```
brk00: We could explicitly import `first` and `last` on `intervals.jl`

src/intervals/intervalcollection.jl
```diff
@@ -0,0 +1,245 @@
...
+            println(io, "  ", i)
+        end
+        if n_entries > max_entries
+            print(io, "  ⋮")
+        end
+    end
+end
+
+
+function length(ic::IntervalCollection)
```
brk00: A one-liner here would be more consistent

src/intervals/intervalcollection.jl
```diff
@@ -0,0 +1,245 @@
...
+end
+
+
+immutable IntersectIterator{S, T}
+    a_trees::Vector{IntervalCollectionTree{S}}
+    b_trees::Vector{IntervalCollectionTree{T}}
+end
+
+
+immutable IntersectIteratorState{S, T}
```
brk00: What `S` and `T` mean? Why aren't they used inside the type definition?

src/intervals/intervalcollection.jl
```diff
@@ -0,0 +1,245 @@
...
+        return new(i, intersect_iterator, intersect_iterator_state)
+    end
+end
+
+
+@doc """
+Iterate over pairs of intersecting intervals in two IntervalCollections.
+""" ->
+function intersect{S, T}(a::IntervalCollection{S}, b::IntervalCollection{T})
+    seqnames = collect(String, intersect(Set(keys(a.trees)), Set(keys(b.trees))))
```
brk00: Instead of:

```julia
intersect(Set(keys(a.trees)), Set(keys(b.trees)))
```

You should try:

```julia
keys(a.trees) ∩ keys(b.trees)
```

I found it simpler to read and also faster

src/intervals/intervalcollection.jl
```diff
@@ -0,0 +1,245 @@
...
+#                         │
+#                         ▼
+#                  ╔════════════╗
+#                  ║ Metadata 2 ║
+#                  ╚════════════╝
+#                         │
+#                         ▼
+#
+#                        ...
+#
```
brk00: Nice diagram! Have you used a GUI tool to produce this?

src/intervals/intervalstream.jl
```diff
@@ -0,0 +1,180 @@
...
+    a_state::U
+    b_state::V
+end
+
+
+@doc """
+Intersect two `IntervalStreams` returning an iterator over all pairs of
+intersecting intervals.
+""" ->
+function Base.intersect{S, T}(a::IntervalStreamOrArray{S},
```
brk00: This generic function is imported explicit on this module, therefore we don't need `Base.`

src/intervals/intervalstream.jl
```diff
@@ -0,0 +1,180 @@
...
+        else # intersection
+            break
+        end
+    end
+
+    return a_interval, b_interval, a_state, b_state
+end
+
+
+function Base.start{S, T}(it::IntervalStreamIntersectIterator{S, T})
```
brk00: This generic function is imported explicit on this module, therefore we don't need `Base.` here.

src/intervals/intervalstream.jl
```diff
@@ -0,0 +1,180 @@
...
+        it.b_interval = b_interval
+    end
+
+    return IntervalStreamIntersectIteratorState(1, a_state, b_state)
+end
+
+
+# Do we need a state if this is going to be stateful. State can maybe just
+# be the current interval in b
+function Base.next{S, T, U, V}(it::IntervalStreamIntersectIterator{S, T},
```
brk00: This generic function is imported explicit on this module, therefore we don't need `Base.` here.

src/intervals/intervalstream.jl
```diff
@@ -0,0 +1,180 @@
...
+            it.b_interval = b_interval
+        end
+    end
+
+    return value, IntervalStreamIntersectIteratorState{U, V}(
+                        a_buffer_pos, a_state, b_state)
+end
+
+
+function Base.done{S, T, U, V}(it::IntervalStreamIntersectIterator{S, T},
```
brk00: This generic function is imported explicit on this module, therefore we don't need `Base.` here.

src/intervals/stream_buffer.jl
```diff
@@ -0,0 +1,94 @@
...
+    first_idx::Int
+    last_idx::Int
+
+    function StreamBuffer()
+        return new(Array(T, 16), 1, 0)
+    end
+end
+
+
+function Base.length{T}(buf::StreamBuffer{T})
```
brk00: This generic function is imported explicit on this module, therefore we don't need `Base.` here.

src/intervals/intervals.jl
```diff
@@ -0,0 +1,23 @@
+
+module Intervals
+
+import Base: show, isless, push!, intersect, start, next, done, length
```
brk00: We should also import `isempty`, `endof`, `shift!` and `getindex`

src/intervals/stream_buffer.jl
```diff
@@ -0,0 +1,94 @@
...
+    return buf.last_idx == 0
+end
+
+
+function Base.endof{T}(buf::StreamBuffer{T})
+    return length(buf)
+end
+
+
+function Base.push!{T}(buf::StreamBuffer{T}, x::T)
```
brk00: This generic function is imported explicit on this module, therefore we don't need Base. here.

src/intervals/stream_buffer.jl
```diff
@@ -0,0 +1,94 @@
...
+        return 0
+    elseif buf.last_idx < buf.first_idx
+        return length(buf.data) + buf.last_idx - buf.first_idx + 1
+    else
+        return buf.last_idx - buf.first_idx + 1
+    end
+end
+
+
+function Base.isempty{T}(buf::StreamBuffer{T})
```
brk00: This generic function should imported explicit on this module, therefore we wouldn't need `Base.` here.

src/intervals/stream_buffer.jl
```diff
@@ -0,0 +1,94 @@
...
+    return buf.last_idx == 0
+end
+
+
+function Base.endof{T}(buf::StreamBuffer{T})
```
brk00: This generic function should imported explicit on this module, therefore we wouldn't need `Base.` here.

src/intervals/stream_buffer.jl
```diff
@@ -0,0 +1,94 @@
...
+
+    buf.last_idx += 1
+    if buf.last_idx > length(buf.data)
+        buf.last_idx -= length(buf.data)
+    end
+    @inbounds buf.data[buf.last_idx] = x
+end
+
+
+function Base.shift!{T}(buf::StreamBuffer{T})
```
brk00: This generic function should imported explicit on this module, therefore we wouldn't need `Base.` here.

src/intervals/stream_buffer.jl
```diff
@@ -0,0 +1,94 @@
...
+        buf.first_idx = 1
+    else
+        buf.first_idx += 1
+    end
+
+    return x
+end
+
+
+function Base.getindex{T}(buf::StreamBuffer{T}, idx::Int)
```
brk00: This generic function should imported explicit on this module, therefore we wouldn't need `Base.` here.

test/intervals/test_intervals.jl
```diff
@@ -0,0 +1,239 @@
+module TestIntervals
+
+using FactCheck
+using Distributions
+using Bio
+using Bio.Intervals
+
+
+# Generate an array of n random Interval{Int} object. With sequence names
```
brk00: This should be a docstring

src/intervals/interval.jl
```diff
@@ -0,0 +1,175 @@
...
+        print(io, "-")
+    elseif strand == STRAND_BOTH
+        print(io, ".")
+    else
+        print(io, "(undefined strand)")
+    end
+end
+
+
+function isless(a::Strand, b::Strand)
```
brk00: An one-liner here would be more consistent

src/intervals/intervalcollection.jl
dcjones: Yeah, I was trying out [Monodraw](http://monodraw.helftone.com/). It's really nice for in-source diagrams like this. I think I may have to buy a license now that it's out of beta.

src/intervals/intervalcollection.jl
dcjones: Oh, I intended to narrow down its member's types but then forgot to, so they are unused. I removed `S` and `T` for now, and will work on optimizing this code later.

dcjones:
Thanks for the review @brk00! I think I addressed everything.

brk00:
I believe someone with more bioinformatics background sould review before merging. Awesome PR by the way.

brk00:
Is Coveralls aware of the new files introduced on this PR?

src/intervals/interval.jl
```diff
@@ -0,0 +1,173 @@
...
+    end
+end
+
+
+@doc """
+Return true if interval `a` entirely precedes `b`.
+""" ->
+function precedes{T}(a::Interval{T}, b::Interval{T},
+                     seqname_isless::Function=alphanum_isless)
+    return seqname_isless(a.seqname, b.seqname) ||
```
binarybana: Rather than silently comparing seqname's I would instead hope that this (and perhaps isless) would instead fail loudly when trying to compare intervals of different sequences?

binarybana: I can understand the ``isoverlapping`` check using seqnames below however (as two intervals clearly don't overlap when they aren't on the same sequence).

binarybana: Oh sorry, I just now read the docstring on ``alphanum_isless``. Perhaps the above function docstrings should reference that docstring to enhance transparency on the semantics.

test/intervals/test_intervals.jl
```diff
@@ -0,0 +1,239 @@
...
+            push!(ic_b, interval)
+        end
+
+        @fact sort(collect(intersect(ic_a, ic_b))) ==
+              sort(simple_intersection(intervals_a, intervals_b)) => true
+    end
+
+
+    context("Show") do
+        nullout = open("/dev/null", "w")
```
brk00: Does this work on Windows?

test/intervals/test_intervals.jl
```diff
@@ -0,0 +1,239 @@
...
+        for interval in intervals_b
+            push!(ic_b, interval)
+        end
+
+        @fact sort(collect(intersect(ic_a, ic_b))) ==
+              sort(simple_intersection(intervals_a, intervals_b)) => true
+    end
+
+
+    context("Show") do
```
brk00: This test just verify that `show` doesn't throw any exceptions/errors, but doesn't certify that it's working properly, right?

test/intervals/test_intervals.jl
```diff
@@ -0,0 +1,239 @@
...
+
+        show(nullout, STRAND_NA)
+        show(nullout, STRAND_POS)
+        show(nullout, STRAND_NEG)
+        show(nullout, STRAND_BOTH)
+    end
+end
+
+
+facts("Alphanumeric Sorting") do
```
brk00: This should come before `facts("IntervalCollection")`, since the functions tested here are used tests in that block.

test/intervals/test_intervals.jl
```diff
@@ -0,0 +1,239 @@
...
+    return intervals
+end
+
+
+# A simple interval intersection implementation to test against.
+function simple_intersection(intervals_a, intervals_b)
+    sort!(intervals_a)
+    sort!(intervals_b)
+
+    intersections = Any[]
```
brk00: Instead of `Any[]`, we could use `(eltype(intervals_a), elemtype(intervals_b))[]` to make the type of this Array tighter. But that's just a silly optimization.

test/intervals/test_intervals.jl
```diff
@@ -0,0 +1,239 @@
+module TestIntervals
+
+using FactCheck
+using Distributions
+using Bio
```
prcastro: Is this needed?

src/intervals/interval.jl
dcjones: I think having a well-defined ordering of intervals across different sequences is useful in practice, especially for intersecting interval streams, which only works if the two files being intersected are sorted the same way. For that reason, it makes some sense that `isless` would order seqnames. Wether or not `precedes` should work the same way is debatable. On one hand, consistency with `isless` is nice, on the other hand, we should do anything in our power to prevent programming errors.

binarybana: Another idea is to modify alphanum_isless to error on string mismatches, but compare chr## type strings just fine so that: 
`chr12` < `chr18`
`chr1` < `chr18`
(as currently) yet,
`alphanum_isless("Firmicutes_CAG_218", "Bacteriodetes_13")` raises a `SequenceComparisonError`.

Though this complicates the semantics even further. 

Or even more exotically, have sequences use a string type parameter of the organism name, so that comparisons are valid inside of a species and invalid otherwise...

My vote is to keep things the way you wrote them, just add a note in the docstrings of `precedes`, and `isless` about the sequence name comparison.

prcastro:
I think we should merge this this week, and open PRs and issues to discuss things further. This PR is opened for more than 60 days, more than sufficient to discuss its core. After all, this feature is working properly and is well documented and well tested.

With this, Bio.jl would move more quickly and become usable by the end-user sooner. Also, JuliaCon is close, and we should have things working until then. What do you think?

binarybana:
+1 LGTM!

Blahah:
+1, let's merge it.

src/intervals/interval.jl
dcjones: It would be great if we could have a way to associate an `IntervalCollection` with a particular genome. This would prevent a lot of silly mistakes caused by mixing data from two different genome revisions. Once or twice I've even accidentally viewed mouse data against the human genome. It seems insane that the software couldn't catch such a basic mistake.

That said, I don't think we can do this reliably from chromosome names. For example, under UCSC naming, mouse and human mostly share the same names. Maybe what we need is for `IntervalCollection` and `read` functions to optionally take a genome accession. When the genome is provided, we can sanity check various operations. So it would work almost like a type annotation in Julia.

I'll keep `isless` and `precedes` as they are and comment them, like you suggest.

PR #33, opened 2015-03-30
Added 26 tests for Nucleotide Sequences

brk00:
Added tests for sequence length, sequence access (iterations and out of bounds access), creation of empty sequences and creation of subsequence from ranges.

I hope the coverage goes up a little bit with this PR.

### DIFF
--- a/test/seq/test_seq.jl
+++ b/test/seq/test_seq.jl
@@ -112,6 +112,12 @@ facts("NucleotideSequence Construction") do
     @fact_throws DNASequence("ACCNNCATTTTTTAGATXATAG")
     @fact_throws RNASequence("ACCNNCATTTTTTAGATXATAG")
 
+    # Check construction of empty nucleotide sequences
+    #  using RNASequence and DNASequence functions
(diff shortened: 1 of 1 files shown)


brk00:
Doesn't Coveralls track the PRs on this repo?

dcjones:
Nice work. We should get coverage, but there are currently some issues with dependencies of the coverage package on julia 0.4 nightlies (https://github.com/jakebolewski/JuliaParser.jl/issues/18). I guess we could change our travis config to get coverage data from julia 0.3.

brk00:
I believe we should change travis config to get coverage data from julia 0.3 util 0.4 is out.

dcjones:
Ok, let's merge this (I don't think anyone would object to more tests) then see if we can get coverage working.

dcjones:
Ok, we have coverage data once again!
https://coveralls.io/r/BioJulia/Bio.jl

brk00:
Very nice, coverage is up as expected. I'll try to improve coverage even more later this week. I expect getting something like ~80% for all Seq files (nucleotides, aminoacids and geneticcodes).

PR #34, opened 2015-04-01
[Seq] More tests, less bugs

brk00:
Added some more tests for nucleotides, nucleotide sequences and k-mers and solved a bug with the function `rnakmer(seq::RNASequence)`. I also modified the organization of the tests, with more contexts and sub-contexts (I believe it's more clear when the tests run). As of now, I don't expect to see coverage going up a lot.

I'll add some more tests later this week.

### DIFF
--- a/src/seq/geneticcode.jl
+++ b/src/seq/geneticcode.jl
@@ -19,14 +19,14 @@ length(code::GeneticCode) = 64
 # Iterating through genetic code
 # ------------------------------
 
-start(code::GeneticCode) = uint64(0)
+start(code::GeneticCode) = @compat uint64(0)
 
 function next(code::GeneticCode, x::Uint64)
     c = convert(Codon, x)
     return ((c, code[c]), (x + 1))
 end
 
-done(code::GeneticCode, x::Uint64) = (x > uint64(0b111111))
+done(code::GeneticCode, x::Uint64) = (x > (@compat uint64(0b111111)))
 
 
 # Default genetic codes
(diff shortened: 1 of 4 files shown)


brk00:
Added more tests and now `next{T}(seq::NucleotideSequence{T}, i)` throws a `BoundsError()` when gets an index that results in some out-of-bounds access.

dcjones:
> I also noticed something unusual: our start, done and next functions for NucleotideSequence are shifted -1 with respect to Julia standards. If I change that, somehow it also breaks the rest of the code. Is this intentional @dcjones?

It is intentional, but it's not really important. If the index starts at 1, computing an offset into the array involves computing `divrem(i - 1, 32)`. You can save a `- 1` if `i` starts at 0 instead of 1. I think that's what I was thinking, anyway.

brk00:
I see. On the other hand, I belive that we have a `+1` on `next` and a `-1` on `start`. So the number of operations still the same.

dcjones:
> I see. On the other hand, I belive that we have a +1 on next and a -1 on start. So the number of operations still the same.

That's entirely true.  Judging from the `# TODO: This can be made a lot faster.` comment I left, I think I was intending to revisit that iterator anyway.

dcjones:
@brk00 I made PR against your fork to get the tests to pass on 0.4: https://github.com/brk00/Bio.jl/pull/1

Merge that, then I think we can merge this PR if the tests pass, unless it's still WIP.

brk00:
I believe it's better if we merge once travis finishes. I'll make a PR later with more tests.

dcjones:
> Coverage increased (+11.67%) to 83.1%
Nice!

Blahah:
Great work! That's a huge bump :)

PR #35, opened 2015-04-10
make unsafe_complement! return a transformed sequence

bicycle1885:
which is a standard behavior of destructive functions
(cf. Base.push!, Base.append!, etc.)


dcjones:
Thanks!

PR #37, opened 2015-04-12
Mutable Sequences

dcjones:
This extends `NucleotideSequence` to allow mutability. Sequences are still immutable by default, but they can be made mutable. Making a sequence mutable means you lose the ability to make cheap subsequences. You can still make subsequences, but they'll copy the underlying data rather than sharing it.

Changing from mutable to immutable with `immutable!` is always cheap, changing from immutable to mutable with `mutable!` is cheap iff subsequences aren't involved, otherwise some data will be copied.

E.g. the following example modifies a sequence without making copies.
```
julia> using Bio.Seq

julia> x = dna"ACGTNNNACGT"
11nt DNA Sequence
 ACGTNNNACGT

julia> x[1] = 'N'
ERROR: Cannot mutate an immutable sequence. Call `mutable!(seq)` first.
 in setindex! at /Users/dcjones/.julia/v0.3/Bio/src/seq/nucleotide.jl:340
 in setindex! at /Users/dcjones/.julia/v0.3/Bio/src/seq/nucleotide.jl:359

julia> mutable!(x)
11nt Mutable DNA Sequence
 ACGTNNNACGT

julia> x[1] = 'N'
'N'

julia> x
11nt Mutable DNA Sequence
 NCGTNNNACGT

julia> immutable!(x)
11nt DNA Sequence
 NCGTNNNACGT
```

TODO:
  * [x] Similar functionality on `AminoAcidSequence`.
  * [ ] Setting a range: `seq[i:j] = other_seq`.
  * [x] Copy functions: `copy!(seq, other_seq)`.
  * [ ] Vector operations: `push!`, `pop!`, `shift!`


brk00:
Very interesting. Would be nice to see some tests included on the PR. =)

src/seq/nucleotide.jl
```diff
@@ -136,16 +147,32 @@ end
 # ------------
 
 # Construct an empty sequence
-NucleotideSequence{T<:Nucleotide}(::Type{T}) = NucleotideSequence{T}(zeros(Uint64, 0), BitVector(0), 1:0)
+NucleotideSequence{T<:Nucleotide}(::Type{T}; mutable::Bool=false) =
+    NucleotideSequence{T}(zeros(Uint64, 0), BitVector(0), 1:0, mutable, false)
```
brk00: What mutable does on this line?

src/seq/nucleotide.jl
```diff
@@ -119,16 +119,27 @@ end
 # represented with an N mask BitArray. If the ns[i] bit is set, then the
 # sequence may have any nucleotide at that position and it must be ignored.
 #
-# Sequences are immutable by convention. Low-level functions will mutate the
-# underlying data, any user-facing function that does so should have `unsafe` in
-# the name. Immutability allows two sequences to share the same underlying data,
```
diegozea: Why aren't we using SubArray for this?

brk00: How do you suggest we use SubArrays here?

diegozea: I'm not sure if they are useful in this situation. In fact, I want to know why are they not useful/used?

dcjones: I don't think SubArray would be especially useful here. What we are doing is quite similar to SubArray, but for subsequences we need to be able to specify the range at the resolution of bits, since this is 2-bit encoded, whereas a SubArray would only let us specify ranges at best at the byte level.

dcjones:
I've added a few tests and extended this interface to `AminoAcidSequence` so this should be safe to merge.

src/seq/aminoacid.jl
```diff
@@ -290,17 +314,57 @@ function getindex(seq::AminoAcidSequence, i::Integer)
     return seq.data[i]
 end
 
+
 # Construct a subsequence
 getindex(seq::AminoAcidSequence, r::UnitRange) = AminoAcidSequence(seq, r)
 
+
 # Replace a AminoSequence's data with a copy, copying only what's needed.
 function orphan!(seq::AminoAcidSequence, reorphan=true)
```
prcastro: `reorphan` isn't used inside this function

src/seq/aminoacid.jl
```diff
@@ -290,17 +314,57 @@ function getindex(seq::AminoAcidSequence, i::Integer)
...
+
+function setindex!(seq::AminoAcidSequence, nt::Char, i::Integer)
+    setindex!(seq, convert(AminoAcid, nt), i)
+end
+
+
+# Mutability/Immutability
+# -----------------------
+
+ismutable(seq::AminoAcidSequence) = return seq.mutable
```
prcastro: We normally don't use `return` keyword in one-line function definitions

src/seq/aminoacid.jl
```diff
@@ -290,17 +314,57 @@ function getindex(seq::AminoAcidSequence, i::Integer)
...
+
+
+function mutable!(seq::AminoAcidSequence)
+    if !seq.mutable
+        if seq.hasrelatives
+            orphan!(seq, true)
```
prcastro: The second argument here isn't necessary

src/seq/aminoacid.jl
```diff
@@ -308,11 +372,13 @@ copy(seq::AminoAcidSequence) = orphan!(AminoAcidSequence(seq.data, seq.part))
 
 start(seq::AminoAcidSequence) = seq.part.start
 
+
 function next(seq::AminoAcidSequence, i)
     aa = seq.data[i]
     return (aa, i + 1)
```
prcastro: Do we need this intermediate variable here? Maybe the compiler can optimize this, but still, I believe it's much clearer to don't use `aa` here (and maybe transforming this function definition in a one-liner).

src/seq/nucleotide.jl
```diff
@@ -200,13 +212,30 @@ NucleotideSequence{T<:Nucleotide}(::Type{T}) = NucleotideSequence{T}(zeros(Uint6
...
-    return NucleotideSequence{T}(other.data, other.ns, start:stop)
+
+    seq = NucleotideSequence{T}(other.data, other.ns, start:stop, mutable, true)
+
+    if other.mutable || mutable
+        orphan!(seq, true)
+    end
+
+    if !other.mutable
+        other.hasrelatives = true
```
prcastro: We could simplify this in the same way you did in the Aminoacid Sequence constructor:

```julia
other.hasrelatives = !other.mutable
seq.hasrelatives = !mutable
```

src/seq/nucleotide.jl
```diff
@@ -563,6 +639,32 @@ end
...
+        error("Cannot mutate an immutable sequence. Call `mutable!(seq)` first.")
+    end
+    if i < 1 || i > length(seq)
+        throw(BoundsError())
+    end
+
+    d, r = divrem(i - 1, 32)
+    if nt == nnucleotide(T)
+        seq.ns[i] = true
+        seq.data[d + 1] $= (@compat UInt64(0b11)) << (2*r)
```
prcastro: Should compat be used? Aren't we aiming only for the v0.4 release?

src/seq/nucleotide.jl
```diff
@@ -563,6 +639,32 @@ end
 getindex{T}(seq::NucleotideSequence{T}, r::UnitRange) = NucleotideSequence{T}(seq, r)
 
 
+function setindex!{T}(seq::NucleotideSequence{T}, nt::T, i::Integer)
+    if !seq.mutable
+        error("Cannot mutate an immutable sequence. Call `mutable!(seq)` first.")
+    end
+    if i < 1 || i > length(seq)
+        throw(BoundsError())
```
prcastro: Since we perform a bound checking here, shouldn't we use `@inbounds` in the rest of this function? Or isn't it necessary?

src/seq/nucleotide.jl
```diff
@@ -487,11 +553,19 @@ convert(::Type{String}, seq::NucleotideSequence) = convert(String, [convert(Char
...
 
-"Convert a RNASequence to a DNASequence"
-convert(::Type{DNASequence}, seq::RNASequence) = DNASequence(seq.data, seq.ns, seq.part)
+convert{T}(::Type{NucleotideSequence{T}}, seq::NucleotideSequence{T}) = seq
+
+
+"Convert a DNASequence to a RNASequence and vice versa"
+function convert{T}(::Type{NucleotideSequence{T}}, seq::NucleotideSequence)
+    newseq = NucleotideSequence{T}(seq.data, seq.ns, seq.part, seq.mutable, seq.hasrelatives)
+    if seq.mutable
```
prcastro: Does any of our tests fall inside this `if` clause?

prcastro:
Very nice PR. After dicussing the things I pointed out, I believe we should merge it, even if some of the planned features, like setting a range and vector operations aren't ready yet. Maybe they can be addressed with another PR.

src/seq/aminoacid.jl
dcjones: Fixed.

src/seq/aminoacid.jl
dcjones: Fixed.

src/seq/nucleotide.jl
dcjones: Not quite, because `other.hasrelatives` may be already true, and I don't want to flip it to false in that case.

src/seq/nucleotide.jl
dcjones: We are aiming for 0.4. We can add this to the compat purge in #91.

src/seq/nucleotide.jl
dcjones: Fixed.

src/seq/aminoacid.jl
dcjones: I changed the default to `false`, so now it is necessary.

src/seq/nucleotide.jl
dcjones: Added tests.

PR #40, opened 2015-04-14
RFC: Parsing infrastructure, FASTA and FASTQ parsers

dcjones:
Was #14


src/seq/fasta.jl
```diff
@@ -0,0 +1,1123 @@
+# FASTA sequence types
```
bicycle1885: I think it would be nice if generated files contain a comment like "This file is generated from *.rl. Do not edit me!" at the top of them.

dcjones: Good idea. Done.

src/seq/fasta.rl
```diff
@@ -0,0 +1,251 @@
+
+
+# FASTA sequence types
+
+@doc """
+Metadata for FASTA sequence records containing just a `description` field.
+""" ->
+type FASTAMetadata
+    description::String
```
bicycle1885: Sequences downloaded from RefSeq will have both name and description text, but real-world FASTA files often lacks this description text (e.g. Protein Data Bank).
So, I think wrapping ti with `Nullable` would be better.

dcjones: My impression has been that descriptions are fairly common. E.g. Ensembl, miRBase, NCBI, SGD, FlyBase all use them. Do you think there would be a practical benefit to using Nullable to indicate no description, as opposed to just an empty string?

bicycle1885: Just because it is impossible to distinguish non-existence from empty string.
Writer of FASTA will not be able to know whether a white space should be added after the sequence name.

brk00:
Maybe we should avoid writing explicit code on `seq.jl`. Reserving this file to module definition, `using`, `import` and `export` statements and calls to `include` may be a good practice, and is the standard in many Julia packages (like Distributions.jl, DataFrames.jl, Images.jl, etc).

In my opinion it also makes code easier to search and modify.

src/seq/seq.jl
```diff
@@ -15,15 +16,69 @@ export Nucleotide, DNANucleotide, RNANucleotide,
...
 
 include("nucleotide.jl")
 include("aminoacid.jl")
 include("geneticcode.jl")
 include("util.jl")
+include("alphabet.jl")
+include("quality.jl")
+
+
+# A sequence record is a named sequence with attached metadata.
```
brk00: Maybe a `SeqRecord.jl` file from here to line 77?

src/ragel.jl
```diff
@@ -0,0 +1,376 @@
...
+    for i in 1:a.pos-1
+        if a.data[i] != b.data[i]
+            return false
+        end
+    end
+    return true
+end
+
+
+function Base.takebuf_string{T}(buf::Buffer{T})
```
brk00: These methods should be explicit imported into the module before extension.

src/seq/seq.jl
```diff
@@ -15,15 +16,69 @@ export Nucleotide, DNANucleotide, RNANucleotide,
...
+@doc """
+`SeqRecord{S, T}` is a type holding a named sequence of type `S`, along with
+some arbitrary metadata of type `T`.
+""" ->
+type SeqRecord{S, T}
+    name::String
+    seq::S
+    metadata::T
+
+    function SeqRecord(name, seq::S, metadata::T)
```
brk00: Since this is the default constructor, it could be omitted as long as `SeqRecord()` (line 53) is defined outside the type definition.

test/seq/test_seq.jl
```diff
@@ -1179,4 +1180,70 @@ facts("Translation") do
     @fact_throws translate(rna"ACGUACGNU") # can't translate N
 end
 
+
```
brk00: I think it would be easier to group all this into a `facts("Parsing") do` block. With this, both `"FASTA Parsing"` and `"FASTAQ parsing"` would become `context` blocks

src/seq/fasta.rl
bicycle1885: Please don't take my comment seriously. My anxiety would be trivial.
I really approve this pull request.

dcjones: It's a reasonable suggestion. In the BED parser I ended up making most of the fields Nullable, since most of the columns in BED are optional. Making the description Nullable here might be the most consistent thing to do.

bicycle1885:
:tada:

PR #36, opened 2015-04-16
Align Module - Spec and Roadmap

Ward9250:
We need a module for working with sequence alignments - Reads to references, pairwise alignments, and multiple sequence alignments. Here is an issue to discuss this module - what is required, what design would satisfy the requirements, and what is required to bring such a module to completion.

Let's collectively spec out what we want to do with the Alignment module!

## Requirements of a sequence alignment type(s)

* Representation of whole sequence alignments. 
* Representation of local sequence alignments.

The kinds of files that store these kinds of alignment are clustal, fasta, MSA, phylip.

* Representation of reads aligned to a reference. Such as contained in SAM and BAM files.
* Representation of regions of genomes aligned together - for example in MAF (multiple-alignment .maf) format.

* Allow the manipulation and editing / building of sequence alignments, allowing people to build alignments and allowing them to create sequence alignment algorithms in Julia. 

* Iteration over base positions in the sequence alignment and get the state of each sequence at that position (a base, an amino acid, or a gap). 

One question is - are we going to go with different types / methods of representation for each type of alignment, or are we going to go for a singe type / representation in Julia that can be used to work with alignments?


Ward9250:
I've made some pushes to my branch SequenceAlignments which contains some code to be starting with - I've made an ArrayGaps type, which stores a series of gap-nongap run lengths in an array. Methods are written to allow translation from a position in a view, to a position in a source, and vice versa. I've also made a start on writing code for, and thinking about AnchorGaps - @dcjones I wonder if this would be a good time to think about if we can use ranges or the range / range tree types you've worked on with an ArrayGaps like implementation of an alignment.

dcjones:
I think the pros of using an IntervalTree are mainly that inserting gaps into the middle of a sequence would be faster. I'm not sure if that's a useful thing to be able to do or not. I don't often work with alignments besides short reads so I'm not sure what common use cases are. With short reads the alignments are so small/simple that data structures don't matter much.

What I'm not really clear on with the SeqAn implementations is when `ArrayGaps` should be preferred to `AnchorGaps`. From the docs

> In general, using Array Gaps is sufficient for most applications. ... Alternatively, one should prefer Anchor Gaps if many conversions between coordinates of the gap and the source space are needed ...

But it doesn't say what the downside of anchor gaps is. My preference would be to have one primary implementation, even if it does have minor downsides. But maybe there is good reason to have multiple that I haven't comprehended yet.

Ward9250:
My use case of alignments is typically as a starting point for any phylogenetic or evolutionary study. An alignment is essential as it reveals substitution history or history of insertion and deletions. 

It is my understanding that ArrayGaps would be better for modifying an existing gaps structure or for building alignments (assuming you're not doing so using graphs as some programs do), and for accessing whole aligned sequences. whereas AnchorGaps is a better for getting individual base positions. 

A while ago, I went a bit mental and built my own C++ classes for working with some sequence alignments - using characters including '-' in the sequances, you could either store each sequence in a vector, or you could have a vector of sites, which has the state of each seq at that position. It think it's a similar situation regarding access of positions in the alignment. 

So I think AnchorGaps would be better for say, looping over sites in an alignment and finding polymorphisms or other such things.

dcjones:
The more I think about this, the more I like the AnchorGaps approach. It's simple, compact, it allows for fast (O(log n)) bidirectional conversion between gapped and ungapped indexes, inserting gaps at the end is O(1), though modifying/inserting/deleting gaps in the middle is O(n). The question is then whether we need any other representation besides AnchorGaps.

For short read alignments, the standard is CIGAR, which is compact, supports a richer set of alignment annotations, but doesn't have the nice indexing properties of AnchorGaps, which makes it less suitable for manipulating alignments of long sequences. So if we adopted AnchorGaps as the primary alignment representation, I think we should try to extend it to be able to losslessly represent CIGAR strings.

I think we can do that by associating CIGAR operations with anchors and introducing some extra anchors. CIGAR operations are

  * `M`: Match
  * `I`: Insertion in reference
  * `D`: Deletion from reference.
  * `N`: Skipped region from reference (i.e. gaps)
  * `S`: Soft clipping (which is typically leading or trailing gaps)
  * `H`: Hard clipping (leading or trailing gaps that overhang the reference sequence)
  * `P`: Deletions from a padded/gapped reference.
  * `=`:  Sequence match. Like `M` but implies the sequences match.
  * `X`: Sequence mismatch: Like `N` but implies the sequences don't match.

Here's basically how I think it would work. 

### Matches and Mismatches (`M`, `N`, `=`, `X`)

These are the easiest to handle. My understanding of what AnchorGaps does is that it takes something like this
```
 ACG---TACGT--AC  
```
and represents the gaps by annotating the position following a gap with `(ungapped position, gapped position)` pairs.
```
  ACGTACGTAC  
     │    │   
     ▼    ▼   
    4,7  9,14 
```

The basic idea, is to annotate anchors with cigar operations corresponding to the preceding gap, and insert additional anchors to represent other non-gap cigar ops.

If it's just match/mismatch it's easy. `3M3N5M2N2M` which is equivalent to the alignment above becomes.

```
   ACGTACGTAC       
     ││    ││       
 ┌───┘▼    ▼└────┐  
 ▼   4,7  9,14   ▼  
3,3   N    N   11,15
 M               M  
```

So there are twice as many anchors, but now we can fully represent CIGAR. So we can use `=/X` instead of `M/N`, and the AnchorGaps algorithms (e.g. binary search to convert coordinates) all work as before.

### Clipping (`S`, `H`)

These are just special trailing/leading gaps, so e.g. `ACG` with CIGAR `3S3M3H` becomes
```
  ACG      
  │ ││     
  ▼ │└──┐  
 1,4│   ▼  
  S │ 4,10 
    │   H  
    ▼      
   3,6     
    M      
```
The last anchor here indicating the trailing gap by giving the coordinate past the end.

### Insertions/Deletions (`I`, `D`)

The only thing tricky here is that with `I` and `D` there would no longer a one-to-one correspondence between gapped positions and ungapped-positions. I think that's fine, we'll just have to be careful to define the behavior.

Examples: `ACGTAC` with `2M2I2M` becomes

```
 ACGTAC   
  │ │ │   
  ▼ │ └─┐ 
 2,2│   ▼ 
  M ▼  6,4
   4,2  M 
    I     
```

`ACGTA` with `3M2D3M` becomes

```
 ACGTAC    
   ││ │    
   ▼│ └──┐ 
 3,3└─┐  ▼ 
  M   │ 6,8
      ▼  M 
     4,6   
      D    
```

Padding should be able to be represented the same as a deletion, but tagged with `P`.

Overall, this would represent a slight trade-off: it's not quite as compact as AnchorGaps or CIGAR, but it's as efficient as either and more general than either.

I wouldn't worry about implementing this yet, I think if we get a basic implementation we can always extend it. I just want to start thinking about how short read alignments will fit in.

Ward9250:
@dcjones I absolutely love this idea of extending AnchorGaps to Cigar notation! (AnchorCigars? .... CigarAnchors?) - I can start coding some of this if you'd like - I guess the efficient thing to do would be like you did with nucleotides - where each kind of nucleotide is represented by a const, so if I have it so as the anchors have references to those consts to denote which cigar operation the anchor applies to?

timbitz:
Hey Guys,
   I began watching this project and thread a while ago, though I have been too busy to really contribute much.  But I do have some experience with the SAM Spec, and will probably use this quite a bit, so I figured I might as well throw in my 2cents.  

Firstly, I like the data structure you have proposed, it makes good sense.  What I have issues with is the SAM spec in general which is written such that in order to find mismatches/polymorphisms, or crosslinking/protocol specific mutagenesis sites within reads, you need to traverse through a CIGAR string and the MD string at the same time, both of which are semi-redundant but neither with comprehensive information.  So a mismatch from the reference is still represented as a valid M by CIGAR, meanwhile there is no way to express an insertion `I`, intron `N`, or clip `S` within the MD string.

The example from the SAM Spec suggests that with the MD string of `10A5^AC6` we can follow the start of alignment on the reference 10bp to a mismatch (where the reference is A) then 5 more to a deletion (where the reference reads AC) then 6 more matches.  The CIGAR should read something like `16M2D6M`, but potentially containing `S`,`N` or `I` as well.  So in order to find the read base that corresponds to the mismatched A, you have to count the `S`, `I` from CIGAR and numeric MD bp in the read sequence to get to it.  Also there is a distinction here between short deletions `D` and large gaps/introns `N` which cannot be represented in the MD string, so for actual reference mismatch position you need to count `N` intron CIGAR bp as well.

My proposal:  Since it is silly to have two semi-redundant data structures representing the information from one alignment.  Perhaps we could instead create an additional version of the proposed julia alignment data structure that represents both CIGAR annotations as well as the relevant mismatch info. 
This would make programming involving both SNPs and indels much more straight forward. Thoughts??

Ward9250:
@timbitz I have to admit I am not familiar with MD strings, but am I right in thinking the difference is in CIGAR, a region of M could contain bases which are identical to the reference, or ones which are polymorphic - which is noted by the MD string. In which case I agree with you, incorporating mismatch information would be useful - I'm an evolutionary biologist and population geneticist by background, so quickly and efficiently accessing and programming with mutational and polymorphic info from sequence alignments is something I am eager to see - for calling alleles/genotypes or building phylogenies and such. For the same reason it is also important to me the gaps/anchor system we use in Align 'can' create multiple sequence alignments (the kind generated by ClustalW) that are agnostic of a reference.

dcjones:
Thanks for the comment @timbitz. Storing mismatch information hadn't occurred to me. I think it would make sense to include an optional field in the alignment representation to store mismatch information.

We could then let mismatch information be accessed either by using stored mismatch info or by computing it against a given reference sequence, like

```julia
function mismatches(alignment, reference)
    # Compute mismatches by comparing the reference and alignment sequence
end

function mismatches(alignment)
    if isnull(alignment.mismatches)
        error("Alignment has no mismatch information.")
    else
        return get(alignment.mismatches)
    end
end
```

I suppose we could store mismatches as an array of `(position, reference nucleotide)` pairs.

timbitz:
@Ward9250 Yes you are correct.  The MD string is a representation of pre-computed mismatches in alignment blocks between read/reference in SAM, be it polymorphisms or even out-right sequencing errors. Meanwhile CIGAR just annotates any non-gap block as `M`.  You are right that @dcjones did mention `=` and `X` which are present in the CIGAR portion of the SAM Spec, however to date I have not seen a short-read alignment program which makes use of these.  Even if they did, just knowing that there is a base at some position _ which does not match the read (`X` in CIGAR) would require look up in the original sequences to determine the exact transition/transversion, which for large genomic references might eventually get expensive.

You are also right that multiple sequence alignment is an important consideration, and I haven't put much thought into this.  Mismatch data gets more complicated at this point and probably would need to be computed in julia as @dcjones suggested.   So I guess ideally it would be awesome to have a specification that allows easy computing of mismatch sites/rates for multiple alignments or loading of pre-computed mismatch info for pairwise alignments when available as-in MD strings.

@dcjones Yah I suppose a simple array of tuples would do.  I am tempted to suggest extending CIGAR with sub-classes or additional single letter codes beyond `X` that define all dinucleotide transition/transverion or single deletion pairs.  This obviously wouldn't help for amino-acid or other alphabets, and might be weird with multiple alignments, but could be useful for short-read data specifically?

Ward9250:
Do you think it would be possible to extend a mismatch concept to have precomputed mismatches for multiple alignments (storing data for every sequence pair in the MSA for example) - if we found some way for CIGAR to work for multiple alignments - through padding or other means?

dcjones:
>  I am tempted to suggest extending CIGAR with sub-classes or additional single letter codes beyond X that define all dinucleotide transition/transverion or single deletion pairs.

That's an interesting idea. We would already be storing CIGAR operations in bytes. There are 9 operations to encode in standard CIGAR, so that would leave us with 247 values to play with. Encoding a reference nucleotide or amino acid would require at most 5 + 21 = 26 values if you allow `N` or `X`, so that seems doable. On the other hand, encoding pairs of nucleotides or amino acids would be 5^2 + 21^2 = 466, so we'd need another byte, but do we really need the query sequence nucleotide? Can't we always look that up, or are you thinking of formats like [goby](http://campagnelab.org/software/goby/) where the query sequence isn't stored?

> Do you think it would be possible to extend a mismatch concept to have precomputed mismatches for multiple alignments (storing data for every sequence pair in the MSA for example) - if we found some way for CIGAR to work for multiple alignments - through padding or other means?

I don't think we'd have to store mismatches for every pair of sequences. If one was chosen arbitrarily as the "reference" and then mismatches for all the others computed against that, I think you could reconstruct the pairwise mismatches.

Ward9250:
Just a question - say we make many types of anchor that all inherit from and abstract type, so a mismatch anchor, insertion anchor etc. and they all have the two integer fields corresponding to alignment and source space. Then you store them as an array Array{abstract anchor} - will Julia recognise the anchors all have two Int fields and store them compactly - or must they be of the same type regardless of their contents?

dcjones:
Unfortunately, it won't.
```julia
abstract A
immutable B <: A x::Uint8 end

julia> sizeof(A[B(1), B(2), B(3), B(4)])
32

julia> sizeof(B[B(1), B(2), B(3), B(4)])
4
```
So it would be more efficient to represent different types of anchors manually rather than rely on the type system.

Ward9250:
Perhaps it's worth considering - whilst we're on the topic of SNPs / polymorphisms with Alignments. How to efficiently store and program with polymorphism data in the light of formats like VFC, SAM Pileup and so on. Feels like a separate module or PR, but in my background, multiple alignments reveal the variants and indwells so the two topics, alignment and variation are connected somewhat.

src/align/anchors.jl
```diff
@@ -0,0 +1,106 @@
...
+----------------
+Gap-Space position: $(anc.gapPos)
+Source-space position: $(anc.seqPos)
+Alignment operation: $(anc.op)")
+end
+
+
+function copy(src::AlignmentAnchor)
+  return GapAnchor(src.gapPos, src.seqPos, src.op)
+end
```
bicycle1885: A typo of `AlignmentAnchor`?

Ward9250:
@bicycle1885 Thanks Kenta, missed the throwback from when I called them GapAnchors - renamed to AlignmentAnchors when I allowed for them to store their kind of operation. What was a GapAnchor is now `AlignmentAnchor(x, y, '-')`

Ward9250:
@dcjones I've added Operations types and made it so as GapAnchors are now AlignmentAnchors that contain Operations. Take a look at operations.jl and anchors.jl. I think the next thing that is needed is efficient sorting of an array of anchors - in Seqan iirc (which is why I ask you, as you've also taken a look at the Seqan codebase), both converting between source and gap space, and inserting anchors, requires sorting and the use of the stl::upper_bound and stl::lower_bound functions.

These two stl algorithms are used with two different lt functors - one that makes it so as upper_bound and lower_Bound sort and consider lt according to the gap space position, and one so as they sort and consider lt according to the sequence / source space position. They use these two functors because the canonical lt for gapAnchors as they define them returns true if a.gapspace < b.gapspace OR a.seqspace < b.seqspace . This is my understanding from reading the code - am I correct?

If so - I made a post on julia-users [here](https://groups.google.com/forum/#!topic/julia-users/ZcH22bhr0EA). Showing my experimentation with sorting. It seems the Julia equivalents of upper_bound and lower_bound are findsortedfirst, and findsortedlast. However, since in seqan there are two lt functors, my first and obvious solution is to simply feed two different functions to findsortedfirst and findsortedlast as the lt=argument. But as is shown in the julia-users thread, passing custom functions to those algorithms reduces their speed somewhat. It is a big goal to me that sorting and searching these arrays of anchors is as fast as possible. It seems there is an improvement coming that will improve performance when passing custom functions, but I wondered if it was worthwhile to define our own set of upper_bound and lower_bound methods for these arrays, essentially the exact same code as searchsortedfirst and searchsortedlast, but with the two different kinds of lt coded in? So there would be a upper_gap_bound, lower_gap_bound, upper_source_bound, and lower_source_bound. If it's hard coded in the algorithm I guess it will be more efficient than passing in a function.

What do you think? I'm happy to take thoughts on this - efficient sorting is a bit new to me. With many short read alignments or long alignments with many gaps/operations we'll need searching these arrays to be as fast as possible.

dcjones:
I know more efficient anonymous functions are a long term goal, but it's not something we should wait on, since I don't think there is an ETA yet.

I threw together a benchmark, and it seems like our best bet would be do define our own versions of `searchsorted`.

```julia
# Stand-in for anchor gaps
immutable T
    a::Int
    b::Int

    T() = new(rand(Int), rand(Int))
end

const ts = [T() for _ in 1:1000000]
sort!(ts, lt=(u, v) -> u.a < v.a)

# passing a function to `by`
m = 100000
println("using searchsorted `by`")
@time for _ in 1:m
    searchsorted(ts, T(), by=u -> u.a)
end

# passing a function to `lt`
println("using searchsorted `lt`")
@time for _ in 1:m
    searchsorted(ts, T(), lt=(u, v) -> u.a < v.a)
end

# defining a fixed `isless`
Base.isless(u::T, v::T) = u.a < v.a
println("fixed `isless` function")
@time for _ in 1:m
    searchsorted(ts, T())
end

# special purpose searchsorted (copied from Base)
function custom_searchsorted(v::Vector{T}, x::T)
    lo = 0
    hi = length(v) + 1
    @inbounds while lo < hi-1
        m = (lo+hi)>>>1
        #if lt(o, v[m], x)
        if v[m].a < x.a
            lo = m
        else
            hi = m
        end
    end
    return hi
end

println("custom searhsorted:")
@time for _ in 1:m
    custom_searchsorted(ts, T())
end
```
```
using searchsorted `by`
 403.174 milliseconds (12839 k allocations: 301 MB, 2.55% gc time)
using searchsorted `lt`
 403.664 milliseconds (12838 k allocations: 301 MB, 2.58% gc time)
fixed `isless` function
  70.458 milliseconds (508 k allocations: 12921 KB, 1.17% gc time)
custom searhsorted:
  45.196 milliseconds (299 k allocations: 6234 KB)
```

So a custom binary search function is even faster than having a fixed `isless`. (This was actually a little surprising to me, I may do this in IntervalTrees to see if I get a speed up.)

dcjones:
Also, if you really want this to be as fast as possible, if the array of anchors is relatively short, it's actually slightly faster to search through them linearly. So you could have two versions and decide which to use based on the size of the array.

```julia
const ts = [T() for _ in 1:20]
sort!(ts, lt=(u, v) -> u.a < v.a)

m = 1000000

println("custom searchsorted:")
@time for _ in 1:m
    custom_searchsorted(ts, T())
end

function linear_searchsorted(v::Vector{T}, x::T)
    @inbounds for i in 1:length(v)
        if v[i].a > x.a
            return i
        end
    end
    return length(v) + 1
end

println("linear searchsorted:")
@time for _ in 1:m
    linear_searchsorted(ts, T())
end
```

```
custom searchsorted:
 221.204 milliseconds (2999 k allocations: 62484 KB, 15.33% gc time)
linear searchsorted:
 175.635 milliseconds (2999 k allocations: 62484 KB, 1.50% gc time)
```

Ward9250:
I guess we can run some dummy simulations and try and find roughly where the tradeoff occurs with linear vs search sorted, and then include some sensible default. If dealing with short-reads I guess for a lot of the time, the array will be somewhat short. With longer reads and genomic alignments the array may be much longer. I suppose it may also depend on the kind of information encoded in an array - if you only have a few operations, say it's an MSA that only uses Gaps operations, then the array would contain less elements than if it also contained more detailed info like matches/mismatches, insertions, and deletions, relative to a reference.

Ward9250:
After another [chat on julia-users](https://groups.google.com/forum/#!topic/julia-users/OEXxlaeZFoU) it seems that a viable option is to create a few functors that manipulate the outcome of the base versions of searchsortedfirst and searchsortedlast.

Ward9250:
Forget functors - by creating singleton types that inherit from Base.Order.Ordering and adding methods to Base.Order.lt we can now sort and searchsorted* for Vectors of AlignmentAnchor efficiently. Take a look at the bottom of anchors.jl to see how I've done this.

Ward9250:
So with sorting sorted (*snort*) (if you'd like I can post up here an example and timings) - I would like to get insertion, deletion, and translation between sourcePosition and alignmentPosition. With that in mind a quick question / consultation with everyone about clipping and cuts and view cuts. Seqan alignment objects contain the array of anchors, and four other values - cutBegin, cutEnd, viewCutBegin, and viewCutEnd. 

I believe these four values are essentially analogous to hard clipping and soft clipping. If I am correct in my understanding that hard clipped bases do not appear in <seq> whereas soft clippings do, however, soft clippings do, but are not used in calculating coverage and do not appear in a viewer.

Some examples are that `10H10M10H` SEQ is 10 bases long, and for `10S10M10S` SEQ is 30 bases long. 

My question then to everyone is - do we do away with cutBegin, cutEnd, viewCutBegin, and viewCutEnd and instead use only hard clipping and soft clipping. I ask because I am familiar with the context and use of cutBegin, cutEnd, viewCutBegin, and viewCutEnd in sedan, which only has one kind of Anchor, the Gap '-' anchor: with those four values is is possible to represent a local alignment - a case where two sequences align only in a given domain or motif or region of the genome/gene.

My knowledge and use of CIGAR thus far has been limited, but I wonder if we could do away with those four values as used in seqan in favour of just using hard and soft clipping. A hard/soft clipping at either ends of an array of anchors could potentially represent such cut and view cut values. In addition an advantage is soft/hard clippings in the middle or throughout the array could be used to denote introns in mRNA, or if the context of the alignment is MSA then multiple locally aligned domains and so forth. I tentatively suggest that in the domain of pairwise/multiple sequence alignment, and in the domain to alignment of reads to reference - we can probably do without the four view and cut values if we have the clipping from CIGAR available.

Ward9250:
To update this thread with progress, I'm close to having code done converting between the alignment view positions, and the source sequence positions, and vice versa. Essential for a lot of tasks with this data structure. I've come up with a neater version of what sedan does, and then a much more concise and clear bit of code that achieves that same thing, and is more obviously robust to edge cases, but it does use two binary searches rather than seqan's one search, which I'm not all that happy about. So I'll probably go with the single binary search solution and we just keep testing it to make sure it holds.

bicycle1885:
Great! I'm now working on pairwise alignment algorithm in my repository (https://github.com/bicycle1885/Bio.jl/tree/pair-align).
To align short reads to a genome sequence, I will need something like range-bounded pairwise alignment though.
I need an alignment data structure to return the result of alignment.

wildgarden:
Hi,
this is great news. I'm going to implement (in the next days) a multiple sequence alignment algorithm in my repository (https://github.com/wildgarden/Bio.jl)
@bicycle1885 At first glance your pairwise alignment looks like Needleman-Wunsch. Does it have any modifications?

bicycle1885:
@wildgarden sounds great! We definitely need multiple sequence alignment implementation available in Julia.
One possible way to achieve that is writing a wrapper for other multiple alignment library.
For example, Clustal Omega (http://www.clustal.org/omega/) provides the API for the C language.
I don't know the state of the art about multiple alignment, but Clustal Omega is widely used in bioinformatics.

> At first glance your pairwise alignment looks like Needleman-Wunsch. Does it have any modifications?

Yes, it includes the Needleman-Wunsch algorithm. It also has a modified version by [Ukkonen](http://www.sciencedirect.com/science/article/pii/S0019995885800462), which is asymptotically faster than the NW algorithm for similar two sequences.
These will be used in my short read aligner implementation written in Julia as a subroutine.
I'm also going to implement some bitparallel algorithm and other variants that are efficient in terms of space.

Ward9250:
Glad to see this looks useful to other people. When the co-ordinate translation is done I can work on insertion and deletion of alignment operations too, getting this PR closer to a higher level API for building and managing alignments of several types, all of which use this underlying system. I've noticed that there is some benefit to having all of the operations encoded, unlike how sedan uses anchors, which only represents the gaps. A lot of checking of values and boundaries becomes irrelevant. For example, in seqan you could try to convert an alignment position to src position. If the alignment position is towards the end of the sequence, and there are no gaps nearby, you're indexing outside of the anchors array and sedan has a lot of conditionals and workarounds for such little details and edge cases. In our system, that area *is* would be encoded by a big M anchor. In other words the full source sequence is covered by the anchors, which means fewer edge cases that require thinking about.

wildgarden:
@bicycle1885 I had a look at your branch 'pair-align'. I'm very impressed and would like to extend your work in my multiple-alignment branch. As a warmup I wanted to start with the center-star-method, which relies on Needleman-Wunsch to compare two sequences. However, I suppose you forgot to push the file "substition_matrix.jl". Could you please upload it?

Regarding 'Clustal-Omega': I like the idea to implement a wrapper for the C-API. But before I start from scratch: Are their already some specifications or ideas  for the use of external applications in BioJulia that I should pay attention ?

bicycle1885:
@wildgarden Oh, sorry. I've pushed that file and now you can see that.
I don't know so much about multiple-alignment implementations, but Clustal Omega is a definitely high-quality and performant alignment implementation in my experience. You can see the paper here: http://msb.embopress.org/content/7/1/539.
I think @Ward9250 knows much about specs or formats about alignment data of multiple sequences.

Ward9250:
We talked about using external applications before. I don't see a specific issue on it. Perhaps we can open an RFC issue on it just to get comments on record. When we first started BioJulia we wanted - in the ideal world to go as far as possible with pure Julia implementations of all algorithms. Of course installation of Bio.jl might become more complicated - requiring compiled components and use of BinDeps and Homebrew.jl, one suggestion has been a completely separate module to Bio.jl, but that requires Bio.jl, which includes extra functionality that may not be suitable for Bio.jl, bundles of external applications being one thing suitable for such an extra module. I've opened a Request For Comment issue here #77.

I guess that the C-API would be the way to go with cluster-omega. I had a quick look at some example code using it, they pass in a filename. I don't know if it would be possible to modify the DNA Sequence structures we have and pass them to the clustal code, without having to write to temporary file first.

Ward9250:
Can someone please take a look at the src/align/cigar.jl file in the branch SequenceAlignments on my Bio.jl fork and tell me why on earth the show method I define for Array{CIGAR, 1} - last function in the file, isn't used, instead the method for generic arrays is used? It's starting to drive me mad - I import the base show method!

breichholf:
What are the plans on reading SAM and BAM files? I've just recently started using Julia, and after going over @kmsquire's Sam package and seeing some gaps I was going to plug those holes and perhaps restructure his code somewhat. Might also be worth thinking about including that somewhere into Bio.jl?

dcjones:
I think plan for SAM/BAM is to get the work on this branch merged, then write a new parser using ragel, similar to how the parsers for fasta, fastq, and bed work. I've also started writing [new zlib bindings](https://github.com/dcjones/Libz.jl), partially with BAM parsing in mind.

Ward9250:
Indeed it is my aim to have the Align module user ready sooner rather than later, because it's useful for a variety of things - SAM, Phylogenetics, and so on. Although you may be able to start work with SAM related stuff if just the intermediary CIGARString structure is complete, which is basically is, as it's just an array of a simple immutable type, which can be built using strings.

Ward9250:
Speaking of Ragel.jl, how far are we from a parser API? It would speed development a lot if we could all understand how to make and understand the Ragel in Julia stuff. For example I do wonder if it would be possible to parse some phylogenetic formats and I'd like to have a crack at it, also for PopGenetics formats too.

dcjones:
I need to merge #51, since that has some changes to the api. It's "done" but needs more tests before merging. After than I'll documenting the parser framework a priority.

breichholf:
Cool, well seeing as my work heavily requires SAM/BAM parsing I'd gladly help out on that end.

prcastro:
Since we are aiming for Julia v0.4, we can omit `Docile.@doc` and `->` from docstrings.

src/align/operations.jl
```diff
@@ -0,0 +1,100 @@
...
+  OP_INSERT, OP_INVALID, OP_INVALID, OP_INVALID, OP_MM,
+  OP_N, OP_INVALID, OP_PAD, OP_INVALID, OP_INVALID,
+  OP_SCLIP, OP_INVALID, OP_INVALID, OP_INVALID, OP_INVALID,
+  OP_MISMATCH
+]
+
+
+function convert(::Type{Operation}, c::Char)
+  @inbounds op = '-' <= c <= 'x' ? char_to_op[c - '-' + 1] : OP_INVALID
+  @assert op != OP_INVALID error("$(c) is not a valid alignment operation.")
```
prcastro: This raises an assertion error, that can be confusing to users. What about raising an `error` inside of an `if` clause?

src/align/align.jl
```diff
@@ -0,0 +1,24 @@
+module Align
+
+using Compat
+using Docile
+using Docile: @doc, @doc_str
```
prcastro: We don't need to use Docile here if we use Julia's docstrings. Docile functionality don't depend on this `using` statements (we can use Docile even on modules that weren't meant to use it).

src/align/align.jl
```diff
@@ -0,0 +1,24 @@
+module Align
+
+using Compat
```
prcastro: Why use Compat if we require Julia v0.4?

Ward9250:
In my sequence alignment feature branch I don't think I use Compat in the Align module for this reason.

Ward9250:
@prcastro Thanks for the readthrough and suggestions! You are quite right about the docstrings and Compat - I think I largely copied and pasted how other modules were set out at the time of starting this branch, so those two things are artifacts I'll remove shortly. Regarding the assertion error. You are right, a more informative error or throw in an if block would be better. I'll change this too shortly.

src/align/anchors.jl
```diff
@@ -0,0 +1,251 @@
...
+integer fields to 0, and the operation field to the global constant OP_INVALID.
+""" ->
+immutable AlignmentAnchor
+  alnPos::Int
+  srcPos::Int
+  op::Operation
+  function AlignmentAnchor(gp::Int = 0, sp::Int = 0, op::Operation = OP_INVALID)
+    return new(gp, sp, op)
+  end
+end
```
bicycle1885: Indent with four spaces is our style :smiley:

Ward9250:
Oh gosh! Of course it is! I switched to a new text editor right before this PR and didn't think to change the settings. I'll rectify, thanks!

Blahah:
We could avoid some of these formatting issues by providing an [editorconfig](http://editorconfig.org/) file in root, then reminding people to install the plugin for their editor in the contributing guide. I'll make a separate issue.

Ward9250:
Great Idea! The Atom editor I use has a package called atom-editorconfig.

bicycle1885:
@Ward9250 Is there any news in this pull request? I'd like to use the CIGAR string and other alignment types to return the result of alignment I'm implementing now.
If you need a help, I will have time :)

Ward9250:
@bicycle1885 So we have types for CIGAR strings, they're pretty simple and ready for use. With the AlignmentAnchors alignment format specified in the above thread, we have the type, we have methods for sorting and searching efficiently through the data, and we have a method for finding the anchors that straddle a given position in an alignment. We have a method for converting an alignment view position, to a source view position. I started a method for converting a source view position to an alignment view position, but didn't have time to get round to finishing it. Those three methods are essential however for many of the other operations someone would want to do when working with a datastructure containing an alignment(s). What other alignment types are you using in your aligner? Help is always appreciated. Between thesis, Phylo, this, and some very early code for SNPs and genetic variant data, I'm getting a bit thin. Thankfully a good Phylo V1 isn't far from completition. There's only the question of annotations left - oh and the testing suite.

dcjones:
I'm going to need this as well to finish BAM support (#101). I can work this week to get it ready to merge, if you're busy.

bicycle1885:
I'm going to need the CIGAR string and gapped sequences (`ArrayGaps`?).
Once the best short-read alignment location has been found, two gapped sequences that represent a pairwise alignment are constructed along with tracing back the dynamic programming matrix.
Then, a CIGAR string and a MD string are computed using the alignment data.

Let me comment a few about the current code:

* Some camelCase names of local variables and function names don't look like a Julian way; I guess these are taken from the C++ code but snake_case would be more suitable.
* The CIGAR string type (`CIGARS`) as an alias of `Vector{CIGAR}` may be inflexible in Julia. This is because multiple dispatching doesn't work for type aliases. Instead, introducing a new type of the CIGAR string would be appropriate like:

```julia
type CIGAR
    ops::Vector{Operation}
    lens::Vector{Int}
end
```

* The `source` field of the `ArrayGaps` type can be parameterized by a biological sequence type:

```julia
type ArrayGaps{S<:Bio.Seq.Sequence} <: Gaps
    source::S
    ...
```

But we may want to use an array of elements (e.g. `Vector{DNANucleotide}`) as a biological sequence, the type parameter restriction (`S<:Bio.Seq.Sequence`) may be too strict.

Ward9250:
@bicycle1885 Thanks for the comments, I'm just about to rebase this so it's up to date with upstream, and then implement your comments. For gapped sequences you have a choice - ArrayGaps as you say are useful for building gapped sequences, because they are simply an array of numbers, which alternate between he size of a chunk of sequence, and the size of a chunk of gap - so extending them is easy, but indexing into them is not efficient. For indexing purposes and other operations - conversion from gap to source space and so forth, the AlignmentAnchors I've been writing are better - inexing is faster as it's performed via binary search. We've extended the AlignmentAnchors concept of Seqan too so instead of just representing gaps, they represent any number of alignment operations (which means they can still of course just represent only gaps - but they can do more, and I think the richer information they carry will help in all sorts of operations).

Ward9250:
@bicycle1885 BTW it occurs to me, the CIGAR string type, as implemented is actually an extension of ArrayGaps, in a similar way that our AlignmentAnchors are an extension of seqan's alignment anchors. So there's even argument for making ArrayGaps redundant, and moving that code over to CIGAR Strings. Either way, if you describe the kind of operations you want and need for your alignment algorithm, I'll see how close we are to those operations with these types and draw up an actionable list of TODO's I can make a priority, or other can fork, implement and merge.

Ward9250:
I'm going to make a few changes tonight, based on your comments, to make the implementation more general - for example, should the underlying code types really be called with respect to the CIGAR way of representing alignments, when really what it should be talking about is operations - CIGAR being one way of denoting alignment operations.

bicycle1885:
@Ward9250 Sorry, I cannot catch up your points. I may misunderstand something.

> BTW it occurs to me, the CIGAR string type, as implemented is actually an extension of ArrayGaps

I think the CIGAR string is just a description of pairwise sequence alignment between two sequences and cannot be a kind of gapped sequences.
In my prototype of pairwise alignment module, a pair of gapped sequences is the result of pairwise sequence alignment and a CIGAR string can be derived from this pair.
This is my prototype implementation of the gapped sequence alignment type corresponding to `ArrayGaps`: https://github.com/bicycle1885/PairwiseAlignment.jl/blob/master/src/gapped_sequence.jl.
And this is the result type of pairwise alignment: https://github.com/bicycle1885/PairwiseAlignment.jl/blob/master/src/result.jl.
This package will be rebased on the Bio.Align module and I'm willing to use your implementations instead of my gapped sequence type.

Ward9250:
Closing, as this was continued and improved on in the redux PR, which has since merged.

PR #43, opened 2015-04-19
add parse method for amino acids

bicycle1885:
We often use three-letter abbreviations to represent amino acids.
For example, Protein Data Bank (PDB) format adopts these three-letter encoding as its standard (http://www.wwpdb.org/documentation/file-format-content/format33/sect3.html#SEQRES).

This pull request implements a parser for one-letter and three-letter abbreviations which is case-insensitive and ignores white space.

### DIFF
--- a/src/seq/aminoacid.jl
+++ b/src/seq/aminoacid.jl
@@ -81,7 +81,29 @@ function show(io::IO, aa::AminoAcid)
     end
 end
 
+const threeletter_to_aa = @compat Dict(
+    "ALA" => AA_A, "ARG" => AA_R, "ASN" => AA_N, "ASP" => AA_D, "CYS" => AA_C,
+    "GLN" => AA_Q, "GLU" => AA_E, "GLY" => AA_G, "HIS" => AA_H, "ILE" => AA_I,
+    "LEU" => AA_L, "LYS" => AA_K, "MET" => AA_M, "PHE" => AA_F, "PRO" => AA_P,
+    "SER" => AA_S, "THR" => AA_T, "TRP" => AA_W, "TYR" => AA_Y, "VAL" => AA_V,
+)
 
+function Base.parse(::Type{AminoAcid}, s::String)
+    s′ = strip(s)
+    if length(s′) == 1
+        return convert(AminoAcid, s′[1])
+    end
+    try
+        return threeletter_to_aa[uppercase(s′)]
+    catch ex
+        if isa(ex, KeyError)
+            error("invalid amino acid string: \"$s\"")
+        end
+        rethrow()
+    end
+end
+
+# TODO: tryparse
 
 # Amino Acids Sequences
 # =====================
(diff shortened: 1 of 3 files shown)


brk00:
Nice! It would be awesome if we had some tests for the new feature.

bicycle1885:
Finished :checkered_flag:

Could you check my code?

src/seq/aminoacid.jl
```diff
@@ -81,7 +81,30 @@ function show(io::IO, aa::AminoAcid)
...
 
+# lookup table of 20 standard amino acids
+const threeletter_to_aa = @compat Dict(
+    "ALA" => AA_A, "ARG" => AA_R, "ASN" => AA_N, "ASP" => AA_D, "CYS" => AA_C,
+    "GLN" => AA_Q, "GLU" => AA_E, "GLY" => AA_G, "HIS" => AA_H, "ILE" => AA_I,
+    "LEU" => AA_L, "LYS" => AA_K, "MET" => AA_M, "PHE" => AA_F, "PRO" => AA_P,
+    "SER" => AA_S, "THR" => AA_T, "TRP" => AA_W, "TYR" => AA_Y, "VAL" => AA_V,
+)
+
+function Base.parse(::Type{AminoAcid}, s::String)
```
brk00: Since you are trying to extend `parse` from Base, I believe you should import this function on `seq.jl` and only write `function parse(::Type{AminoAcid}, s::String)`. That's the way we do it when extending other functions from Base.

src/seq/aminoacid.jl
```diff
@@ -81,7 +81,30 @@ function show(io::IO, aa::AminoAcid)
...
+    if length(s′) == 1
+        return convert(AminoAcid, s′[1])
+    end
+    try
+        return threeletter_to_aa[uppercase(s′)]
+    catch ex
+        if isa(ex, KeyError)
+            error("invalid amino acid string: \"$s\"")
+        end
+        rethrow()
```
brk00: There should be a test that fall into this line: some error thrown at line 98 that isn't a KeyError. Maybe if `uppercase` could fail for a value of s'.

src/seq/aminoacid.jl
bicycle1885: Exactly, I forgot that Base of v0.3 has the same name function. Fixed.

src/seq/aminoacid.jl
bicycle1885: I'm afraid that there is no input that causes an error in the `uppercase` function.
Chinese characters and even emoji won't cause any errors.
Possible exceptions are `MemoryError` and `InterruptException`, but these are hard to reproduce in tests.

dcjones: I can't see any way to reliably trigger an exception other than `KeyError` here, yet the `rethrow()` is good practice. I guess we'll have to live with having an untested line. Eventually we'll be able to mark code as irrelevant for coverage, I hope.

dcjones:
This looks good to me. Thanks @bicycle1885 (and @brk00 for enforcing testing).

PR #47, opened 2015-04-22
[Seq] More documentation

brk00:
Start writing more documentation to Seq module

### DIFF
--- a/src/seq/nucleotide.jl
+++ b/src/seq/nucleotide.jl
@@ -66,7 +66,7 @@ const RNA_C = convert(RNANucleotide, 0b001)
 @doc doc"RNA Guanine" ->
 const RNA_G = convert(RNANucleotide, 0b010)
 
-@doc doc"RNA Uracil"
+@doc doc"RNA Uracil" ->
 const RNA_U = convert(RNANucleotide, 0b011)
 
 @doc doc"Any RNA Nucleotide" ->
--- a/src/seq/nucleotide.jl
+++ b/src/seq/nucleotide.jl
@@ -266,7 +266,7 @@ end
 @doc doc"""
 `repeat(chunk::NucleotideSequence, n)`
 
-Construct a nucleotide sequence by repeating another sequence `n` times" ->
+Construct a nucleotide sequence by repeating another sequence `n` times""" ->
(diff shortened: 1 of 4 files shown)


brk00:
Ready to merge. I'll try to add more docs later.

src/seq/nucleotide.jl
```diff
@@ -966,17 +1118,27 @@ immutable EachKmerIteratorState{T, K}
     nit_state::Int
 end
 
+# Maybe this function should replace the default constructor.
+# Is the (unsafe) default constructor used throughout our code?
```
dcjones: Is this comment misplaced? I'm not sure what it's referring to.

dcjones:
Looks good to me.

One thing we should consider: with newer versions of Docile, there's a macro that let's you drop the `@doc` and `->`, and just put strings before declarations: http://docilejl.readthedocs.org/en/latest/syntax/#plain-strings

brk00:
Nice, I am expecting something like this for a while now. But I think I prefer to stick with Julia's v0.4 syntax for the time. I believe we will have the final syntax soon, so we will need to update it just once.

src/seq/nucleotide.jl
brk00: I'm referrint to `eachkmer`. This is could replace the default constructor of `EachKmerIterator` , since it checks `k` and `step`.

src/seq/nucleotide.jl
```diff
@@ -600,30 +693,33 @@ end
...
+
+Return the number of mismatches between `a` and `b`.
+
+If `a` and `b` are of differing lengths, only the first `min(length(a), length(b))`
+nucleotides are compared.
+
+# Arguments
+* `a`: first sequence to compare
+* `b`: second sequence to compare
+* `nbatches`: if true, N matches anything, if false, N matches only itself.
```
kdmurray91: Hi, excuse the random comment. Is a `s/nbatches/nmatches/` required here?

brk00: Random comments are welcome! Specially when they pont out typos and inconsistencies.

n**m**atches isn't required, and the docs should say so.

dcjones:
:sun_with_face:

Issue #46, opened 2015-04-22
Module files should start with uppercase


prcastro:
Following conventions in other Julia packages, I propose we rename all module files and folders to start with uppercase (e.g. `src/seq/seq.jl` to `src/Seq/Seq.jl`).

dcjones:
This is probably a good idea, though I'm slightly dreading rebasing on that change on my silly case insensitive filesystem.

Blahah:
Can we build this and any other similar issues into a styleguide?

brk00:
That's a great idea

Ward9250:
Has this been done and ready to close?

dcjones:
This is #63, which hasn't been merged. I want to selfishly request that this waits until #51 is merged, because I'm afraid it will be a pain to rebase on this. :smile:

Ward9250:
That's fine by me! :) I just took a brief look at the bigbed branch - a lot to read and understand at first glance, but I like the look of the file buffer code in src/ . *IF* I understand it correctly, it can either contain a memory mapped Mmap.mmap, or a file stream, and incrementally read in data to a buffer. This strikes me as an incredibly useful thing for other areas too - I imagine a butt load of population genetics data from simulation for example, or big tables of polymorphism data.

dcjones:
Yeah the intent is to abstract over reading all at once, incrementally, or mmaping and use that for all the parsers. I'd like to add support for reading from http too.

prcastro:
Solved by #97

PR #49, opened 2015-04-24
Add a basic CONTRIBUTING.md

dcjones:
Adds a CONTRIBUTING.md file with a basic style guide, etc. See #46.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,10 +47,12 @@ considered it must meet the following guidelines.
 
     ```julia
     module AwesomeFeatures
-        using Compat, JSON
 
-        include("feature1.jl")
-        include("feature2.jl")
+    using Compat, JSON
+
+    include("feature1.jl")
+    include("feature2.jl")
+
     end
     ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,65 @@
+
+# Contributing
+
+We welcome contributions in the form of pull requests. For your code to be
+considered for inclusion it must be documented.
+
+  * By making a pull request, you're agreeing to license your code under an MIT
+    license. See LICENSE.md.
+
+  * Types and functions must be documented using
+    [Docile](https://github.com/MichaelHatherly/Docile.jl) style docstring.
+    Documentation regarding specific implementation details that aren't relevent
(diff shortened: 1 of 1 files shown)


brk00:
Nice =)

CONTRIBUTING.md
```diff
@@ -0,0 +1,66 @@
...
+    be included from seperate files. E.g.
+
+    ```julia
+    module AwesomeFeatures
+        using Compat, JSON
+
+        include("feature1.jl")
+        include("feature2.jl")
+    end
+    ```
```
bicycle1885: Is indentation recommended within a `module` block?

CONTRIBUTING.md
```diff
@@ -0,0 +1,66 @@
...
+
+  * Separate logical blocks of code with one blank line, or two blank lines for
+    function/type definitions.
+
+  * When extending method definitions, explicitly import the method.
+
+    ```julia
+    import Base: start, next done
```
bicycle1885: a comma is missing after the `next`

CONTRIBUTING.md
brk00: That's not used throughout our code, and not recommended by Julia documentation.

bicycle1885: Yes. So I thought this may be a misleading example.

dcjones: My mistake. It's fixed now.

bicycle1885:
What about `return` of function definition?

The current code seems to follow the rule:
* Do not omit `return` keywords
* except one-line definition.

dcjones:
I think that's open to debate. I personally like explicit returns and do it out of habit now. It's only one extra word to type, and it makes it immediately obvious what a function is returning.

Ward9250:
I agree I like explicit returns. If it's not too much effort I always prefer code to be explicit.

bicycle1885:
I like omitting final returns but that's just my taste.
I'm not against explicit return convention.

Ward9250:
Should we also say - we adhere to the Julia community guidelines (http://julialang.org/community/standards/) in addition to any we think of?

Blahah:
@Ward9250 absolutely, good idea.

dcjones:
Merged. Feel free to make PRs to amend.

PR #50, opened 2015-04-25
WIP: BED Parser

dcjones:
Here's a parser for BED format. It's not really tested yet. I need to collect BED specimens and build test cases, so there are almost certainly still bugs. Note: all the commits are because this is branched from by `intervals` branch, which is branched from my `io` branch. So don't be alarmed. :smile: 

The one file I tested it on had 10k 6-column entries. That parsed in 0.016 seconds, so perf seems pretty solid.

By the way, one thing I'm pretty excited about: if you have two sorted BED files (or eventually any kind of file that stores intervals), you can iterate over intersections without even reading either into memory like:

```julia
intersect(read("file1.bed", BED), read("file2.bed", BED))
```


Ward9250:
Had a look at the file, looks nice and is very readable until what I guess is Ragel generated stuff again, but then we have the .rl fines so s'all good :)

src/intervals/bed.rl
```diff
@@ -0,0 +1,278 @@
...
+    newline      = '\r'? '\n'     >count_line;
+    hspace       = [ \t\v];
+    blankline    = hspace* newline;
+
+    seqname      = [ -~]*   >pushmark     %seqname;
+    first        = digit+   >pushmark     %first;
+    last         = digit+   >pushmark     %last;
+    name         = [ -~]*   >pushmark     %name;
+    score        = digit+   >pushmark     %score;
+    strand       = [+\-]    >strand;
```
bicycle1885: This may be too strict to parse a wild BED file.
The [*rtracklayer*](http://www.bioconductor.org/packages/release/bioc/html/rtracklayer.html) package of Bioconductor can read `*` and `.` strand.

bicycle1885:
There will be open problem about strand notation and its meaning.
BED format only allows either `+` or `-` strand (https://genome.ucsc.edu/FAQ/FAQformat.html#format1).
`+` (forward) and `-` (reverse) strand are not controversial, but we often see `.` strand notation in BED files (related BioStar's question: https://www.biostars.org/p/73944/).
Moreover, *GenomicRanges* package of Bioconductor uses `*` strand to denote NA values though its semantic is not stated clearly.
So I think the BED parser should allow `.` strand notation.

BTW, GFF3 spec specifies the meaning of these notations:

> Column 7: "strand"
The strand of the feature. + for positive strand (relative to the landmark), - for minus strand, and . for features that are not stranded. In addition, ? can be used for features whose strandedness is relevant, but unknown.
http://www.sequenceontology.org/gff3.shtml

dcjones:
It does sound like a good idea to relax the strand parsing in BED files. I do have a `STRAND_POS`, `STRAND_NEG`, `STRAND_BOTH`, and `STRAND_NA` constants defined which capture the meanings in GFF3 (`+`, `-`, `.`, and `?` respectively). I suppose we should allow `.` and `?` in BED as well.

src/intervals/interval.jl
```diff
@@ -0,0 +1,176 @@
...
+
+
+const STRAND_NA   = convert(Strand, 0b000)
+const STRAND_POS  = convert(Strand, 0b001)
+const STRAND_NEG  = convert(Strand, 0b010)
+const STRAND_BOTH = convert(Strand, 0b011)
+
+function Base.show(io::IO, strand::Strand)
+    if strand == STRAND_NA
+        print(io, "(indeterminate strand)")
```
bicycle1885: If we stick to the convention of GFF3, showing `"?"` instead of `"(indeterminate strand)"` would be more concise and not disturb alignment of texts.

dcjones:
Updated with your suggestions. Thanks!

bicycle1885:
Looks good! I'll continue to review some more.

dcjones:
```julia
using Bio.Intervals
function f()
    for i in intersect(read("trya.bed", BED), read("tryb.bed", BED))
    end
end
@time f()
```
```
elapsed time: 0.284033004 seconds (163 MB allocated, 11.49% gc time in 8 pauses with 0 full sweep)
```

versus
```bash
» time bedtools intersect -loj -sorted -a trya.bed -b tryb.bed > /dev/null
```
```
0.42s user 0.01s system 96% cpu 0.438 total
```

Success! :fireworks:

Now we just have to be competitive with bedtools on fifty other operations...

Blahah:
Nice work! :cactus:

Blahah:
(that's a celebratory cactus, appropriate because they thrive despite the barren conditions)

dcjones:
Of course, I know a celebratory cactus when I see one. :smile:

binarybana:
Holy cow! That's awesome! Great job! It makes me feel even better about the
few hours investment I spent reading Ragel documentation and going over
your Ragel.jl code in hopes/preparation of eventually writing a GTF and/or
VCF parser someday.

dcjones:
I'm definitely going to write documentation explaining how to make ragel parsers. I had to write a few myself and let that code evolve first. I think BigBed is probably the most complex format that ragel will still be useful for (there are other formats, e.g. xml based ones that it probably isn't useful for), so once that's done the ragel.jl stuff should be pretty stable and I can more thoroughly document it.

binarybana:
Thanks, I look forward to the documentation.

Do you have plans to pull the ragel.jl stuff out into it's own package? I
can see this being useful in other domains potentially. (Though between
that and documentation and more parsers I'd much prefer the latter for
priority :). Just a thought.

dcjones:
I agree, it should eventually be split into a separate package, but like you say, I'm prioritizing parsers and docs for now.

PR #51, opened 2015-05-02
BigBed/BigWig reading and writing

dcjones:
The most elaborate parser yet. This one is for BigBed, which is an indexed, compressed version of Bed. It's indexed in such a way that you can find intervals overlapping a region without reading or parsing the whole file.

```julia
bb = read("motifs.bb", BigBed) # this only reads the file's headers

for interval in bb[Interval("chr9", 141110889, 141150059)]
    # `interval` is of type `BEDInterval`
    @show interval
end
```

This still need's tests, so there's more work to do. I'm focusing on BigBed (and BigWig) right now because my plan is make a package for using the [biodalliance](http://www.biodalliance.org/) browser locally (e.g. within IJulia notebooks). This will let you manipulate and visualize genomics data all within Julia.


dcjones:
This is ready to go now. Testing is a little light right now, given the complexity of the format, but it's at least checking that BED to BigBed and BigBed to BED conversions are lossless and that intersection queries are correct.

By the way, after some searching I'm pretty sure this only other code to generate BigBed/Wig besides Jim Kent's, so that's pretty cool!

src/seq/nucleotide.jl
```diff
@@ -1518,6 +1545,7 @@ Count ocurrences of short (<= 32) k-mers in a sequence.
 # Arguments:
   * 'seq`: A NucleotideSequence
   * `step`: K-mers counted are separated by this many nucleotides (deafult: 1)
+<<<<<<< HEAD
```
bicycle1885: A conflict happened here :smile: ?

bicycle1885:
Will `Ragel.Buffer` be superseded by `BufferedReader`?

bicycle1885:
To: @dcjones

dcjones:
I'm actually working on a separate package ([BufferedStreams.jl](https://github.com/dcjones/BufferedStreams.jl)), that I want to use to replace both. I need to do a little more experimentation with that, then I'm going to transfer it to BioJulia and refactor this code.

bicycle1885:
I got it. Thank you :+1:

PR #52, opened 2015-05-09
Accession number types

bicycle1885:
Since accession numbers are routinely used in bioinformatics, it would be worth considering to define accession types in Julia.
This is a quick, suboptimal and incomplete trial to implement accession types for four databases: Entrez gene, GenBank, RefSeq and Gene Ontology.

Advantages (I think) of accession types over simple strings are:
- more explicit meaning
- normalized representation
- smaller footprint in size
- faster operation (e.g. comparison, hashing)

Disadvantages are:
- laborious type definition
- overhead of data conversion

Here I've implemented a few useful tools:
- equality test with string accession numbers (`=={T}(x::Accession{T}, y::String)`)
- open a web browser and access a database entry quickly (`browse(x::Accession)`)

And other tools I imagine:
- quickly `download` an entry with accession
- ID conversions and cross references

I'd like to know your opinions about my proposal.

Thank you.

### DIFF
--- a/src/services/accession.jl
+++ b/src/services/accession.jl
@@ -6,14 +6,15 @@ export Accession, browse, uri
 using URIParser
 
 @doc """
-`Accession{S,T}` is a type holding an accession number of a database named as `S`,
-a symbol like `:RefSeq` or `:GeneOntology`. The accession number is encoded in a
-type `T`, but users don't have to care about `T` in most cases.
+`Accession{S,T}` is a type holding an accession number of a database named as
+`S`, a symbol like `:RefSeq` or `:GeneOntology`. The accession number is encoded
+in a type `T`, but users don't have to care about `T` in most cases.
 
-To make an `Accession` object from a string, you can call `parse(Accession{S}, str)`.
-For example, `parse(Accession{:RefSeq}, "NC_000017.11")` creates an accession object
-of the NCBI RefSeq database. But in this case, `parse(Accession, "NC_000017.11")`
-or even more compactly `Accession("NC_000017.11")` works well because the accession
+To make an `Accession` object from a string, you can call
+`parse(Accession{S}, str)`.  For example,
+`parse(Accession{:RefSeq}, "NC_000017.11")` creates an accession object of the
+NCBI RefSeq database. But in this case, `parse(Accession, "NC_000017.11")` or
+even more compactly `Accession("NC_000017.11")` works well because the accession
 string has a distinctive prefix "NC_" and its accession type can be nicely inferred.
 """ ->
 immutable Accession{S,T}
@@ -58,10 +59,11 @@ const parse_trial_order = [
 ]
 
 @doc """
-Parse an accession string smartly. If the accession string has a distinctive pattern
-like a unique prefix, this function will be able to guess the type of the accession
-number; otherwise it will fail to parse and raise an error. Since the guessing method
-is not so reliable, the result should be checked by a caller.
+Parse an accession string smartly. If the accession string has a distinctive
+pattern like a unique prefix, this function will be able to guess the type of
+the accession number; otherwise it will fail to parse and raise an error.
+Since the guessing method is not so reliable, the result should be checked by
+a caller.
 """ ->
 function parse(::Type{Accession}, s::String)
     for name in parse_trial_order
@@ -147,7 +149,7 @@ function ismatch(::Type{Accession{:EntrezGene}}, s::String)
 end
 
 @doc """
-Parse an accession number string of Entrez Gene (e.g. "7157" for TP53 gene of H.sapiens).
+Parse an accession number string of Entrez Gene (e.g. "7157").
 """ ->
 function parse(::Type{Accession{:EntrezGene}}, s::String)
     @check_match :EntrezGene s
@@ -208,7 +210,7 @@ function ismatch(::Type{Accession{:GenBank}}, s::String)
 end
 
 @doc """
-Parse an accession number of GenBank (e.g. "AB082923.1" for H.sapiens mRNA for P53).
+Parse an accession number of GenBank (e.g. "AB082923.1").
 """ ->
 function parse(::Type{Accession{:GenBank}}, s::String)
     @check_match :GenBank s
@@ -256,7 +258,7 @@ function ismatch(::Type{Accession{:RefSeq}}, s::String)
 end
 
 @doc """
-Parse an accession number string of NCBI RefSeq (e.g. "NC_000017.11" for a genomic assembly).
+Parse an accession number string of NCBI RefSeq (e.g. "NC_000017.11").
 """ ->
 function parse(::Type{Accession{:RefSeq}}, s::String)
     @check_match :RefSeq s
@@ -426,7 +428,10 @@ end
 #   http://www.uniprot.org/help/accession_numbers
 
 function ismatch(::Type{Accession{:UniProt}}, s::String)
-    return ismatch(r"^\s*(:?[OPQ][0-9][A-Z0-9]{3}[0-9]|[A-NR-Z][0-9](:?[A-Z][A-Z0-9]{2}[0-9]){1,2})\s*$", s)
+    return ismatch(
+        r"^\s*(:?[OPQ][0-9][A-Z0-9]{3}[0-9]|[A-NR-Z][0-9](:?[A-Z][A-Z0-9]{2}[0-9]){1,2})\s*$",
+        s
+    )
 end
 
 @doc """
--- a/src/services/accession.jl
+++ b/src/services/accession.jl
@@ -1,4 +1,6 @@
-export Accession, browse
+export Accession, browse, uri
+
+using URIParser
 
 # Accession numbers (S::Symbol database name, T::DataType encoding type)
 immutable Accession{S,T}
@@ -16,6 +18,10 @@ end
 =={S}(x::Accession{S}, y::String) = x == parse(Accession{S}, y)
 =={S}(x::String, y::Accession{S}) = parse(Accession{S}, x) == y
 hash(x::Accession) = hash(x.data)
+
+function browse(accession::Accession)
+    run(`open $(uri(accession, format=:browser))`)
+end
 
 # Smart Accession Parser
 #
@@ -282,6 +288,8 @@ end
 
 # Gene Ontology
 
+# http://geneontology.org/page/ontology-structure#termstru
+
 function ismatch(::Type{Accession{:GeneOntology}}, s::String)
     return ismatch(r"^\s*GO:\d{7}\s*$", s)
 end
@@ -325,4 +333,14 @@ end
 function show(io::IO, uniprot::Accession{:UniProt})
     write(io, uniprot.data)
     return
+end
+
+function uri(uniprot::Accession{:UniProt}; format::Symbol=:browser)
+    if format === :browser
+        ext = ""
+    else
+        @assert format ∈ [:txt, :fasta, :xml, :rdf, :gff]
+        ext = "." * string(format)
+    end
+    return URI("http://www.uniprot.org/uniprot/$uniprot$ext")
 end
--- a/src/services/accession.jl
+++ b/src/services/accession.jl
@@ -1,0 +1,136 @@
+export Accession, EntrezGene, GenBank, RefSeq, GOTerm, browse
+
+
+immutable Accession{T}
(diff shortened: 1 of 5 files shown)


Blahah:
Looks good! I think that when you get down to use-cases like downloading entry data or metadata, it starts to be worth having these types as you could do things like map an array of mixed identifiers to their sequences (this is just a specific benefit of the 'normalised representation' pro you listed).

Another nice thing would be to have a constructor for Accession that parses a string into the correct type, if it can be determined. This would be possible for RefSeq and GO, but probably not for Entrez, in which case it could raise an error about ambiguity of the identifier type.

dcjones:
Great stuff! I agree with @Blahah that it would be nice if we could say `Accession("NC_000001.11")`, and have it be smart enough to know that that's a genbank accession, since many types of accession numbers are globally unambiguous.

Eventually it would be great if we could read entries directly over http, like `read(Accession("NC_000001.11"), FASTA)`. Another possibility to to cache data associated accession numbers, and that way have a sort of package manager for data. There wouldn't be dependencies so we wouldn't have to deal with the hard part of making a package manager, and it would make it very convenient to keep track sequences and gene annotations and such.

That's thinking ahead though. I'm happy to merge this, but have a couple implementation comments:

Is there a reason `Accession` is a parametric type with one field as opposed to just an abstract type?

It's possible to open a URL in a relatively cross-platform way. I used a function like this in Gadfly, which could be adapted.
```julia
function open_file(filename)
    if OS_NAME == :Darwin
        run(`open $(filename)`)
    elseif OS_NAME == :Linux || OS_NAME == :FreeBSD
        run(`xdg-open $(filename)`)
    elseif OS_NAME == :Windows
        run(`$(ENV["COMSPEC"]) /c start $(filename)`)
    else
        warn("Showing plots is not supported on OS $(string(OS_NAME))")
    end
end
```

bicycle1885:
Thank you for your comments.

Guessing accession types is also in my plan. I'll do that :wink: .

I'm thinking about `open`ing entries associated accession numbers over HTTP or local cache like:
```julia
open(Accession("NC_000001.11")) do input
    for seq in read(input, FASTA)
        # ...
    end
end
```
But anyway it would be a next iteration.

> Is there a reason Accession is a parametric type with one field as opposed to just an abstract type?

Just because I thought `Accession{EntrezGene}` looked more explicit than `EntrezGene`; `EntrezGene` may be misunderstood as an entry itself.
But actually I wanted to hide database names to avoid namespace contamination.
Symbol parameters might be helpful but with no luck:
```julia
immutable Accession{Symbol,T}
    data::T  # encoded accession number
end

# something I imagine but won't work
Accession{:EntrezGene}(x::String) = Accession{:EntrezGene,Uint32}(parse(Uint32, x))

x = Accession{:EntrezGene}("1234")
...
```

Some other ideas to avoid namespace would be:

introducing `Accessions` module
```julia
module Accessions
abstract Accession
immutable EntrezGene <: Accession
...
```

prepending a prefix
```julia
abstract Accession
immutable AccEntrezGene <: Accession
....
```

I would like to hear your ideas if you have.

---

@dcjones Your `open_file` looks good! Thanks!

bicycle1885:
This will be much prettier + fallback parser bonus.

```julia
using Compat

immutable Accession{S,T}
    data::T  # encoded accession number
end

if VERSION >= v"v0.4-"
    Base.call{S}(::Type{Accession{S}}, s::String) = parse(Accession{S}, s)
end

# fallback parser for undefiend accession types (just strip white spaces)
function Base.parse{S}(::Type{Accession{S}}, s::String)
    return Accession{S,String}(strip(s))
end

# and specific databases...
function Base.parse(::Type{Accession{:EntrezGene}}, s::String)
    return Accession{:EntrezGene,Uint32}(parse(Uint32, s))
end
```

```
julia> parse(Accession{:EntrezGene}, "1234")
Accession{:EntrezGene,Uint32}(0x000004d2)

julia> parse(Accession{:AkashicRecord}, "AR:42")
Accession{:AkashicRecord,String}("AR:42")

# v0.4 only
julia> Accession{:EntrezGene}("1234")
Accession{:EntrezGene,UInt32}(0x000004d2)
```

src/services/accession.jl
```diff
@@ -0,0 +1,217 @@
...
+    elseif ismatch(Accession{:Ensembl}, s)
+        return parse(Accession{:Ensembl}, s)
+    elseif ismatch(Accession{:GeneOntology}, s)
+        return parse(Accession{:GeneOntology}, s)
+    end
+    error("cannot guess accession number type")
+end
+
+# parser of decimal integer; this is very common among accession numbers
+function parse_decimal_int(s::String, start::Int=1, stop::Int=endof(s))
```
dcjones: Could `parse(Int, s)` be used instead of this?

src/services/accession.jl
```diff
@@ -0,0 +1,217 @@
...
+            break
+        else
+            error("invalid character: '$c'")
+        end
+        i += 1
+    end
+    return n
+end
+
+function search_nonspace(s::String)
```
dcjones: This could also be done with `findfirst(c -> !isspace(c), s)`.

src/services/accession.jl
```diff
@@ -0,0 +1,217 @@
...
+function ismatch(::Type{Accession{:CCDS}}, s::String)
+    return ismatch(r"^\s*CCDS\d+\.\d+\s*$", s)
+end
+
+
+# Ensembl
+
+# http://www.ensembl.org/info/genome/stable_ids/index.html
+
+function ismatch(::Type{Accession{:Ensembl}}, s::String)
```
dcjones: Since Ensembl is versioned, and the gene, protein, transcript, etc, referred to may change between versions, it seems important to associate the Ensembl version with the identifier. I'm not sure how best to handle that.

src/services/accession.jl
bicycle1885: I think `parse(Int, s)` has two problems:
1. `parse(Int, "0x1234")` is valid
2. there is no way to start parsing from a specified position

src/services/accession.jl
bicycle1885: Calling anonymous functions will be slower.
But you think this is a premature optimization, I'll follow your advice.

src/services/accession.jl
bicycle1885: I've forgot that. The version number should be added as a field. Thanks!

src/services/accession.jl
Blahah: Can you include an explanation of *why* this shouldn't be used outside an interactive session?

src/services/accession.jl
Blahah: I think that with Accessions, almost any optimisation will be premature. My vote is for readability over speed for this module.

src/services/accession.jl
bicycle1885: Of course. I wrote the reason.
I'd like to write `@doc`s after coding finished.

src/services/accession.jl
bicycle1885: Okay, I'll put a high priority on readability and will rewrite it.

bicycle1885:
I've found that we have to make a decision about lack of accession versions.
[Annovar](http://annovar.openbioinformatics.org/) and some other common tools don't append versions at the end of accession numbers.

If we admit that, we should care about equality of accession numbers.
Assuming that "NM_022162" == "NM_022162" and "NM_022162.1" == "NM_022162.1", there will be three reasonable options:

1. Match up to numbers: "NM_022162" == "NM_022162.1", "NM_022162.1" == "NM_022162.2"
2. Match up to versions if completely versioned: "NM_022162" == "NM_022162.1", "NM_022162.1" != "NM_022162.2"
3. Match up to versions: "NM_022162" != "NM_022162.1", "NM_022162.1" != "NM_022162.2"

IMHO, option 3 would be less confusing and hence preferable.
I'd like to hear your thoughts.

bicycle1885:
Still working.

This is my TODO memo:
* equality of accession numbers
* hash functions
* more tests
* more accession types (if needed)

dcjones:
This is looking good. A few comments:

I think using `Nullable` for the version fields will make it more clear when there is no version provided. It's possible that some database may actually have a version 0, so indicating an "unversioned" accession with `version == 0` could cause problems. I agree that considering versions in equality in equality comparisons is the least confusing option.

I'm still not totally convinced that `parse_decimal_uint` is needed, because because in julia 0.4 (and in the Compat module) you can say `parse(Int, s, 10)` to restrict the string to decimal digits.

Please add some docstrings, especially to exported names, even if they are just one-liners like
```julia
@doc "Accession number for a sequence stored in GenBank" ->
immutable GenBank
...
```

I really like the idea of defining `uri` methods for accession numbers. That's nice!

bicycle1885:
@dcjones Thank you for your comment. I was busy this weekdays, but I'm willing to fix up my code and get ready it to be merged this weekend.

bicycle1885:
@dcjones 

Almost ready to be merged.
I might have to parse (almost) all accession numbers of all databases to validate the correctness.
If you think it is needed, I'll try. But it would be unreasonable to include all accession numbers in the test directory.

By the way, this morning I knew Julia Summer of Code 2015 (JSoC) (http://julialang.org/blog/2015/05/jsoc-cfp/) and I'm very interested in it though I don't know the detail.
If someone could be my mentor, I'd like to apply to JSoC under the BioJulia organization.
Since the deadline of the application is very soon (June 1st), if you are also interested in JSoC, we will have to decide a project as soon as possible.

My short background:
I am a graduate student at University of Tokyo, Japan, studying human genetics of rare diseases and being engaged in processing and analysing high-throughput sequencing data.
I am an organizer of JuliaTokyo group (http://julia.tokyo/).
I have 1.5 years' experience of Julia programming and wrote several packages:
* https://github.com/docopt/DocOpt.jl
* https://github.com/bicycle1885/RandomForests.jl
* https://github.com/bicycle1885/ANMS.jl
* https://github.com/bicycle1885/Snappy.jl

Some projects in my mind:
* BAM/SAM/CRAM/VCF/BCF format parser
* variant annotation toolkit
* extending DataFrame.jl for bioinformatics

Thank you.

dcjones:
I think it would be terrific if you applied for JSOC. I would be free to act as mentor, but if anyone else is keen, feel free to speak up. The project ideas you listed are good ones, but we'll need to flesh out something concrete for a proposal. To avoid hijacking this issue, will you open a new issue to discuss project ideas? That might be the best way to solicit feedback.

bicycle1885:
Thank you! I opened an issue at https://github.com/BioJulia/Bio.jl/issues/54.

src/services/accession.jl
```diff
@@ -0,0 +1,461 @@
...
+
+
+# UniProt
+# -------
+
+# webpage:
+#   http://www.uniprot.org/
+# accession format:
+#   http://www.uniprot.org/help/accession_numbers
+
```
diegozea: Is this going to deal also with [Entry names](http://www.uniprot.org/help/entry_name)? There are very useful for working with Pfam and other databases.

bicycle1885: At the moment, No. I'd like to have a one-to-one mapping between identifiers and entries that is stable and primary accession number.
But the entry name of UniProt would be useful to access the database, so it should be supported as a database-specific package/module in the future.
Thank you for pointing :+1:

diegozea:
What is the state of this PR?

Ward9250:
AFAIK @bicycle1885 has been working on data-structures an algorithms for alignment algorithms and such too as part of a funded project so he's had a lot on his plate. @bicycle1885 can you maybe give an update here - almost tick box style of what is currently done with this, and what is still required? Others may be able to help on parts then too if they have time.

bicycle1885:
@diegozea Almost ready to be merged. I think I have to check that it can parse all accession numbers before merging.

I don't know a good way to download all accession numbers from a database. Does anyone know?
Or, testing for all accession numbers may not be necessary.

kescobo:
Bump.

Any word on this PR in the last year? Would be useful to use in my genbank parser

bicycle1885:
I'd say this might not be so useful as I thought. Accession numbers are fairly complicated and they are so different from each other. So, maybe a better way to handle accessions is regarding them as usual strings.

Ward9250:
Maybe we can stick a few more tags on here: Comments Appreciated, Help Wanted and see if people have ideas for how this would be done, and "pass the idea ball" around a bit?

PR #56, opened 2015-06-19
WIP: Optimization/Bugfix Omnibus

dcjones:
I'm rushing to put together my talk for JuliaCon next week. In the process I'm optimizing and fixing things I encounter. Though ideally I would make a separate PRs for each of these changes, I'm in a hurry so I'm just going to push changes as I make them to this PR and merge sometime before my talk on Thursday.

None of these should be breaking changes, only optimizing, fixing, and adding minor features.

Changes so far:
  * Fix a out of bounds exception in parsing that I somehow never encountered before.
  * Significant optimizations to nucleotide and k-mer iterators.
  * Add an option to `translate` to turn ambiguous codons into `X` rather than throwing at error.
  * Add `KmerCounts` that works analogously to `NucleotideCounts`.


Ward9250:
Ah yes I recall that using @inline was recommended for iteration methods by Tim Holy on a PR on the Julia repo once - making sure to avoid the cost of the function calls.

dcjones:
It was a big boon to kmer and nucleotide iterator speed, maybe not surprisingly since those can involve calling the same relatively simple function hundreds of millions of times. We're faster than SeqAn and Biostrings with a lot of basic sequence operations now.

Ward9250:
When you're done with the JuliaCon talk will you upload it? I'd love to see it!

dcjones:
I'll put it up in github repo this weekend in an incomplete state so people can comment. I'd be happy to get feedback on it, or suggestions for things to add.

bicycle1885:
Great work! Though I cannot fully check the code, it looks well-tested. I think it's okay to merge it.

dcjones:
This fails tests, but only because I need to tag a new version of IntervalTrees. I'm going to go ahead and merge.

PR #57, opened 2015-06-24
Writing of FASTQ sequences

kdmurray91:
NB: this is work in progress

Hi all,

As discussed on gitter, I've implemented a function to write a FASTQ record to an IO stream. I imagine that it could be implemented better, including using some functionality from `quality.jl` to do the quality string conversion.

Also, I'd assume a FASTA writer would come in handy.

Let me know what you think! I'm still not very experienced w/ Julia, so any and all criticism is welcome :)

Cheers,
Kevin

### DIFF
--- a/src/seq/fastq.jl
+++ b/src/seq/fastq.jl
@@ -27,9 +27,9 @@ end
 "A `SeqRecord` for FASTQ sequences"
 typealias FASTQSeqRecord DNASeqRecord{FASTQMetadata}
 
-@doc """
+"""
 Show a `FASTQSeqRecord` to `io`, with graphical display of quality scores.
-""" ->
+"""
 function Base.show(io::IO, seqrec::FASTQSeqRecord)
     write(io, "@", seqrec.name, " ", seqrec.metadata.description, "\n")
     for c in seqrec.seq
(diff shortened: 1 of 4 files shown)


kdmurray91:
(pasting in from gitter, as I'm not sure @dcjones saw the comment)

> @dcjones I'm not sure, but I think I may have found a bug in `quality.jl`, with respect to Illumina 1.5-1.8 quality scores. By my reading of [wikipedia](https://en.wikipedia.org/wiki/FASTQ_format#Encoding), the range of the Illumina 1.5 pipline values is 3-40. Which with the offset of 64, is the ASCII range `CDE...fgh` (typically, full range goes to 62 and  `~`). So, should [the Illumina 1.5 quality config](https://github.com/BioJulia/Bio.jl/blob/2ae3aab1f383a36579edc12820e9fe0984783264/src/seq/quality.jl#L62) be `['C', '~',  3]`? Or perhaps `['B', '~', 2]`?

dcjones:
Sorry to let this collect dust. I think it looks good. There are a few minor syntax updates needed that I can do (`@doc` is no longer needed for docstrings, and factcheck uses `-->` instead of `=>` now).

I think you're right that Illumina 1.5 is not correct, but it seems like it should be `['B', '~', 2]`, since `B` can occur even though it has some special meaning. Sound right?

kdmurray91:
@dcjones that sounds right, yeah. I've rebased onto current master and will fix the factcheck issues.

kdmurray91:
And the Illumina 1.5 issue too.

kdmurray91:
@dcjones I think all your comments have all been addressed.

dcjones:
Thanks!

PR #63, opened 2015-06-30
Uppercase module files

prcastro:
Making module files start with uppercase and update the styleguide with this change. Solves #46.


prcastro:
Superseded by #97

PR #65, opened 2015-07-10
Blastwrapper

kescobo:
## Blast Tools module for Bio.jl - Spec and Roadmap

Bio.jl needs a way to run BLAST and other command line tools commonly used in biology applications, taking inputs and capturing outputs in Bio.jl formats. 

### Minimum requirements for BLAST module
With Blast+ installed on users' system:

**Input**

- [x] accept seq object~~, string~~ or file as query and subject
    - [x] accept single sequence objects
    - [x] accept multi-sequence objects
- [x] blastn, blastp, blastall
    - [ ] others?
- [x] accept flags for search modification
- [x] take/parse blast xml output

**Output**

- [x] return alignment object or strings for alignment
- [x] return stats object w/other information
- [ ] allow return of other blast outputs (tsv, csv etc)

### Other Useful Features

- [ ] bundle blast package from within Bio.jl (is this possible?)
- [ ] pull queries/subjects from NCBI
    - [ ] incorporate into Bio.jl formats
- [ ] run Blast through NCBI servers instead of locally
- [ ] ... others?

### DIFF
--- a/src/tools/blast/README.md
+++ b/src/tools/blast/README.md
@@ -1,0 +1,30 @@
+##Blast Tools module for Bio.jl - Spec and Roadmap
+
+Bio.jl needs a way to run BLAST and other command line tools commonly used in biology applications, taking inputs and capturing outputs in Bio.jl formats.
+
+###Minimum requirements for BLAST module
+With Blast+ installed on users' system:
+
+**Input**
+
+- [x] accept seq object or file name as query and subject
+    - [x] accept single sequence objects
+    - [x] accept multi-sequence objects
+- [x] blastn, blastp, ~~blastall~~
+    - [ ] others?
+- [x] accept flags for search modification
+- [x] take/parse blast xml output
+
+**Output**
+
+- [x] return alignment object or strings for alignment
+- [x] return stats object w/other information
+- [ ] allow return of other blast outputs (tsv, csv etc)
+
+###Other Useful Features
+
+- [ ] bundle blast package from within Bio.jl (is this possible?)
+- [ ] pull queries/subjects from NCBI
+    - [ ] incorporate into Bio.jl formats
+- [ ] run Blast through NCBI servers instead of locally
+- [ ] ...
--- a/src/tools/blast/blastcommandline.jl
+++ b/src/tools/blast/blastcommandline.jl
@@ -1,0 +1,170 @@
+# wrapper for BLAST+ command line functions
+
+immutable BlastResult
+    bitscore::Float64
+    expect::Float64
+    queryname::ASCIIString
+    hitname::ASCIIString
+    hit::BioSequence
+    alignment::AlignedSequence
+end
+
+"""
+`readblastXML(blastrun::ASCIIString)`
+Parse XML output of a blast run. Input is an XML string eg:
+```julia
+results = readall(open("blast_results.xml")) # need to use `readstring` instead of `readall` for v0.5
+readblastXML(results)
+```
+
+Returns Vector{BlastResult} with the sequence of the hit, the Alignment with query sequence, bitscore and expect value
+"""
+function readblastXML(blastrun::ASCIIString; seqtype="nucl")
+    xdoc = parse_string(blastrun)
+    xroot = root(xdoc)
+    params = get_elements_by_tagname(xroot, "BlastOutput_param")
+    iterations = get_elements_by_tagname(xroot, "BlastOutput_iterations")
+    results = BlastResult[]
+    for iteration in collect(child_elements(iterations[1]))
+        queryname = content(find_element(iteration, "Iteration_query-def"))
+        hits = get_elements_by_tagname(iteration, "Iteration_hits")
+        for hit in collect(child_elements(hits[1]))
+            hitname = content(find_element(hit, "Hit_def"))
+            hsps = get_elements_by_tagname(hit, "Hit_hsps")
+            for hsp in collect(child_elements(hsps[1]))
+                if seqtype == "nucl"
+                    qseq = BioSequence{DNAAlphabet{4}}(content(find_element(hsp, "Hsp_qseq")))
+                    hseq = BioSequence{DNAAlphabet{4}}(content(find_element(hsp, "Hsp_hseq")))
+                elseif seqtype == "prot"
+                    qseq = AminoAcidSequence(content(find_element(hsp, "Hsp_qseq")))
+                    hseq = AminoAcidSequence(content(find_element(hsp, "Hsp_hseq")))
+                else
+                    throw(error("Please use \"nucl\" or \"prot\" for seqtype"))
+                end
+
+                aln = AlignedSequence(qseq, hseq)
+                bitscore = float(content(find_element(hsp, "Hsp_bit-score")))
+                expect = float(content(find_element(hsp, "Hsp_evalue")))
+                push!(results, BlastResult(bitscore, expect, queryname, hitname, hseq, aln))
+            end
+        end
+    end
+    return results
+end
+
+"""
+`readblastXML(blastrun::Cmd)`
+Parse command line blast query with XML output. Input is the blast command line command.
+
+Returns Vector{BlastResult} with the sequence of the hit, the Alignment with query sequence, bitscore and expect value
+"""
+function readblastXML(blastrun::Cmd; seqtype="nucl")
+    # need to use `readstring` instead of `readall` for v0.5
+    readblastXML(readall(blastrun), seqtype=seqtype)
+end
+
+
+"""
+`blastn(query, subject, flags...)``
+Runs blastn on `query` against `subject`.
+    Subjects and queries may be file names (as strings), BioSequence{DNAAlphabet{4}} type or
+    Array of BioSequence{DNAAlphabet{4}}.
+    May include optional `flag`s such as `["-perc_identity", 95,]`. Do not use `-outfmt`.
+"""
+function blastn(query::AbstractString, subject::AbstractString, flags=[]; db::Bool=false)
+    if db
+        results = readblastXML(`blastn -query $query -db $subject $flags -outfmt 5`)
+    else
+        results = readblastXML(`blastn -query $query -subject $subject $flags -outfmt 5`)
+    end
+    return results
+end
+
+function blastn(query::BioSequence{DNAAlphabet{4}}, subject::BioSequence{DNAAlphabet{4}}, flags=[])
+    querypath, subjectpath = makefasta(query), makefasta(subject)
+    blastn(querypath, subjectpath, flags)
+end
+
+function blastn{S <: BioSequence{DNAAlphabet{4}}}(query::BioSequence{DNAAlphabet{4}}, subject::Vector{S}, flags=[])
+    querypath, subjectpath = makefasta(query), makefasta(subject)
+    blastn(querypath, subjectpath, flags)
+end
+
+function blastn(query::BioSequence{DNAAlphabet{4}}, subject::AbstractString, flags=[]; db::Bool=false)
+    querypath = makefasta(query)
+    if db
+        blastn(querypath, subject, flags, db=true)
+    else
+        blastn(querypath, subject, flags)
+    end
+end
+
+function blastn{S <: BioSequence{DNAAlphabet{4}}}(query::Vector{S}, subject::Vector{S}, flags=[])
+    querypath, subjectpath = makefasta(query), makefasta(subject)
+    blastn(querypath, subjectpath, flags)
+end
+
+"""
+`blastp(query, subject, flags...)``
+Runs blastn on `query` against `subject`.
+    Subjects and queries may be file names (as strings), `BioSequence{AminoAcidSequence}` type or
+    Array of `BioSequence{AminoAcidSequence}`.
+    May include optional `flag`s such as `["-perc_identity", 95,]`. Do not use `-outfmt`.
+"""
+function blastp(query::AbstractString, subject::AbstractString, flags=[]; db::Bool=false)
+    if db
+        results = readblastXML(`blastp -query $query -db $subject $flags -outfmt 5`, seqtype = "prot")
+    else
+        results = readblastXML(`blastp -query $query -subject $subject $flags -outfmt 5`, seqtype = "prot")
+    end
+end
+
+function blastp(query::AminoAcidSequence, subject::AminoAcidSequence, flags=[])
+    querypath, subjectpath = makefasta(query), makefasta(subject)
+    blastp(querypath, subjectpath, flags)
+end
+
+function blastp{S <: AminoAcidSequence}(query::AminoAcidSequence, subject::Vector{S}, flags=[])
+    querypath, subjectpath = makefasta(query), makefasta(subject)
+    blastp(querypath, subjectpath, flags)
+end
+
+function blastp(query::AminoAcidSequence, subject::AbstractString, flags=[]; db::Bool=false)
+    querypath = makefasta(query)
+    if db
+        blastp(querypath, subject, flags, db=true)
+    else
+        blastp(querypath, subject, flags)
+    end
+end
+
+function blastp{S <: AminoAcidSequence}(query::Vector{S}, subject::Vector{S}, flags=[])
+    querypath, subjectpath = makefasta(query), makefasta(subject)
+    blastp(querypath, subjectpath, flags)
+end
+
+"""
+`makefasta(sequence::BioSequence)`
+Create temporary fasta-formated file for blasting.
+"""
+function makefasta{T <: BioSequence}(sequence::T)
+    path, io = mktemp()
+    write(io, ">$path\n$(convert(AbstractString, sequence))\n")
+    close(io)
+    return path
+end
+
+"""
+`makefasta(sequence::Vector{BioSequence})`
+Create temporary multi fasta-formated file for blasting.
+"""
+function makefasta{T <: BioSequence}(sequences::Vector{T})
+    path, io = mktemp()
+    counter = 1
+    for sequence in sequences
+        write(io, ">$path$counter\n$(convert(AbstractString, sequence))\n")
+        counter += 1
+    end
+    close(io)
+    return path
+end
--- a/src/tools/blast/blastcommandline.jl
+++ b/src/tools/blast/blastcommandline.jl
@@ -33,8 +33,8 @@ function readblastXML(blastrun::ASCIIString; seqtype="nucl")
             hsps = get_elements_by_tagname(hit, "Hit_hsps")
             for hsp in collect(child_elements(hsps[1]))
                 if seqtype == "nucl"
-                    qseq = BioSequence{DNAAlphabet{4}}(content(find_element(hsp, "Hsp_qseq")))
-                    hseq = BioSequence{DNAAlphabet{4}}(content(find_element(hsp, "Hsp_hseq")))
+                    qseq = DNASequence(content(find_element(hsp, "Hsp_qseq")))
+                    hseq = DNASequence(content(find_element(hsp, "Hsp_hseq")))
                 elseif seqtype == "prot"
                     qseq = AminoAcidSequence(content(find_element(hsp, "Hsp_qseq")))
                     hseq = AminoAcidSequence(content(find_element(hsp, "Hsp_hseq")))
@@ -54,21 +54,25 @@ end
 
 """
 `readblastXML(blastrun::Cmd)`
-Parse command line blast query with XML output. Input is the blast command line command.
+Parse command line blast query with XML output. Input is the blast command line command, eg:
+```julia
+blastresults = `blastn -query seq1.fasta -db some_database -outfmt 5`
+readblastXML(blastresults)
+```
 
 Returns Vector{BlastResult} with the sequence of the hit, the Alignment with query sequence, bitscore and expect value
 """
 function readblastXML(blastrun::Cmd; seqtype="nucl")
     # need to use `readstring` instead of `readall` for v0.5
-    readblastXML(readall(blastrun), seqtype=seqtype)
+    return readblastXML(readall(blastrun), seqtype=seqtype)
 end
 
 
 """
 `blastn(query, subject, flags...)``
 Runs blastn on `query` against `subject`.
-    Subjects and queries may be file names (as strings), BioSequence{DNAAlphabet{4}} type or
-    Array of BioSequence{DNAAlphabet{4}}.
+    Subjects and queries may be file names (as strings), DNASequence type or
+    Array of DNASequence.
     May include optional `flag`s such as `["-perc_identity", 95,]`. Do not use `-outfmt`.
 """
 function blastn(query::AbstractString, subject::AbstractString, flags=[]; db::Bool=false)
@@ -80,28 +84,28 @@ function blastn(query::AbstractString, subject::AbstractString, flags=[]; db::Bo
     return results
 end
 
-function blastn(query::BioSequence{DNAAlphabet{4}}, subject::BioSequence{DNAAlphabet{4}}, flags=[])
+function blastn(query::DNASequence, subject::DNASequence, flags=[])
+    querypath, subjectpath = makefasta(query), makefasta(subject)
+    return blastn(querypath, subjectpath, flags)
+end
+
+function blastn(query::DNASequence, subject::Vector{DNASequence}, flags=[])
     querypath, subjectpath = makefasta(query), makefasta(subject)
     blastn(querypath, subjectpath, flags)
 end
 
-function blastn{S <: BioSequence{DNAAlphabet{4}}}(query::BioSequence{DNAAlphabet{4}}, subject::Vector{S}, flags=[])
-    querypath, subjectpath = makefasta(query), makefasta(subject)
-    blastn(querypath, subjectpath, flags)
-end
-
-function blastn(query::BioSequence{DNAAlphabet{4}}, subject::AbstractString, flags=[]; db::Bool=false)
+function blastn(query::DNASequence, subject::AbstractString, flags=[]; db::Bool=false)
     querypath = makefasta(query)
     if db
-        blastn(querypath, subject, flags, db=true)
+        return blastn(querypath, subject, flags, db=true)
     else
-        blastn(querypath, subject, flags)
+        return blastn(querypath, subject, flags)
     end
 end
 
-function blastn{S <: BioSequence{DNAAlphabet{4}}}(query::Vector{S}, subject::Vector{S}, flags=[])
+function blastn(query::Vector{DNASequence}, subject::Vector{DNASequence}, flags=[])
     querypath, subjectpath = makefasta(query), makefasta(subject)
-    blastn(querypath, subjectpath, flags)
+    return blastn(querypath, subjectpath, flags)
 end
 
 """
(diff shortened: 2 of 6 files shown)


kescobo:
Some questions I have about style/implementation:

1) Types - does it make sense to make an additional type to hold the results of a BLAST search, or is there a better way to do this? My plan was to use the XML output of BLAST and parse it using [LightXML](https://github.com/JuliaLang/LightXML.jl), so the results will already be in a structured form, but I don't have a good grasp on memory usage or performance (or even a good sense of how I would structure a Type).

2) File structure - are there guidelines for when to split into a new `.jl` file? I've got the module file on its own, but I could imagine putting everything I'm doing into `blastcommandline.jl`, or having separate files for the functions and for the xml parsing portion etc. 

3) I'm sure there will be other questions...

Blahah:
@kescobo sorry that your questions have gone unanswered for so long.

1. Yes, IMO a type for the blast results makes sense. Look at the structure of the incoming data to guide your type definition. And ask here or on Gitter for help.
2. We don't have a rule on this, except that the file declaring the module should not do anything else (see [the style guide](https://github.com/BioJulia/Bio.jl/blob/master/CONTRIBUTING.md)). Beyond that I would say that you should keep everything in one file until it makes sense to split, e.g. if the file is too long to navigate easily, or if functionality from one file is used in others in the same module.

src/tools/blast/blastcommandline.jl
```diff
@@ -0,0 +1,24 @@
+# wrapper for BLAST+ command line functions
+
+using LightXML
```
prcastro: This could go into blast.jl

src/tools/blast/blastcommandline.jl
```diff
@@ -0,0 +1,24 @@
...
+Runs blastn on `query` against `subject`. May include optional `flags` such as
+    `[-pident, 95]`. Do not use `-outfmt`.
+"""
+function blastn(query::String, subject::String, flags...)
+    result = readall(`blastn -query $query -subject $subject -outfmt 5 $flags`)
+    parse_string(result)
+end
+
+#testing
+println(blastn("/Users/KBLaptop/computation/tmp/testseq.fasta", "/Users/KBLaptop/computation/tmp/testseq2.fasta"))
```
prcastro: Would be nice to move this to `/test` before merging. Consider using `FactCheck` when making this kind of test.

src/tools/blast/blastcommandline.jl
```diff
@@ -0,0 +1,24 @@
...
+    queryname::String
+    subjectname::String
+    query::NucleotideSequence
+    subject::NucleotideSequence
+end
+    
+
+
+"""
+Runs blastn on `query` against `subject`. May include optional `flags` such as
```
prcastro: Could you put the functions signature right before this line? You can find examples of such signatures on `/src/seq/nucleotide.jl`

src/tools/blast/blastcommandline.jl
```diff
@@ -0,0 +1,24 @@
...
+    `[-pident, 95]`. Do not use `-outfmt`.
```
prcastro: We could have a link (or a pointer of any kind) to where these flags are documented.

src/tools/blast/README.md
```diff
@@ -0,0 +1,30 @@
+##Blast Tools module for Bio.jl - Spec and Roadmap
+
+Bio.jl needs a way to run BLAST and other command line tools commonly used in biology applications, taking imputs and capturing outputs in Bio.jl formats.
```
prcastro: imput -> input

kescobo:
No worries - I haven't had a chance to work on this much anyway. Turns out teaching 3 courses in a semester while trying to write a paper and apply for jobs takes up a lot of time. Thanks for the comments, I'll hopefully return to this in the next couple of months, but it someone else wants to work on it, I won't be offended.

src/tools/blast/blastcommandline.jl
kescobo: Good to know, though upon further reflection I may try to grab the command-line output rather than write and parse the xml.

src/tools/blast/blastcommandline.jl
kescobo: Agree - I'm going to need a lot of help devising proper tests I think. This is just part of my own development process. Do you know of a good resource that provides some basic principles for developing tests?

src/tools/blast/blastcommandline.jl
kescobo: Do you mean something like:

    # Basic Functions
    # ---------------

?

prcastro: No, I mean put the function signature inside the docstring:

```julia
"""
`blastn(query::String, subject::String, flags...)`

Runs blastn on `query` against `subject`. May include optional `flags` such as
`[-pident, 95]`. Do not use `-outfmt`.
"""
```

kescobo: Ahh, gotcha.

src/tools/blast/README.md
kescobo: Fixed

kescobo:
[Link to comment](https://gitter.im/BioJulia/Bio.jl?at=56782cc4f240f5a004178ad2) from @Ward9250 about parsing blast results. This implementation might be a bit beyond me - any help/pointers would be appreciated.

Ward9250:
@kescobo A Ragel implementation is not essential for this as said by @Blahah, as long as the implementation works, a Ragel parser could come along later. I wouldn't ever force that someone has to use the Ragel framework we've got, I just suggest it would be neat, seeing's as it's there and does have some advantages, like a standard form of specifying parsers and so on.

kescobo:
@Ward9250 I agree it makes sense to use it - and my goal for doing this is to learn. Part of me just wants to get anything working, but I think a bigger part wants to get it done in the best way possible. Since this module isn't critical, I might as well take my time and do it well, right?

Ward9250:
I would say the right way is your way. Some people, like to see stuff working rapidly, and then optimize, redesign, and refine later, some people prefer to take their time and do something once. As for myself, it depends entirely on the project, what I'm doing, and whether or not I know enough about the problem I have. My HybridCheck project was something done in something of a rush according to other projects I got dragged into. Now I have all the time in the world I'm doing a V2 I'm taking my time with, in some cases redesigning whole swathes of it.

src/tools/blast/blastcommandline.jl
```diff
@@ -0,0 +1,49 @@
...
+    May include optional `flags` such as `[-pident, 95]`. Do not use `-outfmt`.
+"""
+function blastn(query::ASCIIString, subject::ASCIIString, flags...)
+    result = readall(`blastn -query $query -subject $subject -outfmt 6 $flags`)
+    # do stuff
+end
+
+function blastn(query::DNASequence, subject::DNASequence, flags...)
+    querypath, subjectpath = makefasta(query), makefasta(subject)
+    result = readall(`blastn -query $querypath -subject $subjectpath -outfmt 6 $flags`)
```
kescobo: Would it make sense here simply call `blastn(querypath, subjectpath)`, which would use the version of the function that operates on `ASCIIString`s? Can't tell if it's super unclear to call a function of the same name (but that acts on different types) within a function.

Ward9250: I think it's ok. I've done similar things before in method dispatch. So long as people read the docs this is pretty non-confusing. You define a method blast, which works on DNASequence, it achieves it function, by writing the DNASequence objects to temp files, and it passes the paths, to the blasts method that is designed to work on filepaths. Indeed defining a method of blasts for use with file paths, and then not using it in other methods seems like a case of the method to me.

bicycle1885:
Let me put two cents about this wrapper.
I think the best output format is XML because we already have a parser of the XML format: https://github.com/JuliaLang/LightXML.jl.
Moreover, I'm already working on supporting the XPath interface for the package: https://github.com/JuliaLang/LightXML.jl/pull/38. So, my implementation idea is to write a simple BLAST wrapper that calls the BLAST+ CLI tools from Julia which returns an XML document object (or a thin wrapper object of it) as a result of BLAST search. The necessary information of the BLAST search result can be extracted from the XML document object on demand using XPath queries. Of course wrappers of XPath queries that returns a Julia object are very useful.

Ward9250:
I concur about use of LightXML - I used it before to implement a PhyloXML parser - it is nice!

kescobo:
@bicycle1885 Thanks for the suggestion - that was my original idea - also the author of biopython has mentioned that the XML output tends to change less frequently (though they updated it [earlier this year](ftp://ftp.ncbi.nlm.nih.gov/blast/documents/NEWXML/xml2.pdf)). I suppose then it would just be a matter of documenting how the xml is parsed? 

I can't remember [why i decided](https://github.com/BioJulia/Bio.jl/pull/65#discussion-diff-40507095) to move away from that idea, possibly because the outputs to xml are fixed, whereas there are a lot of additional options with the tabular output, but this seems minor. I think I agree with you.

PS. Also, in googling for that XML2 documentation, I saw that they just last week [released v2.3.0](http://www.ncbi.nlm.nih.gov/books/NBK131777/). How much should I worry about supporting older versions? I'm thinking it makes the most sense to use the most recent XML output, and not bother with legacy support.

kescobo:
OK, I've got the initial call of `blastn` working, and am going to start working on parsing the XML output. while I do that, I could use some feedback on the sort of information that the result should contain, and the form it should take. The blast result contains a lot of information - most of which I've never used, but I suspect others might.

The good news is that a lot of it will be contained in the alignment (things like alignment length, mismatches etc), so if I can figure out how to construct an alignment object with the result, that will cover a lot of bases (no pun intended). I think I can build a type for a single `BlastResult` type that contains the bit score, e-value and the alignment.

The things I'd like comments on: 

First, for the `BlastResult` type, if the initial inputs are `Sequence` objects already, that's great - the alignments can reference those. But if they're fasta files (or particularly if the subject is a blast database), it might get tricky. The blast result does not necessarily contain the entire sequence of the query or the subject. So I can build a new Seq object out of what's there, but let's say `hit1` is `subject_A` bases 20-120, and I make that sequence, then hit 2 is `subject_A` bases 15-120... should I re-make `subject_A`, make `subject_A` mutable and add the bases on the beginning and end? 

Second, making an iterable that contains all of the `BlastResult`s for a given search. eg. should these be separated by query? Should they somehow link back to the initial inputs (this seems related to the above question)?. What other information should be included? The BioPython parser is basically just a wrapper around the xml so that all of the information is obtained, but this has always felt clunky to me. And if I'm parsing it enough to make the alignment objects mentioned above, is seems like leaving the rest of it in its present form is a waste.

Third, I haven't used Blast for proteins much, and I suspect there's different information contained in the results. Should `BlastResult` be an abstract type, and then have `NucleicAcidBlastResult` and `ProteinBlastResult` be subtypes?

Ward9250:
@kescobo Regards the idea of supporting older versions of BLAST. I'd say because it's recently changed as you say, support for the very latest, and then one version back is probably the most sensible.

kescobo:
Can I get some eyes on current implementation? I think I'm getting closer to something that can be merged; I have `blastn` and `blastp` working for `BioSequence` types and strings that represent fasta formatted files. Functions currently return a vector of a new `BlastResult` type that contains bit scores, e-values and an alignment. 

I also pulled out the function that reads the xml output, so someone can potentially do custom command line commands and read the results in the same way (eg. `readblastXML(`blastp -query test_seqs1.faa -subject test_seqs2.faa -outfmt 5`, seqtype="prot") `), or by running blast in the command line and then passing the file to the function in julia eg:

```julia
results = readall(open("blast_results.xml"))
readblastXML(results)
```

I have a couple of questions/requests for comment...

First, does what I have so far make sense? I'm not entirely clear on types in julia, I'm especially unsure about the `BlastResult` type, is it useful? Also for other stuff, I got it to work but I'm not sure I'm using best practices (I'm thinking specifically about the `flags` argument and the `db` argument)

Second, I'm unsure how to deal with sequence names from fasta files and blast databases. I think ideally they'd be stored in the `BlastResults` type, but Blast+ is notorious for screwing with sequence names and making them hard to find unless they adhere to very specific conventions ([see here](http://blastedbio.blogspot.com/2013/12/blast-should-keep-its-blordid.html) [and here](http://blastedbio.blogspot.com/2012/10/my-ids-not-good-enough-for-ncbi-blast.html) for example). Being able to find the name of the sequence would be super useful, but I'm not sure I want to try to figure out and support all of the possible ways this can get f'd up. 

Third, I think I'm going to need some guidance on writing tests. I've got a couple of test sequences and files I've been using to test stuff, but I've never actually written any tests for anything.

Finally, I have a crazy mess of commits here because I still don't fully grok git. I read that it's not a good idea to squash old commits if you've already pushed, but I have no idea how to deal with the messy history.

Ward9250:
Ok just had a read, I think you might need to rebase on BioJulia:master as I think there's sequence stuff there that's out of date, NucleotideSequence and AminoAcidSequence types have been modified and replaced by the BioSequence framework, which allows for sequences of any kind of biological alphabet you can define. 

The Blast Result type you introduce looks ok to me, if you think it's practical, making it immutable would be a benefit - both for aggregate storage and arrays/vectors of the type, and to allow the Julia compiler to make optimisations. It might be worthwhile exporting the type from the module Blast too. Regarding the names, if you want, you can just support a few clear cases and not worry about the fucked up cases for now just to get what works merged.

I think it's possible for us to help you out with tests. It's possible to add other people's forks as remotes. Once you're done with working on this issue. We can let it hang temporarily, and then open issues on your fork to help you out with tests and maybe rebasing.

Re the commits, I think rebasing and squashing would be fine. You've been working on your own fork of Bio.jl, and haven't been messing with any files also being worked on by anyone else, and have only been pushing this to your fork.

ScottPJones:
Which files are actually part of your change?  Just the ones in the `src/tools/blast` directory + the test file in `test/tools/blast`?  If so, it might be easier to copy those 5 files to another directory, delete your current branch (kescobo:blastwrapper, i.e. `git branch -D blastwrapper` on your local machine), recreate the branch after switching to master and making sure it is up-to-date (i.e. `git checkout -b blastwrapper`, from your master, *after* having done a `git pull master`), copy the files into `src/tools/blast` and `test/tools/blast`, and finally do a `git commit -a -m "BLAST support"`, and then force push it back to GitHub (`git push origin +blastwrapper`).  (I hope this helps, probably you knew all that anyway, in which case, sorry for the noise!)

Ward9250:
Actually yes very good point @ScottPJones, @kescobo  that might be easier than trying to rebase, whilst you are still getting used to Git.

kescobo:
There are a couple of other changes (like adding the module to the main `tools.jl`), but these are quite minor. I had hoped to avoid the [xkcd approach](https://xkcd.com/1597/), but that may be the best option.

Following @ScottPJones ' protocol, would I think do a new pull request from the new branch, or will it apply here since it's the same name? 

I might try to make a copy of the folder and attempt the rebase to learn though.

Ward9250:
@kescobo I believe if it's the same name and you do a forced push, that you would not need a new PR.

kescobo:
@Ward9250 but would that screw with this comment history? Since some of the comments are linked to lines of code it seems like a bad idea... Perhaps making a new PR linking to this one would be cleaner? That way I don't muck up the history but also don't pollute master with my git wanderings. Or am I over-thinking it?

ScottPJones:
From my experience, it would only effects comments that are made on a specific commit (and even then, it shows that there were out-of-date comments, and you can click and expand to see them)

Ward9250:
Exactly, I think it's the same as it is with rebases, as @ScottPJones said, it only affects commits with specific line comments. It still keeps them, but it tells you the specific comment is out of date.

Ward9250:
The complex commit history dies with the old branch you're going to replace, and when you do a force push, github will just accept it as the new history, no resolving history or anything like that as you're forcing a complete overwrite.

kescobo:
OK - rebase was going to be nightmare. Early on I made commits to master before making a new branch, and decided it wasn't worth dealing with. 

Ready to start working on tests now... I have a couple of files/sequences that I've been using to test via a script, what would be the easiest way to upload those? I've got a protein fasta, a nucleic acid fasta, and "prot" and "nucl" blastdb's. I'm happy to write the tests, just need some guidance.

Ward9250:
So if these are good and representative files of their kind, then it will probably be worth adding them to the BioJulia/BioFmtSpecimens repo. Which is a repo we are building of a set of data and format specimens. This repo can be used to get data from during tests then.

kescobo:
Cool - I didn't know that existed. Just [made a PR](https://github.com/BioJulia/BioFmtSpecimens/pull/9)

Ward9250:
Ok I merged it into the specimens repo.

kescobo:
ugh... I think I mucked up the git history again. But I've got some tests started, though there are some problems. When I try to test, I get an error 

```julia
ERROR: LoadError: LoadError: UndefVarError: Bio not defined
 in include at /Applications/Julia-0.4.5.app/Contents/Resources/julia/lib/julia/sys.dylib
 in include_from_node1 at /Applications/Julia-0.4.5.app/Contents/Resources/julia/lib/julia/sys.dylib
 [inlined code] from /Users/KBLaptop/.julia/v0.4/Bio/test/runtests.jl:27
 in anonymous at no file:0
 in include at /Applications/Julia-0.4.5.app/Contents/Resources/julia/lib/julia/sys.dylib
 in include_from_node1 at /Applications/Julia-0.4.5.app/Contents/Resources/julia/lib/julia/sys.dylib
 in process_options at /Applications/Julia-0.4.5.app/Contents/Resources/julia/lib/julia/sys.dylib
 in _start at /Applications/Julia-0.4.5.app/Contents/Resources/julia/lib/julia/sys.dylib
while loading /Users/KBLaptop/.julia/v0.4/Bio/test/tools/runtests.jl, in expression starting on line 11
while loading /Users/KBLaptop/.julia/v0.4/Bio/test/runtests.jl, in expression starting on line 26
```

Also, the tests aren't looking for proper outputs at this point, they're just checking to make sure that the functions return something. Guidance would be appreciated.

Ward9250:
I don't know how the findnext stuff got in there. Do you have BioJulia/Bio.jl set up as a remote for kescobo/Bio.jl?

kescobo:
Yes... I've been pulling BioJulia/Bio.jl into kescobo/Bio.jl master. and then rebasing my blastwrapper branch off kescobo/Bio.jl master.

What should I be doing? I'll go back and sanitize this again, but would be good to avoid doing this in the future.

Ward9250:
When I do it, I fetch BioJulia master. Rebase or merge my master using BioJulia master, and then rebase my branching I'm working on, on my master. Which sounds pretty much the same as what you're doing.

ScottPJones:
I have origin and upstream set up, and always keep to the following sequence:
`git checkout master ; git fetch upstream ; git pull upstream master ; git push origin master`
to update my local copy and my fork on GitHub.
You really have to be careful to not do the `git pull` while in your working branch (I used to frequently make that mistake).
When I'm in my working branch, I modify the files, then do `git commit -a -m "some message"`, and `git push origin spj/mybranch` (or `git push origin +spj/mybranch` if I've rebased).  Since I started being careful like that, I haven't had any more of those sorts of problems.

ScottPJones:
Oh, and another thing people showed me how to do: just before I do something potentially damaging like a `git rebase -i master`, I always (while in my working branch) do `git checkout -b backup/<mybranch>-<nn> ; git checkout spj/<mybranch>` to make a good backup.  If things go badly, I can simply do `git branch -D spj/<mybranch> ; git branch -m backup/<mybranch>-<nn> spj/<mybranch>`. To help keep my fingers from slipping, I've made some little scripts that do those git commands for me, given the name of the branch, and (for backing up or restoring) my backup version number.
Hope that's helpful!

kescobo:
@Ward9250 @ScottPJones OK, thanks. I'm not sure what I did differently... maybe I did a pull from BioJulia/Bio.jl master into my blastwrapper branch by accident? I'll be a bit more diligent in the future. 

In any case, this should be good for people to take a look at the tests. I'm still getting the `Bio not defined` error when trying to run the tests on my system. But on TravisCI, it seems to get passed that point, but fails because the blast+ command line tools aren't installed. Any ideas on how to get around that? 

It also occurs to me that it would be good to check if blast is installed and throw a helpful error message if it isn't, but I don't know how to do that.

Ward9250:
Maybe Blast will need to be installed as a Binary Dependency for Julia... this might complicate things. Alternatively, we will just have to add an installation step to Travis for blast before the julia tests are run.

ScottPJones:
@kescobo Yes, that what it looks like to me (I wrote the little scripts doing the above because before that, I shot myself in the foot with distressing regularity, typing pull instead of push, for example, so don't feel bad!)

kescobo:
OK, I got my tests to work, though there's a test in a different package that's currently making tests fail. But if I move the tests for "tools" up to the front, they pass. I'm just checking that it returns the correct type, `Array{BlastResult, 1}` - should I check the content as well or is this sufficient?

If this is good, I think this is ready to merge...?

src/tools/blast/blastcommandline.jl
```diff
@@ -0,0 +1,170 @@
...
+function blastp{S <: AminoAcidSequence}(query::Vector{S}, subject::Vector{S}, flags=[])
+    querypath, subjectpath = makefasta(query), makefasta(subject)
+    blastp(querypath, subjectpath, flags)
+end
+
+"""
+`makefasta(sequence::BioSequence)`
+Create temporary fasta-formated file for blasting.
+"""
+function makefasta{T <: BioSequence}(sequence::T)
```
bicycle1885: There is no need to write a docstring that is not exported to users. Also it would be better to write the signature as `makefasta(sequence::BioSequence)`.

src/tools/blast/blastcommandline.jl
```diff
@@ -0,0 +1,170 @@
...
+    for iteration in collect(child_elements(iterations[1]))
+        queryname = content(find_element(iteration, "Iteration_query-def"))
+        hits = get_elements_by_tagname(iteration, "Iteration_hits")
+        for hit in collect(child_elements(hits[1]))
+            hitname = content(find_element(hit, "Hit_def"))
+            hsps = get_elements_by_tagname(hit, "Hit_hsps")
+            for hsp in collect(child_elements(hsps[1]))
+                if seqtype == "nucl"
+                    qseq = BioSequence{DNAAlphabet{4}}(content(find_element(hsp, "Hsp_qseq")))
+                    hseq = BioSequence{DNAAlphabet{4}}(content(find_element(hsp, "Hsp_hseq")))
```
bicycle1885: `DNASequence` is an alias for `BioSequence{DNAAlphabet{4}}`, so you can use that name here.

src/tools/blast/blastcommandline.jl
```diff
@@ -0,0 +1,170 @@
...
+
+"""
+`readblastXML(blastrun::Cmd)`
+Parse command line blast query with XML output. Input is the blast command line command.
+
+Returns Vector{BlastResult} with the sequence of the hit, the Alignment with query sequence, bitscore and expect value
+"""
+function readblastXML(blastrun::Cmd; seqtype="nucl")
+    # need to use `readstring` instead of `readall` for v0.5
+    readblastXML(readall(blastrun), seqtype=seqtype)
```
bicycle1885: Explicit `return` is our (my?) favorite style if you intended to return a value.

Ward9250: Agreed, it's clearer if you explicitly return the value. I'll check that this is in the code guidelines, and add it if it is not.

src/tools/blast/blastcommandline.jl
kescobo: @Ward9250 had suggested that I change to the new syntax. Do we know that the alias is going to persist? Is it better to use the alias do you think?

src/tools/blast/blastcommandline.jl
kescobo: OK. I'm still slightly confused by this syntax - at one point I had `function blastn(query::Vector{DNASequence}...` and that didn't work. Is there a simple explanation for when you need to do the `{t <: Whatever}` thing before the arguments?

El-Zurdo-Sordo: That shouldn't really be necessary, because you are not doing anything with the type variable `T`.
`function makefasta(sequence::BioSequence)` should be fine in that case.

src/tools/blast/blastcommandline.jl
```diff
@@ -0,0 +1,170 @@
...
+    write(io, ">$path\n$(convert(AbstractString, sequence))\n")
+    close(io)
+    return path
+end
+
+"""
+`makefasta(sequence::Vector{BioSequence})`
+Create temporary multi fasta-formated file for blasting.
+"""
+function makefasta{T <: BioSequence}(sequences::Vector{T})
```
El-Zurdo-Sordo: Here, you do need the T<: BioSequence, because you want a vector of some type that is BioSequence or a super type of BioSequence.

src/tools/blast/blastcommandline.jl
bicycle1885: I'm going to use `DNASequence` as an alias for the most general DNA sequence type so that it covers 99% of use cases. So I believe `DNASequence` is suitable for the purpose of handling "unseen" sequences returned from BLAST. Do you concur with me, @Ward9250 ?

src/tools/blast/blastcommandline.jl
kescobo: Do you know if there is a section in the docs that describes this difference? I need to do some reading. 

In any case, is there any reason not to use the same letter for all of them (eg `{T <: BioSequence}` and `{T <: AminoAcidSequence}` and {T <: DNASequence}`? In other words, should I do `{B <: BioSequence}` and `{A <: AminoAcidSequence}` and {D <: DNASequence}` or something?

ScottPJones: It has to do with the fact that `Vector{Int8}` for example is not a subtype of `Vector{Integer}`.
It seems the unwritten convention for type variables is to use `T` if it's for a type, and is the only one (I've seen `T` and `S`, when there are two type variables).

src/tools/blast/blastcommandline.jl
bicycle1885: The concept of a parametric type is a little bit complicated. But `function blastn(query::Vector{DNASequence}...)` will work; try this snippet, for example:
```julia
using Bio.Seq

function sumlen(seqs::Vector{DNASequence})
    len = 0
    for s in seqs
        len += length(s)
    end
    return len
end

@show sumlen([dna"ACGT", dna"AAA"])
```

In general, a method definition with parametric types whose type parameters contain an abstract type does not make sense in Julia. This is called *invariance* of parametric types (http://julia.readthedocs.org/en/latest/manual/types/#parametric-composite-types). In the above example, the method definition is valid because the type parameter of `Vector`, `DNASequence`, is a concrete type.

src/tools/blast/blastcommandline.jl
ScottPJones: From the types section of the manual, if you have `var::type`, if type is concrete, then `var` must be of that type, but it `type` is abstract, `var` can be any subtype of `type`.

kescobo: Ahh, ok - that makes sense I think. So if my function only should accept a vector of `Int8`, I can just do `function foo(bar::Vector{Int8})`, but if I want it to be able to use any integer, I have to do `function foo{T <: Int}(bar::Vector{T})`?

To the later point, you mean use `T` and `S` if there are two type variables *in the same function*, but just use `T` if I have multiple functions that all need a single type variable?

ScottPJones: Almost!  Remember, `Int` is just an alias for the bitstype `Int32` or `Int64` (depending on platform begin 32-bit or 64-bit), you need to say `T <: Integer` if you want to accept any subtype of `Integer`.

Yes, that seems to be the convention, or `T1` and `T2`, for example, when there are more than on in the same function type variables.

src/tools/blast/blastcommandline.jl
kescobo: I've definitely read that part of the docs... clearly don't understand it completely. I could have sworn I tried something like `function blastn(query::Vector{DNASequence}...)` and something went wrong. Though now that I'm thinking about it, it might have been early on when I was trying to use `Vector{AbstractString}`. So if I'm understanding correctly, when a `Vector{ASCIIString}` got passed, that would have triggered an error, yes?

src/tools/blast/blastcommandline.jl
kescobo: @ScottPJones I blame python brain - but yes, that makes sense.

kescobo:
@bicycle1885 Thanks for all the comments - I think I addressed them all.

Also, got my first real glimpse of how awesome unit tests are - saved me from pushing a stupid mistake :+1:

Ward9250:
Unit tests really do you a service, especially when you define them by how you *think* your code should work, before you've run it and know what pops out the end. Makes sure you implemented what you think you've implemented.

kescobo:
Sorry to pester, but does anyone else want to add comments to this? Trying to get things off my mental ledger...

Ward9250:
So is there anything else to add to this PR? If not then I'll take the WIP label off, and in a day or two if there are no complaints I'll merge it.

blahah:
Let's ask the channel if anyone has time for a review pass - I'm swamped at the moment.

Ward9250:
LGTM :shipit:

bicycle1885:
Merged. Thank you, @kescobo !

PR #68, opened 2015-07-30
import operators from Base explicitly

bicycle1885:
`import Base.Operators` has been removed in Julia: https://github.com/JuliaLang/julia/pull/12235.
So we have to import all operators explicitly from the `Base` module from now on.
This pull request adds missing operators to the import list.

### DIFF
--- a/src/seq/seq.jl
+++ b/src/seq/seq.jl
@@ -3,9 +3,11 @@ module Seq
 using Compat
 using Base.Intrinsics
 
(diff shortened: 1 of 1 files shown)


Ward9250:
Nice catch. Very simple change - I'm satisfied. Can someone give a second yay?

prcastro:
Seems OK to me also. Merging.

PR #69, opened 2015-07-30
New FactCheck syntax (=> / -->)

prcastro:
Update to the new syntax required by FactCheck


prcastro:
This is very straightforward, so I'm merging the change

PR #74, opened 2015-08-07
Write FASTA to file

wildgarden:
This PR allows to write FASTA sequences to files. The sequence is written in chunks that obey the maximum character per line restriction. I'm not quite happy how I iterate over the sequence to split it into chunks. Since I'm new to Julia I didn't find a more elegant approach that uses string manipulation/list-comprehension.

### DIFF
--- a/test/seq/test_seq.jl
+++ b/test/seq/test_seq.jl
@@ -1292,4 +1292,27 @@ facts("Sequence Parsing") do
     end
 end
 
+facts("Sequence Writing") do
+    context("FASTA writing") do
+        dna_seq = random_dna(79)
+        seq_name = "Some sequence name"
+
+        dna_with_line_breaks = string(dna_seq, "\n", dna_seq, "\n", dna_seq, "\n")
+        expected = string(">", seq_name, "\n", dna_with_line_breaks)
+
+        long_dna_seq = dna_seq ^ 3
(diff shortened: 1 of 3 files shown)


dcjones:
LGTM :+1:

test/seq/test_seq.jl
```diff
@@ -1292,4 +1292,37 @@ facts("Sequence Parsing") do
...
+        sequences = [fasta_seq1, fasta_seq2]
+
+        tempfile = tempname()
+        try
+            open(tempfile, "w") do fh
+                for seq in sequences
+                    write(fh, seq)
+                end
+            end
+            @fact readall(open(tempfile)) --> expected
```
bicycle1885: `open(readall, tempfile)` can automatically close the temporary file after reading its whole content.

wildgarden:
@bicycle1885 Thank you, I somehow overlooked this option and adopted the PR in  0f1170623e30ef457f2536867bd9edb524c99e9d 

To conform with the logic used in fastaq.jl the conversion of nt was changed in 4e31f99017c54364c4b82966ae8c533ac0c9eb22 to delegate it to function show().

Ward9250:
Indeed, looks good.

prcastro:
@wildgarden , this looks good. Could we rebase it on the latest `master` so it gets merged?

dcjones:
I've rebased and merged this to save @wildgarden the headache.

Ward9250:
@dcjones Thanks :+1:

wildgarden:
@dcjones Thank you!

PR #76, opened 2015-08-10
Added waffle.io badges to the Bio.jl README.md

Ward9250:


Ward9250:
Any objections to having waffle badges? Just let's people know at a glance how many issues there are, and how many are being worked on by someone.

prcastro:
Nice!

PR #80, opened 2015-08-18
Tokenizer tool

Ward9250:
This was originally part of the Phylo module, but I’ve decided it should be rehired in the tools.jl file or Bio.jl and treated as a general utility available to Bio.jl users/modules as I’ve found several uses for it, first for tokenising Newick, and now doing the same for CIGAR strings.


Ward9250:
This is basically ready to merge now. It's a very basic tokeniser, which we might expand functionality of later. Phyllo uses it and it's probable the alignment module may use it.

bicycle1885:
Sorry, I'm not sure the use case of this tokenizer.
This functionality looks a dup of ragel.

Ward9250:
@bicycle1885 It is a tokenizer that was inside the Phylo module for the newick string parser - which I wrote without Ragel as IIRC @dcjones was not sure Ragel was suitable for Newick as it was not a regular language. But I would like the Phylo files to be parsed with Ragel if possible. I have asked recently on PR #17 whether we are at a point now whether Ragel could be used for newick strings given the grammar here: https://code.google.com/p/phylopy/wiki/Newick and here: https://en.wikipedia.org/wiki/Newick_format

The second reason it exists currently is I was trying to parse newick trees and such for Phylo for a long time before Ragel and our parser framework was as solid as it is now. I think some of this even goes back to Phylogenetics.jl - which I made redundant and merged into Bio.Phylo. It's also used for parsing CIGAR strings in Bio.Align on my feature branch, which has a macro/string-literal like `cigar"3M3N3M"` which uses the tokenizer to build a cigar string variable. So far I've had two use cases for it.

prcastro:
Should this live inside a module of its own? Why not on the root of Bio? Also, if Ragel can be used for something not sequence-related, it should be moved to `/src/tools`.

dcjones:
You can tokenize newick with ragel. You just can easily parse it completely, since matching parenthesis is context free and not regular. 

Here's a basic newick tokenizer:
```julia

import Bio
import Bio.Ragel
using BufferedStreams, Switch


immutable Newick <: Bio.FileFormat end


type NewickTree
    # ...
end


%%{
    machine newick;

    action anchor { Ragel.@anchor! }
    action token {
        Ragel.@copy_from_anchor!(token)
        println(token)
    }

    length = '-'? digit+ ('.' digit*)?;
    unquoted_label = [A-Za-z_\-] [A-Za-z0-9_\-]*;
    quoted_label = '"' [^"]* '"';
    token = ('(' | ')' | ',' | ':' | ';' | length | unquoted_label | quoted_label) >anchor %token;

    main := space* (token space*)**;
}%%


%% write data;


type NewickParser <: Ragel.AbstractParser
    state::Ragel.State

    function NewickParser(input::BufferedInputStream)
        %% write init;

        return new(Ragel.State(cs, input))
    end
end


function Base.open(stream::BufferedInputStream, ::Type{Newick})
    return NewickParser(stream)
end


Ragel.@generate_read_fuction("newick", NewickParser, NewickTree,
    begin
        token = Bio.StringField()

        %% write exec;
    end)


example = """
    (
    (raccoon:19.19959,bear:6.80041):0.84600,
    (
        (sea_lion:11.99700, seal:12.00300):7.52973,
        (
            (monkey:100.85930,cat:47.14069):20.59201, 
            weasel:18.87953
        ):2.09460
    ):3.87382,
    dog:25.46154
);
"""

parser = open(example.data, Newick)
tree = NewickTree()
read!(parser, tree)
```

Which prints:
```
(
(
raccoon
:
19.19959
,
bear
:
6.80041
)
:
0.84600
,
(
(
sea_lion
:
11.99700
,
seal
:
12.00300
)
:
7.52973
,
(
(
monkey
:
100.85930
,
cat
:
47.14069
)
:
20.59201
,
weasel
:
18.87953
)
:
2.09460
)
:
3.87382
,
dog
:
25.46154
)
;
```

For a complete parser, you just need to do something with the tokens beyond printing them.

Ward9250:
@dcjones This is perhaps one of the times where I look a bit thick - and a lot of biologists will look thick because any self taught biologist I can think of on campus would hear context free and not-regular and say or think several things, as I just did - all of them dumb "But regex works on it - as in my tokenizer", "But there are grammars describing it that kindof look like those in the ragel machines", "But why couldn't ragel hit a '(' and then execute arbitrary code handling the creation, of a node, and then hit tokens and execute code for filling, and then finalising of nodes - which is pretty much what my parsers do anyway". And this probably needs explaining in the learning stuff we are going to plan for BioJulia. From a quick google search I think I kindof now see why newick can't be completely parsed:

It has to do with the fact that with matching parentheses you have to keep track of whether you have seen an open one already, or how many open ones already. This means it can't strictly be determined if a 'word' is in the language with a fixed size memory by examining all symbols in the word one after the other. The second thing stumbling block is "context-free", which I've figured as far as it has something called non-terminating states in the grammar, which I think means it's because newick is recursive - it's clades in clades in clades, and so the definition of a clade includes clades.

dcjones:
Sorry I was being pretty opaque. You absolutely can create nodes on `(` tokens, then moves back up the tree on `)` tokens and so on. I think that's a good way to go about this. I only meant that ragel can't match parenthesis, so it's up to you to make sure you see the same number of `(` and `)` tokens, which is easy to do.

"Context-free" and "regular" are two too classes of [formal languages](https://en.wikipedia.org/wiki/Formal_language). This was just a jargony way of saying that ragel can't parse everything, and checking that parenthesis match is one of the things it can't automatically do. Which is not to say that ragel isn't useful for these problems, just that they require some extra manual checking.

Ward9250:
So with Ragel then I may be able to set states for ( and ) (moving them out of the token definition) and set actions for them? But in the actions I have to track myself the parens that have been seen and the action would have to account for this. If this is doable then the sounds awesome - really helps me integrate the newick code Phylo's already got with the parsing framework, and avoids duplication - like this Tokenizer.

bicycle1885:
I think you can use a stack to track the current opening parentheses: when you encounter a `(`, push a new node into the stack, and when you encounter a `)`, pop the premature node from the stack, read the label and distance of it, and connect it to the parent node (at the top of the stack, if any).
I'm not familiar with the exact format of the format, but parsing would be doable like this.

Ward9250:
Ok, thanks guys, I've decided I'll try and implement a Newick parser with Ragel, and then get some more tests in, and then RFR Phylo #17 for merge. Furthur additions, improvements and PhyloXML can then come in later PRs. One final question before I read through code for Bufferd Streams, Ragel's docs and the docs @dcjones has written and take a crack at making this Tokenizer obsolete for Phylo: In Phylo currently, there is a sting literal for newick - so I can do this `newick"((A,B)C);"` and it would parse the newick string and create the tree object, can this be done with Ragel? So far we've only talked about it in the contect of reading files, but I wondered (maybe I'm being naievely) - hey if you can make a buffer/stream of a string in memory, then perhaps that's perfectly doable.

Ward9250:
This is now moot. Closing.

Issue #82, opened 2015-08-24
Subsequence of subsequence


bicycle1885:
The following behavior is counter-intuitive to me and I believe this is a bug, isn't it?

```julia
julia> read = dna"AAAATTTT"
8nt DNA Sequence
 AAAATTTT

julia> seed = read[5:end]
4nt DNA Sequence
 TTTT

julia> @assert seed[1:2] == dna"TT"
ERROR: AssertionError: seed[1:2] == @dna_str("TT")

julia> seed[1:2]
2nt DNA Sequence
 AA

```

Ward9250:
I can confirm this behaviour, and indeed seems like a bug when a range is used to getindex() a subsequence, using a range type - single Ints seem fine:

```julia
julia> using Bio.Seq

julia> read = dna"AAAATTTT"
8nt DNA Sequence
 AAAATTTT

julia> seed = read[5:end]
4nt DNA Sequence
 TTTT

julia> seed
4nt DNA Sequence
 TTTT

julia> seed[1:2]
2nt DNA Sequence
 AA

julia> seed[1]
T

julia> seed[2]
T

julia> seed[3]
T

julia> seed[4]
T

julia> seed[5]
ERROR: BoundsError
 in getindex at /Users/axolotlfan9250/.julia/v0.4/Bio/src/seq/nucleotide.jl:552

julia> seed[1:1]
1nt DNA Sequence
 A

julia> seed[1:2]
2nt DNA Sequence
 AA

julia> seed[1:3]
3nt DNA Sequence
 AAA

julia> seed[1:4]
4nt DNA Sequence
 AAAA

julia> seed[1:5]
ERROR: Invalid subsequence range
 in call at /Users/axolotlfan9250/.julia/v0.4/Bio/src/seq/nucleotide.jl:207
 in getindex at /Users/axolotlfan9250/.julia/v0.4/Bio/src/seq/nucleotide.jl:563
```

bicycle1885:
@Ward9250 Thank you. I've already patched in my local work space, so I'll fix it soon.

Ward9250:
Is it clear from the code what the issue is? I haven't used or looked at the subsequence code in Seq yet.

bicycle1885:
Yes, it is. I hope this will fix the problem https://github.com/BioJulia/Bio.jl/commit/cb595c6c8baf4b7acd1ee4024ff3fc0f281bc432.

PR #83, opened 2015-08-24
Fix subseq

bicycle1885:
This will fix the problem of #82


Ward9250:
Tests work! Looks good!

PR #85, opened 2015-09-05
Parsing Infrastructure Iteration

dcjones:
This is a major iteration on the parsing infrastructure (as you can see by the giant change set). It does the following:

  * Remove buffering code and use [BufferedStreams.jl](https://github.com/BioJulia/BufferedStreams.jl) instead.
  * Makes sequence and interval types fully mutable.
  * Introduce a simple mutable string implementation: `StringField`
  * Parsers now revolve around a in-place `read!` function. Iteration (non-destructive parsing) comes for free. `open` is used to initialize a parser rather than `read`.
  * Use [Libz.jl](https://github.com/BioJulia/Libz.jl) instead of Zlib.jl for BigBed

Net effect:
  * Defining parsers becomes simpler in several ways, e.g. you don't have to write your own iterator and intermediate values don't have stored in the parser, since that can be written one-by-one to the output type.
  * In-place parsing allows for extremely efficient non-allocating parsing. This is basically a C-style interface, and it should let us be competitive with C/C++ libraries.

    ```julia
    stream = open("reads.fastq", FASTQ) # open a stream
    x = FASTQSeqRecord() # allocate an empty fastq record
    while read!(stream, x)
        # do something with x
    end

    # iteration also still works
    for x in stream
        # do something with x
    end
    ```

~~This is based on mutable sequences (#37), so that needs to be merged first. Also, BufferedStreams and Libz aren't yet registered, so this will fail on travis until I do so.~~


bicycle1885:
Great stuff! Thank you for implementing it!

bicycle1885:
A bikeshedding of naming. I think the abstract sequence type, `Sequence`, looks too general and can easily crash with a type name exported from other module. `BioSequence` may be more explicit and safe.

src/seq/seqrecord.jl
```diff
@@ -31,6 +36,11 @@ function Base.getindex(seqrec::SeqRecord, r::UnitRange)
 end
 
 
+function Base.copy{T <: SeqRecord}(seqrec::T)
```
prcastro: I believe `Base.` isn't needed here , since `copy` was imported explicitly.

src/intervals/bed.jl
```diff
@@ -1,42 +1,76 @@
...
+          (n < 5 || a.thick_last == b.thick_last) &&
+          (n < 6 || a.item_rgb == b.item_rgb) &&
+          (n < 7 || a.block_count == b.block_count)
+    if !ans
+        return false
+    end
+
+    if n >= 8
+        for i in 1:a.block_count
+            if a.block_sizes[i] != b.block_sizes[i]
```
prcastro: Should we have `@inbounds` here, since `a.block_count == b.block_count` ?

src/intervals/bed.jl
```diff
@@ -1,42 +1,76 @@
...
+                return false
+            end
+        end
+    end
+
+    if n >= 9
+        for i in 1:a.block_count
+            if a.block_sizes[i] != b.block_sizes[i]
```
prcastro: And also here?

src/seq/seqrecord.jl
dcjones: I think I ended up doing a lot of this, which is inconsistent with our style. I was getting pretty annoyed with importing things. It's starting to feel like C/C++ where you have to make sure declarations and definitions in different files match. I want to rethink this style choice, but maybe this PR isn't the right place to do that, so I can change these.

prcastro: I actually think that both things are annoying. Having to type `Module.` on every definition and importing things on other files. Importing the whole base (`importall Base`) is a bad decision?

dcjones:
Addressed things we discussed on gitter, which cleans it up a bit.

src/seq/aminoacid.jl
```diff
@@ -348,6 +353,27 @@ function setindex!(seq::AminoAcidSequence, nt::Char, i::Integer)
 end
 
 
+"""
+Reset the contents of a mutable sequence from a string.
+"""
+function Base.copy!(seq::AminoAcidSequence, strdata::Vector{UInt8},
```
bicycle1885: `Base.` :eyes:

Ward9250: :laughing: That buttery biscuit `Base.`, leaves crumbs all over the code!

dcjones: Fixed.

prcastro:
Tests are passing on 0.4.

:shipit:

Ward9250:
Awesome! :shipit: I can rebase Phylo on this then and continue looking at getting the newick parser up.

Issue #87, opened 2015-09-08
Provide an editorconfig file


Blahah:
[Editorconfig](http://editorconfig.org/) is a file format for defining style elements that text editors should adhere to (e.g. indent style and size, whitespace trimming, etc.), and a collection of plugins for pretty much every editor that work with the file format. If we provide a file, contributors don't have to remember to switch tab numbers etc. in their editor when working on this project.

prcastro:
Closed by #88

PR #88, opened 2015-09-08
Added editorconfig file

Ward9250:
Written an .editorconfig as suggested in #87

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+# EditorConfig file for source code files of the Bio.jl Package.
+
+root = true
+
+[.jl]
+indent_style = space
+indent_size = 4
+tab_width = 4
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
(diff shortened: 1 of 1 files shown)


prcastro:
Bio.Align introduces some C code to the table. It would be nice to see it already in editorconfig.

dcjones:
Could you change the pattern to `*.{rl,jl}` to include our julia-hosted ragel files? I don't think @Ward9250  meant to include those SeqAn C++ files in #36, so adding C/C++ probably isn't necessary.

Ward9250:
Correct, the C/C++ files are not going to be part of Align, I did put them in, but they are there only for reference whilst I work on Align, A). For me, and B). For others so they can see the basis of what I've done, and where I've altered the design based on our needs.

prcastro:
Just remove WIP tag when you're ready to merge this PR

Ward9250:
Ok cool, WIP removed, two people say yep and it goes in. The WIP tag is used by Waffle, so maybe we should use some other tag to signal a "I'm done with this, please review" tag.

prcastro:
RTR (Ready to Review)?

Ward9250:
Sounds good!

Blahah:
All looks good - merged

PR #91, opened 2015-09-09
Remove Compat dependence

prcastro:
We require Julia v0.4, therefore there's no need to use Compat.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,7 @@ considered it must meet the following guidelines.
     contexts, and into separate files based on module.
 
   * Contributions are included if the code has been reviewed by at least two
-    team members who are **not** the author of the proposed contribution, 
+    team members who are **not** the author of the proposed contribution,
     and there is general consensus (or general lack of objections) that it's useful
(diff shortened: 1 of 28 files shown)


bicycle1885:
Great! This is what I wanted to do for a while.
But one problem would be this will conflict with many other pull requests.
So, I think merging this should be delayed and should be done after merging some large pull requests.

bicycle1885:
Also it would be better to rename `String` to `AbstractString`: the former is no longer used in the standard library.

prcastro:
The only conflicting PR I can see is #37. Other PRs aren't affect AFAIK. @dcjones , do you mind merging this PR and rebasing #37? I usually mess things up when solving merge conflicts on Git.

dcjones:
I think #37 wouldn't be too bad, but #85 worries me a little since it modifies a lot of files.

prcastro:
Oh, you're right. Maybe we will have to wait until both of them are merged.

dcjones:
@prcastro You're free to rebase this when you have a chance.

prcastro:
Superseded by #97

PR #92, opened 2015-09-09
Generating a manual

dcjones:
jI've been working towards rewriting [Judo](https://github.com/dcjones/Judo.jl) to use docstrings and use Base.Markdown instead of pandoc. I've mostly completed that, and we can start using it to generate an html manual.

It works by processing markdown files in the `doc` directory that look like
```md
---
part: Intervals
order: 2
title: Collections of intervals
modules: [Bio.Intervals]
...

# Types

{{IntervalCollection}}
```

You'll notice two unusual things:
  * Arbitrary YAML metadata can be included in a header.
  * Words in brackets (e.g. `{{IntervalCollection}}`) get expanded into their corresponding docstring. The `modules` list in the header gives a set of modules to look for docstrings in.

You can also include code blocks that get executed and the results inlined, which is useful to include examples that have to stay up to date for the documentation to build.

I only started making a skeleton, which you can see here: http://homes.cs.washington.edu/~dcjones/biojl/
(That's not a permanent home, just a convenient place for me to upload).

This is intended as a prototype. I don't want to force Judo on everyone. Try it out, and if there's a better option, let me know.

There is some work ahead to make this nice:
  * Add more docstring references to the markdown files.
  * Write expository documentation in the markdown files to give context to the docstrings.
  * Add docstrings to more things.
  * Tweak how we are writing docstrings: for it to look right we need to use level 3 headers (`###`) for sections like "Arguments" and "Returns" in docstrings. We should also not put function signatures in docstrings, since those should be able to be automatically printed.
  * Add executable examples to docstrings.

This branch is in the main repository so feel free to push documentation improvements to it. This doesn't have to be perfect or totally complete before merging, but we should really have something browsable before publishing Bio.jl.


prcastro:
I really like using Judo to generate documentation.However, before merging, I believe most people from the organization should agree to it.

Ward9250:
I agree to Judo.

dcjones:
I have a first draft manual now, which you can look at here: http://homes.cs.washington.edu/~dcjones/biojl/

It's not amazing, but it's a start, and basically enough for us to go ahead and make a release. I'll put this in `gh-pages` when merged, since I can't make one PR against two branches. So it will live at https://biojulia.github.io/Bio.jl for the time being, and we can figure domain names and hosting later.

Ward9250:
:+1: Looks Great!

prcastro:
It's very nice! Merging, so we can create the Bio.jl manual page on the `gh-pages` branch.

PR #93, opened 2015-09-09
add `convert` to a byte sequence from bio sequences

bicycle1885:
I defined the `convert` methods to byte sequences.
This would be useful in many ways: passing biological sequences to C libraries, allviating overhead of accessing packed nucleotides, and so on.

I think `DNASequence(seq::Union(Vector{Uint8}, String)) = NucleotideSequence(DNANucleotide, seq)` et cetera should be removed/replaced by corresponding `convert` methods since type conversion is much more versatile in Julia.
If you agree with me, I'll fix them and add more commits to this pull request.

### DIFF
--- a/src/seq/nucleotide.jl
+++ b/src/seq/nucleotide.jl
@@ -309,7 +309,7 @@ end
 
 Construct a nucleotide sequence from the `seq[startpos:stoppos]` string
 """
-function NucleotideSequence{T<:Nucleotide}(::Type{T}, seq::Union(AbstractString, Vector{UInt8}),
+function NucleotideSequence{T<:Nucleotide}(::Type{T}, seq::Union{AbstractString, Vector{UInt8}},
                                            startpos::Int, stoppos::Int,
                                            unsafe::Bool=false; mutable::Bool=false)
     len = stoppos - startpos + 1
@@ -325,8 +325,7 @@ function NucleotideSequence{T<:Nucleotide}(::Type{T}, seq::Union(AbstractString,
     return NucleotideSequence{T}(data, ns, 1:len, mutable, false)
 end
 
-function NucleotideSequence{T<:Nucleotide}(t::Type{T}, seq::Union(AbstractString,
-                                           Vector{UInt8}); mutable::Bool=false)
+function NucleotideSequence{T<:Nucleotide}(t::Type{T}, seq::Union{AbstractString, Vector{UInt8}}; mutable::Bool=false)
     return NucleotideSequence(t, seq, 1, length(seq), mutable=mutable)
 end
 
@@ -554,9 +553,9 @@ DNASequence(seq::AbstractString; mutable=false) =
 
 "Construct a DNA nucleotide sequence from other sequences"
 DNASequence(chunk1::DNASequence, chunks::DNASequence...) = NucleotideSequence(chunk1, chunks...)
-DNASequence(seq::Union(Vector{UInt8}, AbstractString); mutable::Bool=false) =
+DNASequence(seq::Union{Vector{UInt8}, AbstractString}; mutable::Bool=false) =
     NucleotideSequence(DNANucleotide, seq, mutable=mutable)
-DNASequence(seq::Union(Vector{UInt8}, AbstractString), startpos::Int, endpos::Int, unsafe::Bool=false; mutable::Bool=false) =
+DNASequence(seq::Union{Vector{UInt8}, AbstractString}, startpos::Int, endpos::Int, unsafe::Bool=false; mutable::Bool=false) =
     NucleotideSequence(DNANucleotide, seq, startpos, endpos, unsafe, mutable=mutable)
 DNASequence(seq::AbstractVector{DNANucleotide}; mutable::Bool=false) =
     NucleotideSequence(seq, mutable=mutable)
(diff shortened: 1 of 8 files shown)


prcastro:
The difference between convertion and construction is still somehow obscure to me. Could we maintain both? The constructor could fall back to the conversion. This way, users wouldn't be confused on what to use.

dcjones:
I agree, I think we should prefer convert functions instead of constructors, at the very least for single argument constructors. Defining convert functions automatically gets you constructors (this is maybe only on 0.4, I don't remember).
```julia
type Foo
    x
end

Base.convert(::Type{Foo}, a, b) = Foo(a + b)

@show Foo(2, 3) # --> Foo(5)
```

prcastro:
Oh, that's nice! If that's the case, let's only define `convert` and let Julia do the job creating constructors.

bicycle1885:
How should I implement conversion from a byte sequence to a biological sequence?
The following thing is the problem:

```julia
function NucleotideSequence{T<:Nucleotide}(::Type{T}, seq::Union(String, Vector{Uint8}),
                                           startpos::Int, stoppos::Int, unsafe::Bool=false)
```
https://github.com/BioJulia/Bio.jl/blob/de004dd0789c75596c1cd4b7bd1ab0d4d6abebf9/src/seq/nucleotide.jl#L267.

I expected that this constructor accepts an encoded byte sequence. For example, I thought `[0x00, 0x01, 0x02, 0x03]` would become `dna"ACGT"`. But actually this constructor accepts an ASCII-encoded byte sequence; `[0x41, 0x43, 0x47, 0x54]` becomes `dna"ACGT"`.

I believe the use case of the current behavior is very limited and want to change it as I expected.
Which would be better?

dcjones:
Maybe these conversion functions should be defined over `Vector{DNANucleotide}` and `reinterpret` used for UInt8, since you can say
```julia

julia> reinterpret(UInt8, [DNA_A, DNA_C, DNA_T, DNA_G, DNA_N])
5-element Array{UInt8,1}:
 0x00
 0x01
 0x03
 0x02
 0x04


julia> reinterpret(DNANucleotide, reinterpret(UInt8, [DNA_A, DNA_C, DNA_T, DNA_G, DNA_N]))
5-element Array{Bio.Seq.DNANucleotide,1}:
 A
 C
 T
 G
 N
```

bicycle1885:
@dcjones I didn't know `reinterpret` is applicable to arrays and it doesn't create a copy. It would be much better than defining over `Vector{UInt8}`. Thank you!

bicycle1885:
Inline documentation of the `convert` methods looks a little bit verbose to me since their functionality is apparent from their types.
Can I remove them if you don't mind?

prcastro:
Agreed. Obvious methods (like `convert`) don't need docstrings.

Ward9250:
Agreed, they don't need doctrings, but will need to be in the expository documentation: as in for example "hey here's all the things you can convert into a DNASequence" and "here's the things you can convert a DNASequence into".

prcastro:
Yes. Since they won't have docstrings, it will be hard to automate the generation of docs. I suggest we make a section of the docs dedicated to these operations.

dcjones:
That sounds fine, we can just have a "constructing sequences" section in the manual.

bicycle1885:
OK, I think it's ready to be reviewed.

Here is the summary:
* Removed lots of constructors and defined corresponding `convert` methods instead.
* Added conversion between biological sequences and vectors of nucleotides/amino acids.
* Small refactoring of code.

I'd like to squash all commits, so please do not merge it now.

bicycle1885:
Oops, I didn't know the mutable sequence type is merged. I'll rebase this branch on the current master.

bicycle1885:
To avoid more conflict, I'll wait for #91 to be merged then finish this one.

dcjones:
@bicycle1885 Could you rebase this when you have a chance?

bicycle1885:
Finished to rebase this on the master branch.
The commit log will not make sense :stuck_out_tongue: , so I'll squash commits into one.
This is ready to be reviewed.

bicycle1885:
Cleaned up the log of my commits.

src/seq/nucleotide.jl
```diff
@@ -253,8 +250,7 @@ end
 
 # Return the number of UInt64s needed to represent a sequence of length n
 function seq_data_len(n::Integer)
-    d, r = divrem32(n)
-    return d + (r > 0 ? 1 : 0)
+    return cld(n, 32)
```
dcjones: Nice, I didn't know about `cld`.

dcjones:
Looks good! :shipit:

prcastro:
What's the advantage of using `falses` over using `BitArray(n)` to create BitArrays? On my machine (79 days old master) they give:

```julia
julia> @time falses(10)
22.705 microseconds (9 allocations: 320 bytes)
```

and

```julia
@time BitArray(10)
10.405 microseconds (7 allocations: 272 bytes)
```

These times seems to be consistent.

dcjones:
`falses(n)` will set all the bits to 0, whereas `BitArray(n)` leaves them uninitialized. What I was doing before was `BitArray(n)` followed by `fill!(x, false)`. If you time it with the `fill!` its they are about the same.

prcastro:
Oh, I see. If that's the case, I'm merging.

bicycle1885:
Thanks!

PR #94, opened 2015-09-10
Add Codecov support.

dcjones:
Codecov has some nice features that coveralls does not, so we might as well use both.


prcastro:
:+1:

dcjones:
I like it: https://codecov.io/github/BioJulia

Ward9250:
:+1:

PR #97, opened 2015-09-13
Code refactor

prcastro:
* Removed Compat module
* Removed Docile module
* `uint`/`Uint` -> `UInt`
* Split `export`/`using`/`import` into multiple lines
* Removed some `Base.` qualifications (at least for now)
*  `export`/`using`/`import` are now only used in module files
* `Bio` module doesn't import anything anymore
* `String` -> `AbstractString`

**Questions**: Can I use this PR to change module files to begin with uppercase (#63)? Also, can I move both `Ragel` and `StringField` modules into `src/tools`?


bicycle1885:
LGTM :+1: 

> Can I use this PR to change module files to begin with uppercase (#63)?
This is a good chance to do that.

> Also, can I move both Ragel and StringField modules into src/tools?
I think it's okay.

dcjones:
Looks good :+1:

I agree that this would be a good time to change file name cases. Honestly I don't see the need for a `tools` directory, but don't feel too strongly about it.

dcjones:
I'm not sure, but I think the original intention with `tools` may have been code for interfacing with externals tools, like BLAST.

Ward9250:
:shipit: I'll make sure module files on my feature branches get renamed to uppercase in that case, just to make things easier. Tools was empty for a long time, and even now it's only got something on my branches. Because it's largely unused, I think this may be a time to decide - ok so what goes in tools, and include what is decided in the docs. For external programs, because A). There is a need to download sources and compile, and B). because nobody has an immediate need for it, or PR where they've tried to include any, I suggested a while back that an addon to Bio.jl with wrappers for compiled programs or other C libraries might be better as a seperate package and repo, a BioExtra.jl or some such thing.

prcastro:
Let's think about moving things to `src/tool` later. Now it's ready to review/merge.

bicycle1885:
All check passed, you all agreed, :shipit:

PR #100, opened 2015-09-14
Round trip tests for reading/writing. And some bug fixes.

dcjones:
This adds tests that reads examples, writes them, then reads them back. This tests that contents of what we are reading and writing is well-formed.

It also fixed a few things that were caught by these tests:
  * Amino acid sequences were missing two residues when read from fasta.
  * fasta and fastq files print trailing space when there is no description.
  * Solexa fastq quality scores (which can be negative) were not encoded correctly when written.

### DIFF
--- a/src/StringFields.jl
+++ b/src/StringFields.jl
@@ -155,6 +155,13 @@ function hash(field::StringField, h::UInt64)
 end
 
 
+function (==)(a::StringField, b::StringField)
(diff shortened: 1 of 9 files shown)


prcastro:
I believe this is pretty safe to merge.

dcjones:
Yep, should be safe.

PR #101, opened 2015-09-15
BAM Parser

dcjones:
This is my work in progress on a BAM parser. <del>It's currently about 10% slower than the one in htslib, but will probably be at parity with a little tweaking.</del> Performance is now indistinguishable from htslib/samtools, but with much simpler code.

TODO:
  <del>* [ ] Data structure to read sequence into: I thought I would just use `DNASequence`, but I need to figure out what to do with `*` and `=` in the sequence field.</del>
  <del>* [ ] Data structure to read cigar strings into.</del>
  <del>* [ ] Data structure to read auxiliary data into.</del>
  * [x] `readname(::BAMAlignment)`
  * [x] `qualities(::BAMAlignment)` and `qualities!(::BAMAlignment, ::Vector{Int8})` to extract qualities.
  * [x] `sequence(::BAMAlignment)` and `sequence!(::BAMAlignment, ::DNASequence)` to extract sequences.
  * [ ] `last(::BAMAlignment)` so alignments can be treated as intervals.
  * [ ] `cigar(::BAMAlignment)`
  * [x] `auxiliary(::BAMAlignment)` and `auxiliary(::BAMAlignment, tag)`.
  * [ ] Iteration for `BAMParser`
  * Write BAM
    * [ ] BGZF output stream
    * [ ] `write(::BAMWriter, ::BAMAlignment)`
  * Tests.
    * [ ] Harvest test cases for BioFmtSpecimens
    * [ ] Round trip tests.

I ended up not using ragel. Parsing BAM is mostly reading fixed sized fields: there's very little matching involved, just counting bytes for which ragel isn't much of a help. SAM will use ragel, but BAM is more important, so that will be another PR.

### DIFF
--- a/src/BGZF.jl
+++ b/src/BGZF.jl
@@ -64,48 +64,44 @@ Read the next compressed BGZF block into `output`.
 """
 function read_bgzf_block!(source::BGZFSource)
     input = source.input
-    output = source.compressed_block
 
     # read header up to xlen
-    p = 1
-    id1 = output[p] = read(input, UInt8); p += 1
-    id2 = output[p] = read(input, UInt8); p += 1
-    cm  = output[p] = read(input, UInt8); p += 1
-    flg = output[p] = read(input, UInt8); p += 1
+    id1 = read(input, UInt8)
+    id2 = read(input, UInt8)
+    cm  = read(input, UInt8)
+    flg = read(input, UInt8)
 
     if id1 != 0x1f || id2 != 0x8b || cm != 0x08 || flg != 0x04
         throw(MalformedBGZFData)
     end
 
-    for _ in 1:8
-        output[p] = read(input, UInt8); p += 1
-    end
-
-    xlen = (Int(output[p - 1]) << 8) | Int(output[p - 2])
+    seekforward(input, 6)
+    xlen = Int(read(input, UInt16))
     if xlen < 6
         throw(MalformedBGZFData)
     end
 
     # read extra subfields
-    nb = readbytes!(input, output, p, p + xlen - 1)
+    nb = seekforward(input, 4)
+    bsize = Int(read(input, UInt16))
+    nb += 2
+    nb += seekforward(input, xlen - 6)
     if nb != xlen
         throw(MalformedBGZFData)
     end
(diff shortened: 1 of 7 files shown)


Ward9250:
For CIGAR Strings you can use the structures I'm working on in SequenceAlignment? I have a cigar string type - seperate to the anchors stuff.

Ward9250:
Just to check my understanding - Ragel would still work though no? If BAM is written to in bytes and characters are just bytes...

dcjones:
Maybe I'll try to cherry-pick or rebase to use your cigar type.

Ragel would work. I started writing it in ragel, but quickly realized all the rules looked like:
```
refid = any{4};
pos = any{4};
bin_mq_nl = any{4};
flag_nc = any{4};
...
```
It's just a little silly and inefficient since it's not testing for anything, just reading n bytes.

breichholf:
I've been curious, and wanted to see if I can put the BAM parser to work. However, it doesn't seem to work for me:

```
julia> bamFile = "test.bam"
"test.bam"

julia> for seq in open(bamFile, BAM)
       println(seq)
       end
ERROR: MethodError: `convert` has no method matching convert(::Type{Any})
This may have arisen from a call to the constructor Any(...),
since type constructors fall back to convert methods.
Closest candidates are:
  convert{T}(::Type{T}, ::T)
  call{T}(::Type{T}, ::Any)
 in start at /Users/reichholf/.julia/v0.4/Bio/src/Ragel.jl:287
 in anonymous at no file
```
Am I trying to access the bam file wrongly?

I thought I'd dig in to the code and see if there's anything glaringly obvious missing. Keep in mind, I'm pretty new to Julia, so I have no clue if I'm on the right track here:
The offending line in Ragel.jl is in `function start` for Iterators and parsers. When I compared bam.jl with the plain text parsers I noticed it's missing a definition of `eltype`. Is this required? Including it, in an identical manner to how it is in for BED or FASTQ files doesn't help.

breichholf:
Nevermind, iteration is obviously not implemented at the moment. Must have missed that in the list above.

Are BAM-specific `start`, `next` and `done` functions the only thing that is required for iteration?

bicycle1885:
Superseded by #265.

Issue #103, opened 2015-09-19
[RFC] Next-gen Read QC library


kdmurray91:
Hi all,

As a bit of a learning experience, I'd like to implement a next-gen read QC library. Feature brainstorming:
- [sicke](https://github.com/najoshi/sickle)-like windowed quality trimming
- Needleman-Wunsch based paired-end read merge/trimmer (a la [libqc++](https://github.com/kdmurray91/libqcpp/blob/master/src/qc-adaptor.cc#L50))
- [scythe](https://github.com/vsbuffalo/scythe)-like Bayesian single-end adaptor trimmer
- Fastqc-style quality/bias/gc content/kmer enrichment measurement

Overall features
- All reports for a qc pipeline available as a single YAML file, which can be compiled into an HTML report
- read-block based parallelism (i.e, chunks of ~2k reads read, processed in parallel, written)

Possible features:
- A Illumina v4 (or is it v5) chemistry specific trimmer, to remove stretches of high phred score poly-G from reads, which you get if a cluster falls of the flow cell, resulting in a very nice clear, dark spot.

Feel free to suggest any other features. The actual implementation of this stuff is probably going to be a bit slow, as I'll be busy for the next month and a bit. But I'll tinker in my spare time :smile:.

Cheers,
K

Ward9250:
@kdmurray91 Does window quality trimming involve sliding windows? They are used quite a lot measuring various population genetic parameters across genomes - identifying regions of selection, recombination and so on. If so an interface for moving across sequences and the like would be something I'm interested in helping dar the specs for and help coding.

kdmurray91:
@Ward9250 yep, you basically move across a read (with qualities), trimming when the window's average quality score is below a threshold.

Windows are also used in k-mer analysis, so I think some generic, zero-copy `windows(size, thing)` function would be very helpful. And with type dispatch could probably handle the QC case, where you want matched windows across both sequence and phred scores.

dcjones:
This would be a nice application. A lot of the functionality needed to do this is in place, and developing it will help us figure out what's missing.

> Windows are also used in k-mer analysis, so I think some generic, zero-copy windows(size, thing) function would be very helpful.

We actually already have `each` where you can do `each(DNAKmer{10}, seq)`, iterating over every 10-mer in a sequence without copying or allocating anything. Something else is needed for quality scores, which are plain `Int8` vectors.

Ward9250:
[...]

PR #104, opened 2015-09-21
[WIP] Window iterator

kdmurray91:
Hi all,

This PR implements an iterator over `K`-length windows of a (initially) `Vector{T}`. This is required for (and discussed in) #103 

I intend to implement unit tests, and eventually extend this to work on things that aren't vectors.

Cheers,
K

### DIFF
--- a/src/util/windows.jl
+++ b/src/util/windows.jl
@@ -1,0 +1,41 @@
+import Base:
+    start,
+    next,
+    done
+
+# Iterate through every window in type
+immutable EachWindowIterator{T}
+    vec::Vector{T}
+    K::Int
+    step::Int
+    npos::Int  # Number of windows, length(vec) - K + step
+
+    EachWindowIterator(vec, K, step=1) = new(vec, K, step, length(vec) - K + 1)
+end
+
+immutable EachWindowIteratorState{T}
+    i::Int  # start position of current window
+end
+
+function eachwindow{T}(K::Integer, vec::Vector{T}, step::Integer=1)
+    @assert K >= 0 "K (window size) must be ≥ 0 in EachWindow"
+    @assert step >= 1 "step must be ≥ 1"
+
+    return EachWindowIterator{T}(vec, K, step)
+end
+
+function start{T}(it::EachWindowIterator{T})
+    i = 1
+    npos = length(it.vec) - it.K + it.step
+    return i
+end
+
+function next{T}(it::EachWindowIterator{T}, state::Integer)
+    i = state
+    value = sub(it.vec, i:i+it.K - 1)
+    return value, i + it.step
+end
+
+function done{T}(it::EachWindowIterator{T}, state::Integer)
+    return state > it.npos
+end
--- a/src/util/windows.jl
+++ b/src/util/windows.jl
@@ -19,18 +19,18 @@ function eachwindow{T}(K::Integer, vec::Vector{T}, step::Integer=1)
     return EachWindowIterator{T}(vec, K, step)
 end
 
-function start{T}(it::EachWindowIterator{T})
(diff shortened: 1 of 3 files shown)


Ward9250:
Awesome, later today I might get time to have a play.

kdmurray91:
The following should work

```julia
v = [1:30;]
for win in eachwindow(5, v)
    println(mean(win))
end
```

src/util/windows.jl
```diff
@@ -0,0 +1,36 @@
...
+end
+
+function eachwindow{T}(K::Integer, vec::Vector{T}, step::Integer=1)
+    @assert K >= 0 "K (window size) must be ≥ 0 in EachWindow"
+    @assert step >= 1 "step must be ≥ 1"
+
+    return EachWindowIterator{T}(vec, K, step)
+end
+
+function start{T}(it::EachWindowIterator{T})
```
Ward9250: Can we @inline these functions? I recall that inlining iterator methods increases performance by a not negligible margin.

kdmurray91: Oh yeah, I forgot about that. I originally had them inline, but I was on a train and julia-0.4 wasn't building properly so I reverted to 0.3 and had to remove the `@inline`. Restored in dd52090

kdmurray91: Doh, wrong branch. make that 40a1903

bicycle1885:
I think there is room for generalization:

* We may be interested in a specific region in a sequence. So, adding start and stop positions to the iterator type and sliding within this range would be useful. Setting start position to 1 and stop position to the end of a sequence as default values would be reasonable.
* Sequence type can be generalized. The current type is limited to the `Vector{T}` type but we may want to iterate other type of sequences like `AminoAdidSequence` type. Parameterizing the sequence type is one way to do that. For example:

```julia
immutable EachWindowInterator{S}
    seq::S
    K::Int
    ...
```

kdmurray91:
@bicycle1885 thanks for the comments!

> - We may be interested in a specific region in a sequence. So, adding start and stop positions to the iterator type and sliding within this range would be useful. Setting start position to 1 and stop position to the end of a sequence as default values would be reasonable.

Could this be done using `for w in eachwindow(5, seq[4:20]) ... end`? 

About your second point, I completely agree, I intend to generalise it to anything which implements `sub(x, i::UnitRange)`, which is what I currently use to generate windows. Which I think would mean that we need to implement `sub(seq, i:UnitRange)` for any sequence type we care about going through in windows, which I'd say would be all of them (including `SeqRecord` and subclasses). Does that sound right?

Ward9250:
I would probalby do it something like:

`eachwindow(seq, range::Range, size::Int, step::Int)`

bicycle1885:
> Could this be done using for w in eachwindow(5, seq[4:20]) ... end?

Exactly. Never mind my comment :sweat_smile: 

About the second point, we need careful consideration.
Given that `sub` creates a view object of an underlying object, I think we need to keep mutability of a sequence. That means a view of a mutable sequence is also mutable and mutating the view object also should mutate the original sequence.
This may cause problems because a view object may accidentally change an immutable sequence.

Ward9250:
Will mutability be a problem for our seq sequences? Aren't they immutable by convention, but able to be set as mutable. So behaviour depends on that.

Ward9250:
How are things progressing with this? I've been given a task that involves sliding windows, so if you don't have time to finish this, I can try and take it furthur.

kdmurray91:
@Ward9250 I probably don't have much time for the next week or so, but regardless I'd love to see how someone more experienced with Julia improves upon my work. So go for it mate!

Ward9250:
The related redux pull request #116, has been completed and merged.

PR #106, opened 2015-09-27
Regenerate parsers with ragel 7.

dcjones:
The good news: Ragel 7 is available now with official Julia support: http://colm.net/news/2015/09/26/first-dev-version-of-ragel-7-released.html

The bad news: there were some glitches I had to sort out so we have to use a fork for now: https://github.com/BioJulia/ragel7

As far as I can tell, there's not any major change in performance. This version doesn't make use of the `@switch` statement macro, so I removed that dependency (performance might improve for these parsers if Julia got a real switch statement). The generated code became somewhat longer, but I think that's just because there's more superfluous `begin`/`end` blocks. Overall this PR should be safe and shouldn't have a noticeable difference other than the comforting feeling of using a more modern and supported version of ragel.


Blahah:
Perhaps my comments above don't matter since the files are auto-generated, but I do think it would be nice to be able to read the code. Apart from those things, happy to merge.

dcjones:
It would be nice to be able to read the generated code, but lacking any kind of a "julia fmt" tool I don't know of good way to automate that. It's weird that it replaces all our spaces with tabs.

As for newlines, I think it would be nice to have an option, but I don't think behavior should vary depending on platform, since other tools may not accept windows newlines.

src/seq/fasta.jl
```diff
@@ -38,482 +38,560 @@ typealias FASTARNASeqRecord       RNASeqRecord{FASTAMetadata}
...
-const _fastaparser_error  = convert(Int , 0)
-const _fastaparser_en_main  = convert(Int , 6)
+_fastaparser_start  = 6
+_fastaparser_first_final  = 6
+_fastaparser_error  = 0
+_fastaparser_en_main  = 6
+__fastaparser_nfa_targs = Int8[ 0, 0 ,  ]
+__fastaparser_nfa_offsets = Int8[ 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 ,  ]
+__fastaparser_nfa_push_actions = Int8[ 0, 0 ,  ]
+__fastaparser_nfa_pop_trans = Int8[ 0, 0 ,  ]
```
bicycle1885: `const`s are removed from these variables. Is that okay in terms of performance?

dcjones: They should be const, though there isn't a big performance impact. Fixed with https://github.com/BioJulia/ragel7/commit/ba684f9ecb0bcb5c318d640ee7e5f34cdd8ce767

bicycle1885: Fine :+1:
:shipit: if tests passed.

PR #107, opened 2015-09-28
Annotations

Ward9250:
Adding a flexible but type-stable container for storing annotations for Phylogenies, Sequences, and so on. Ideas are taken largely from DataFrames discussion and prototype code on making DataFrames type stable.


dcjones:
You're branching from a pretty old version of master here, you might want to rebase before it gets harder.

Ward9250:
Weird, could have sworn I pulled changes from upstream into my master a few weeks ago.

dcjones:
Oh, by "old" I meant prior to #97 which was actually a couple weeks ago.

Ward9250:
Ok I sorted it out. Hopefully this will solve the issue of what if annotations are indeed needed in perf critical code. This is easier to see as a possiblity in evo software I've used that build up and manip rather complex models of evolution around phylogenies with much info attached. A few more commits to come e.g. making sure you can't add several fields of the same name, but it's largely there.

Ward9250:
The more I think about this - the more I think we might put it in a seperate package - GenericList or some such thing - I imagine it's an opportunity to make an awesome R-esque list-like datastructure, that can be indexed by name, by index, popped, pushed and so on.

Ward9250:
I'm happy to take review and comments on this now.

Ward9250:
Ok, so the test suite is currently falling down at the `@annotation` macro, for example, after pasting the Annotations module in REPL, and then `using Annotations`, and trying the following:

```
frank = collect(1.0:5.0)
john = [true, false, false, true, true]
@annotations(:a = frank, :b = john)

ERROR: UndefVarError: frank not defined

# Try to figure out why:
macroexpand(:(@annotations(:a = frank, :b = john)))

:(Annotations.AnnotationContainer((Annotations.frank,Annotations.john),Annotations.Tuple{:a,:b}))
```

The macro is just supposed to make it easer than writing out the constructor in full everytime, but instead of generating such a line i.e. `AnnotationContainer((frank, john), Tuple{:a,:b})`, it sticks `Annotations.` in front of everything like a pain. I don't know why it's doing this, I don't want it to, I just want it to take the variable names as they are and splice them into the constructor!

bicycle1885:
Looks useful in many situations :+1: 
As for the interface, the `@annotations` may be improved further:
* The name looks a little bit long if this macro is intended to be used frequently. IMHO, `@annot` or `@annots` might be better.
* Other options of the syntax:
    1. function style: `@annotations(a = 1, b = "foo")`
    2. dictionary style: `@annotations(a => 1, b => "foo")`

Any thoughts?

Ward9250:
@bicycle1885 I think those are good suggestions, I think because the abstract type `Field{a}` is parametric, you could have anything used as keys, not just symbols, because you can have Field{:a}, `Field{"howdy"}`, `Field{5}` and so on. So long as the macro deals with them properly. So for `@annotations(a = 1, b = "foo")`, we'd have to either convert `a` and `b` to `'a'` or `:a`, to avoid it being misinterpreted as some variable.

Ward9250:
Changed the macro name to `annots`, as suggested. If anyone knows how I can stop the macro doing that annoying thing with variable names as decribed in my next to last comment that would be awesome, I'm still trying to figure it out.

src/annotations.jl
```diff
@@ -0,0 +1,110 @@
...
+
+
+"Get any variable from an annotation container variable field, whilst achieving type stability."
+@generated function getindex{D, L, a}(annotation::AnnotationContainer{D, L}, ::Type{Field{a}})
+    index = findfirst(parameters(L), a)
+    return :(annotation.data[$index])
+end
+
+"Get any variable from an annotation container variable field, whilst achieving type stability."
+getindex{D, L, a}(annotations::AnnotationContainer{D, L}, i, ann::Type{a}) = annotations[field][i]
```
dcjones: What is `field` in `annotations[field][i]`?

src/annotations.jl
```diff
@@ -0,0 +1,110 @@
...
+    end
+end
+
+"Calling Annotations() without any arguments generates a NoAnnotations object."
+AnnotationContainer() = NoAnnotations()
+
+"Generated function outer constructor for extending AnnotationContainer in a functional programming manner."
+@generated function AnnotationContainer{D, L, A, a}(annotations::AnnotationContainer{D, L}, addition::A, ::Type{Field{a}})
+    L2 = Tuple{parameters(L)..., a}
+    println(L2)
```
dcjones: Presumably these `println`s are just for debugging and can be removed.

dcjones:
I wonder if this really solves your concern about type stability. If you have annotations in a node type like
```julia
type Node
    annotations::AbstractAnnotation
    other_stuff
end
```
Then it still has to dynamically determine the type of `annotations`. Of course you could say
```julia
type Node{A <: AbstractAnnotation}
    annotations::A
    other_stuff
end
```

Then if you know the concrete type of a node you know the annotation type. But if you have a tree structure.
```julia
type Node{A <: AbstractAnnotation}
    annotations::A
    children::Vector{Node}
end
```

The concrete types of the children aren't known and have to be determined dynamically. I think you'll end up in a situation where the only way to completely avoid dynamic type resolution is to make the tree very rigidly defined, as in, every node has the exact same type of annotations:
```julia
type Node{A <: AbstractAnnotation}
    annotations::A
    children::Vector{Node{A}}
end
```

Maybe that's ok, but it's something to consider.

The other consideration with this type of annotations, is there's not a way to efficiently update them, since the values are stored in a tuple. That could be worked around by storing a tuple of references, but we should think about whether that's something we want to support.

dcjones:
I should also add that doing dynamic type resolution is not necessarily a defect. I think it's more or less unavoidable if you want to store annotations of arbitrary type.

If you were implementing this in C, you would still be doing dynamic type resolution, just implementing it manually with enums and casts. For example, here's a json encoder written in C where they dispatch on type dynamically with a big switch statement: https://github.com/esnme/ultrajson/blob/master/lib/ultrajsonenc.c#L768 That sort of thing isn't uncommon, we can just do it automatically in julia.

Ward9250:
@dcjones You are right about if Nodes had annotation types. I don't think that node types should store annotations like this. Rather the annotations type is a field in the Phylogeny type, which can be parameterised. Then the idea is that the annotations field will have the dictionaries that map nodes, to values. It may be that we don't need this - but I hope that at least for phylogenies it's useful. DataFrames faced a similar problem not knowing the type in advance of any columns of the DF, in R it doesen't affect anything because it's not strongly typed. This solution was well recieved there, but theres the hope that staged types might be a possible solution now we have staged functions. 

With this there's the possibility that aliases can be added for the most common types of phylogenetic model and the tuple of annotations they will contain - Cladogram, Phylogeny, BD-Model, and so on, as a personal preference I'd like if annotations were stored and manipulated in a type like this, than by convention and telling the compiler Any, but then maybe I should trust the compiler more - I just dread the scenario of indxing or getting annotations from a phylogeny being a drag, because it's the sort of thing some applications will want to do a lot.

Ward9250:
This is now in Phylo.

Issue #96, opened 2015-09-13
Importing functions from the Base module.


bicycle1885:
In Bio.jl, we often extend functions exported from the `Base` module. But as @dcjones said (https://github.com/BioJulia/Bio.jl/pull/85#discussion_r39339225), it is annoying to synchronize method definitions with the `import Base` statement.

I listed four possible styles for importing names to extend. pros and cons are my opinion, you may not agree.
The last style is a suggestion from my side.

### 1. always add `Base.` to definitions like `function Base.foo()`

* pros:
    * explicit
    * no need to keep synchronizing method definitions with the import statement
* cons:
    * verbose when defining methods

### 2. explicitly import methods from `Base` in every module (the current style)

* pros:
    * explicit
    * less verbose when defining methods
* cons:
    * need to keep synchronizing method definitions with the import statement

### 3. `importall Base` (suggested by @prcastro https://github.com/BioJulia/Bio.jl/pull/85#discussion_r39340739)

* pros:
    * less verbose when defining methods
* cons:
    * implicit
    * possible to accidentally extend methods

### 4. define an extendable method list shared across `Bio` modules

This is something like this:

```julia
macro bio_base_import()
    quote
        # these methods can be extended
        import Base:
            convert,
            length,
            ...
    end
end
```
, and call this macro in each modules:

```julia
module Bio.Seq

@bio_base_import

...
```

* pros:
    * explicit
    * less verbose when defining methods
* cons:
    * possible to accidentally extend methods as `importall Base` style, but somewhat less chance to do

dcjones:
Thanks for opening this. I started out preferring to import everything at the beginning of the module (option 1), but my opinion has gradually changed as I wrote more julia code, and now I'd rather just qualify things with `Base.` (option 2).

The problem with explicit imports is that when extending a method, you have to switch to another file, look through a long list of names to see if it's already there, then add it to the list if it isn't. Then when reading code it's not immediately clear if a method is being extended. To know for sure, you have to again switch to another file and look through a list of names. If you forget to import the method, your code can fail in subtle ways when some unrelated code dispatches to the wrong function.

With `importall` (option 3), you no longer have the option of defining a function that has the same name as something in base, but doesn't extend the base method. It's rare that you might want to do that, but I think it does happen. We would also have to `importall` from multiple modules, because sometimes we are extending methods not from Base but from other packages.

[...]

PR #108, opened 2015-09-29
import hash from Base

bicycle1885:
I'm inclined believe that appending the `Base.` prefix is a better solution (ref: #96).


bicycle1885:
I'll merge this since this is an easy fix.

PR #110, opened 2015-10-08
Align Redux

dcjones:
This PR is basically #36, squashed, pared down, and hacked on a little. My goal is to get a basic alignment representation merged that we can all work from (e.g. my bam parser, Kenta's aligner, BLAST wrappers, more sophisticated Align functionality). There's a lot removed from Ben's work, but I think we can add stuff back in once this is merged.

## Alignment Representation

For those that didn't follow the discussion in #36, let me try to describe the alignment representation proposed here. It's somewhat unusual, but I think it's our best bet at having one alignment type that can represent any alignment efficiently and compactly.

The easiest way to think of this representation is as an encoding of a path in a matrix like that used in Smith-Waterman or Needleman-Wunsch (or Leveshtein or Wagner-Fischer, or a million other names the algorithm for computing edit distances goes by).

Say we are aligning `ACGT` against `ACCTGT`. We would set up a matrix like this:

```
    0 1 2 3 4 5 6
      A C C T G T
0   . . . . . . .
1 A . . . . . . .
2 C . . . . . . .
3 G . . . . . . .
4 T . . . . . . .
```

The optimal path might then look like
```
    0 1 2 3 4 5 6
      A C C T G T
0   ↘ . . . . . .
1 A . ↘ . . . . .
2 C . . → → → . .
3 G . . . . . ↘ .
4 T . . . . . . ↘
```
Which corresponds to the alignment
```
ACCTGT
AC--GT
```
The alignment representation used here works by storing just the vertices on this path where it starts, stops, or changes direction, almost like you are drawing lines in a graphics program, like
```
    0 1 2 3 4 5 6
      A C C T G T
0   o . . . . . .
1 A . . . . . . .
2 C . . o . o . .
3 G . . . . . . .
4 T . . . . . . o
```

So you would store an array of "anchors": `(0,0), (2,2), (2,4), (4,6)`, which is enough to know the full alignment assuming only match, insert, and delete operations are allowed. Furthermore, this representation is sorted by both reference and query position, so you can map between positions in O(log n) by binary search, something you can't do with CIGAR strings or run-length encoded gaps.

In global alignments, the start and stop anchors are always `(0,0)` and `(m,n)` but we store them anyway, since it let's us store local alignments in the same representation and differentiate between the two.

# Anchor semantics

Some alignment formats have a richer set of operations beyond just insertions, deletions, and matches. To capture this, we associate an operation with each anchor. For convenience these operations are just CIGAR operations from SAM/BAM, which we'll extent if need be. With attached operations, we can for example differentiate between deletions and skips (like across introns in rna-seq).

That's basically the plan. Let's nail this down so we can start building stuff on it.


bicycle1885:
Wow, great work, @dcjones ! I'll review it later.

Ward9250:
Looks really good :+1: , I wish I knew you were doing it, I wouldn't have made the commits I have since in an attempt to address comments made on 36. 

Briefly these involved making CIGAR a small immutable type that wrapped a String variable. The idea being it is intermediate between string and the full alignment representation - you know the CIGARString has been checked of format and is fine to pass through to construct an alignment. 

Operation was changed to be called OpKind, and now an operation was a small immutable with an OpKind and a size (Int). 

These operations could then be stored in either a an OperationArray (like the Gaps Array but that would support all the operations, not just gap+non-gap), or they could be stored in a sorted array of OperationAnchors (AlignmentAnchors, just renamed). Essentially two structures, like Seqan's but that work with more operations than just gaps and non-gaps (you could just still use gap and non gap with OP_GAP and OP_MM). All those changes are in commit 9e47574. One note on the naming of the OP_* constants. I named them origionally to be slightly more generic than the CIGAR ops, so as the representation was not nessecerily directly tied to CIGAR (You probably noticed I had some non CIGAR OPs in there), although it's fairly obvious CIGAR has been the starting point for this (for example converting op to char uses the characters used in CIGAR!).  

But this looks realy neat, good job!

src/align/Align.jl
```diff
@@ -0,0 +1,57 @@
...
+    >=,
+    *
+
+import Base.Order: lt
+import Base.Multimedia: MIME, writemime
+
+export Operation,
+    CIGAR,
+    CIGARS,
+    @cigar_str,
```
bicycle1885: These are exported, but not defined.

src/align/anchors.jl
```diff
@@ -0,0 +1,271 @@
...
+
+
+# Basic operators for AlignmentAnchors
+# -------------------------------------
+
+"""
+Print to screen or other IO stream, a formatted description of an
+AlignmentAnchor.
+"""
+function show(io::IO, anc::AlignmentAnchor)
```
bicycle1885: This docstring looks verbose to me because it is apparent.

src/align/anchors.jl
```diff
@@ -0,0 +1,271 @@
...
+            for _ in anchors[i-1].refpos+1:anchors[i].refpos
+                write(io, '.')
+            end
+        elseif isinsertop(anchors[i].op)
+            for _ in anchors[i-1].seqpos+1:anchors[i].seqpos
+                write(io, '-')
+            end
+        elseif isdeleteop(anchors[i].op)
+            for _ in anchors[i-1].refpos+1:anchors[i].refpos
+                write(io, '.')
```
bicycle1885: Is there any reason you chose a dot character for delete op?

src/align/anchors.jl
```diff
@@ -0,0 +1,271 @@
...
+        elseif isdeleteop(anchors[i].op)
+            for _ in anchors[i-1].refpos+1:anchors[i].refpos
+                write(io, '-')
+            end
+        end
+    end
+end
+
+
+immutable AlignedSequence{S <: Sequence}
```
bicycle1885: I think this `<: Sequence` restriction is too strict, we can align, for example, `ASCIIString`s.

test/align/TestAlign.jl
```diff
@@ -2,5 +2,149 @@ module TestAlign
...
+            end
+            anchors[i] = AlignmentAnchor(u.seqpos, u.refpos, v.op)
+            anchors[j] = AlignmentAnchor(v.seqpos, v.refpos, u.op)
+            @fact_throws Alignment(anchors)
+        end
+
+        # cigar string round-trip
+        for _ in 1:100
+            #path = random_alignment(rand(1000:10000), rand(1000:10000))
+            path = random_alignment(rand(40:80), rand(40:80))
```
bicycle1885: Did you forget to get this line back?

bicycle1885:
Looks good, but give me a little bit more time to check if this woks well with my pairwise alignment algorithms.

Ward9250:
I'm also pretty sure that the sorting and searching functionality I wrote which uses the Julia sorting API will have to go back in. The conversion between alignment space and source sequence space was really important for other functionality.

src/align/anchors.jl
```diff
@@ -0,0 +1,271 @@
...
+Also stores the position in the alignment view of the sequences, and the
+corresponding position in the unaltered source sequence (nucleotide or protein).
+
+It stores the position values as Int types and the alignment operation is stored
+as a type `Operation`, these are defined in the file `operations.jl` and loaded
+into Bio.Align namespace as a series of global constants.
+
+Calling the AlignmentAnchor default constructor with default arguments sets both
+integer fields to 0, and the operation field to the global constant OP_INVALID.
+"""
```
Ward9250: There is no default constructor. Was it intended for the constructor to be carried from #36, or that the constructor not be brought over and the docstring edited?

bicycle1885:
Let me clarify two points about specs.

The first is about anchoring positions.
Let's think this global-global alignment:

         122234
    seq: AC--GT
         ||  ||
    ref: ACCTGT
         123456

If we represent it with the `Alignment` type, it will be:

    Alignment([
        AlignmentAnchor(0, 0, OP_START),
        AlignmentAnchor(2, 2, OP_MATCH),
        AlignmentAnchor(2, 4, OP_DELETE),
        AlignmentAnchor(4, 6, OP_MATCH)
    ])

And this global-local alignment:

            1123
    seq:    T-CG
            | ||
    ref: TTATACGTG
         123456789

is represented as:

    Alignment([
        AlignmentAnchor(0, 3, OP_START),
        AlignmentAnchor(1, 4, OP_MATCH),
        AlignmentAnchor(1, 5, OP_DELETE),
        AlignmentAnchor(3, 7, OP_MATCH)
    ])

So, alignment anchors point to the ending positions of its operation, the first anchor (which is `OP_START`) always points to the position just before the alignment starts, and the last alignment anchor points to the position where the alignment ends.
Do I correctly understand the concept?

The second is about the mismatch operation.
When we encounter a mismatch of pairwise alignment, in `Alignment`, should we encode it as `OP_SEQ_MISMATCH` or `OP_MATCH`?

    seq: ACGT
         | ||
    ref: AGGT
          ^
          OP_SEQ_MISMATCH or OP_MATCH

I know conventional CIGAR strings do not use 'X' and '=' operations, but I think there is no reason not to use them for internal representation.
When we need a conventional CIGAR string, we can easily convert the full alignment operations into a CIGAR string.

src/align/anchors.jl
Blahah: I think we should consider new users when writing docstrings - things that seem obvious to us are often no obvious to a beginner. My vote is that we err on the side of including more.

Ward9250:
@bicycle1885 You are correct about the anchors point to the end of the operation when you say.

> So, alignment anchors point to the ending positions of its operation, the first anchor (which is
> OP_START) always points to the position just before the alignment starts, and the last alignment
> anchor points to the position where the alignment ends.
> Do I correctly understand the concept?

Except here @dcjones has added an  OP_BACK - now I'm unfamiliar with BACK - I don't know if it means the end of the alignment, as in [OP_START, ... other ops, ..., OP_BACK]. Im also not sure if OP_START is always meant to be at position (0,0) or not. I'm also unsure whether youre global-local alignment should include some skip operations.

As for whether we should be as explicit as possible i.e. OP_SEQ_MISMATCH & OP_SEQ_MATCH over just OP_MATCH. My vote is for yes absolutely. This will allow calculations like finding polymorphisms be faster, whereas with OP_MATCH you still have to go through each site in the MATCH and check their states are the same by checking the underlying sequences. With the explicit operations, the info is there in the alignment.

src/align/anchors.jl
bicycle1885: True, but the `show` function is defined in the `Base` module and is prevalent in every code (http://julia.readthedocs.org/en/latest/stdlib/io-network/#Base.show). It is hard to draw a line between documented and undocumented functions. So, whether the function is already documented in the standard library would be a reasonable criterion to this decision.

bicycle1885:
@Ward9250 Thank you. I thought `OP_START` should point to the starting position of an alignment, and it can point any position in local-local and global-local alignments. According to the specs of the SAM format, skip operation (`OP_SKIP`) seems to be intended to represent an intron in mRNA mapping:

screen shot 2015-10-09 at 22 35 03
https://samtools.github.io/hts-specs/SAMv1.pdf

I also like the idea to distinguish matching and mismatching operations based on the equivalence of sequence symbols. I think this kind of semantic should be clarified in the source code for later reference.

bicycle1885:
The more I read the specs of the SAM format, the more I strongly believe that the alignment match (`OP_MATCH` or 'M') should not be used in our internal alignment representation. The sequence match and the sequence mismatch are more specific and can be mapped to the alignment match just before writing a CIGAR string to a BAM/SAM output.

src/align/anchors.jl
Blahah: I think the guiding principle should be whether a new user (new to both Julia and BioJulia) could understand how to use the function. If we are going to depend on the base documentation, we will have to link to it where appropriate.

src/align/anchors.jl
dcjones: The `.` here is actually a stand in for sequence characters. It was an arbitrary choice, it could be something else.

src/align/anchors.jl
dcjones: Good point, I'll let `S` be anything.

test/align/TestAlign.jl
dcjones: Yes :grimacing:

dcjones:
> Except here @dcjones has added an OP_BACK - now I'm unfamiliar with BACK

It was a proposed, but non-standard operation that is nevertheless included in samtools/htslib, so I put it here for completeness. (The proposal is [here](http://sourceforge.net/p/samtools/mailman/message/28469949/)) It represents moving backwards on the reference sequence, so basically the opposite of OP_SKIP. I'm not sure if we should really support it beyond having an encoding for it.

> So, alignment anchors point to the ending positions of its operation, the first anchor (which is OP_START) always points to the position just before the alignment starts, and the last alignment anchor points to the position where the alignment ends.
Do I correctly understand the concept?

This is correct, I should have been more clear in my description.

> The more I read the specs of the SAM format, the more I strongly believe that the alignment match (OP_MATCH or 'M') should not be used in our internal alignment representation.

Do you mean get rid of `OP_MATCH`, or just not use it when we generate alignments. I agree with using the more specific operations, but I want to retain the ability to work with CIGAR losslessly, which means being able to represent a non-specific match.

dcjones:
> I'm also pretty sure that the sorting and searching functionality I wrote which uses the Julia sorting API will have to go back in. The conversion between alignment space and source sequence space was really important for other functionality.

It is important, but I wasn't 100% sure about the interface, and there's still a lot of work to do to properly test it, so I excluded it in the interest of getting the core representation figured out quickly.

dcjones:
Marking this as WIP, because I realize it really needs a page in the manual before it should be merged.

src/align/anchors.jl
dcjones: Now that I think about it, adding docstrings to `show` methods might not clarify anything, since the base docstring pretty much tells you what you need to know.
```
help?> show
search: show showall showerror showcompact @show @Show Cshort Cushort print_shortest searchsorted searchsortedlast searchsortedfirst flush_cstdio SharedVector

  show(x)

  Write an informative text representation of a value to the current output stream. New types should overload show(io, x) where the first argument is a stream. The representation used by show generally includes Julia-specific formatting
  and type information.

```
I don't think that's true of all Base functions. Sometimes we need more specific documentation.

Blahah: I am thinking of the documentation generated for the site - does Judo fill in the base documentation?

dcjones: Oh, I see. It doesn't now, but it easily could!

Blahah: ;) https://github.com/dcjones/Judo.jl/issues/18

bicycle1885:
> Do you mean get rid of OP_MATCH, or just not use it when we generate alignments.

No, not include `OP_MATCH` as a result of our alignment algorithm. So, I think we reach agreement on it.

src/align/anchors.jl
prcastro: The default constructor is generated automaticaly if no inner constructor is defined by the developer.

src/align/anchors.jl
```diff
@@ -0,0 +1,271 @@
...
+        # optionally check coherence of the anchors
+        if check
+            # empty alignments are valid, representing an unaligned sequence
+            if !isempty(anchors)
+                if anchors[1].op != OP_START
+                    error("Alignments must begin with on OP_START anchor.")
+                end
+
+                for i in 2:length(anchors)
+                    if anchors[i].refpos < anchors[i-1].refpos ||
```
bicycle1885: Checking all anchors are valid operations would be nice.

dcjones: Done.

dcjones:
I think this is ok to merge now. Tests are failing only on 0.5 for mysterious reasons. I don't think we should worry about 0.5 until it's farther along.

bicycle1885:
Looks Good :+1: :shipit:

bicycle1885:
Does anybody have objections to this pull request? I'd like to merge this and work on pairwise alignment.

Ward9250:
No objections from me! - Is the pairwise algorithm the Branch and Bound one that was talked about a while back, or is that something else?

bicycle1885:
@Ward9250 Pairwise alignment, I meant, is now work in progress in this pull request (#111). Since it is based on this pull request, I'd like to merge this pull request to the master branch before going further.

dcjones:
@Ward9250 was speaking of [fogsaa](https://github.com/BioJulia/fogsaa). That's still something we should consider including in some form. It seems like it could be well suited for global alignments of relatively long sequences.

Ward9250:
Awesome, so what shall we do now - the old PR and my branches still exist, I guess I'll keep them around because we'll want some of the stuff from them later.

dcjones:
Yeah, keep it open even though it's not directly mergable now. There's more we should take from it, though we'll probably have to do so is separate PRs.

bicycle1885:
Ah, I missed @kjenova's work. The FOGSAA algorithm looks really awesome and would be promising for aligning long-read sequences like PacBio.

PR #112, opened 2015-10-22
fix printing of anchor

bicycle1885:
I think this is a typo. Please merge this if so, @dcjones.


dcjones:
Definitely a type, thank!

PR #113, opened 2015-10-22
fix firstref and lastref

bicycle1885:
If I read [this comment] (https://github.com/BioJulia/Bio.jl/blob/93a83c11edcff057849ea84e41d5aa82ce308131/src/align/anchors.jl#L77-Lundefined) correctly, this would be the intended behavior.


bicycle1885:
@dcjones Can you check this patch when you have time?

dcjones:
That's better, thanks!

PR #115, opened 2015-11-03
Add more amino acids

bicycle1885:
The current amid acid set is too strict to handle real-world data sets.
I've found this problem while parsing protein sequences downloaded from [Swiss-Prot](http://www.uniprot.org/downloads).
In 549,646 sequences, 'B' is included in 138 sequence, 'Z' in 101 sequences, 'O' in 29 sequences, and 'U' in 254 sequences and these letters are not allowed in the current amino acid set.
These abbreviations are defined in IUPAC and UniProt follows this (http://www.uniprot.org/help/sequences).
You can also find the list in [INSDC](http://www.insdc.org/documents/feature_table.html) (see "7.4.3 Amino acid abbreviations").

With these two commits, we will be able to parse all of the sequences included in Swiss-Prot.

Edit: One open problem is about adding 'J' (Leucine or Isoleucine). This is not included in Swiss-Prot, but listed in INSDC.


Ward9250:
The checks fail, but on inspection it looks like that's because of the nightly builds and not 0.4. The nightly on OSX gets terminated because of "No output has been received in the last 10 minutes"... for some reason. But I'm satisfied this isn't due to the changes made here from what I can see. :+1:  :shipit:

bicycle1885:
Thank you. I'm going to merge this this weekend if there is no objections.

diegozea:
I'm also in favor of dealing with non standard aminoacids and ambiguities +1
Swiss-Prot is a curated data set and maybe there aren't  'J's because of that.

bicycle1885:
I'm incline to think that 'J' should be added, too; someone will almost certainly encounter the same problem.

bicycle1885:
CI failure is due to the nightly build. I'll merge.

PR #116, opened 2015-11-07
Sliding windows - redux

Ward9250:
Modified slightly the sliding windows code of @kdmurray91 - instead of requireing a vector of something, it is parametric and works fo any type T (instead of Vector{T}) that implements the sub() method. I added the sub() method for nucelotide sequences too - which is the same as constructing a subsequence with `seq[x:y]`. 

A rebase was done so thought it was easier to just open a new PR and then close both, than try and merge or push back to the origional PR. But @kdmurray91's contribution is reflected in the commit history.

### DIFF
--- a/src/seq/nucleotide.jl
+++ b/src/seq/nucleotide.jl
@@ -689,6 +689,7 @@ end
 # Construct a subesequence
 getindex{T}(seq::NucleotideSequence{T}, r::UnitRange) = NucleotideSequence{T}(seq, r)
 
+sub{T}(seq::NucleotideSequence{T}, r::UnitRange) = NucleotideSequence{T}(seq, r)
 
 function setindex!{T}(seq::NucleotideSequence{T}, nt::T, i::Integer)
     if !seq.mutable
--- a/src/util/windows.jl
+++ b/src/util/windows.jl
@@ -1,0 +1,41 @@
+import Base:
+    start,
+    next,
+    done
+
+# Iterate through every window in type
+immutable EachWindowIterator{T}
+    vec::Vector{T}
+    K::Int
+    step::Int
+    npos::Int  # Number of windows, length(vec) - K + step
+
+    EachWindowIterator(vec, K, step=1) = new(vec, K, step, length(vec) - K + 1)
+end
+
+immutable EachWindowIteratorState{T}
+    i::Int  # start position of current window
+end
+
+function eachwindow{T}(K::Integer, vec::Vector{T}, step::Integer=1)
+    @assert K >= 0 "K (window size) must be ≥ 0 in EachWindow"
+    @assert step >= 1 "step must be ≥ 1"
+
+    return EachWindowIterator{T}(vec, K, step)
+end
+
+function start{T}(it::EachWindowIterator{T})
+    i = 1
+    npos = length(it.vec) - it.K + it.step
+    return i
+end
+
+function next{T}(it::EachWindowIterator{T}, state::Integer)
+    i = state
+    value = sub(it.vec, i:i+it.K - 1)
+    return value, i + it.step
+end
+
+function done{T}(it::EachWindowIterator{T}, state::Integer)
+    return state > it.npos
+end
--- a/src/util/windows.jl
+++ b/src/util/windows.jl
@@ -15,7 +15,7 @@ immutable EachWindowIterator{T <: arrayOrStringOrSeq}
     function EachWindowIterator{T}(data::T, width::Int, step::Int = 1)
         @assert width >= 1 "Window width must be ≥ 1."
         @assert step >= 1 "step must be ≥ 1."
-        @assert width < length(data) "The window size cannot be greater than number of data elements."
+        @assert width <= length(data) "The window size cannot be greater than number of data elements."
         return new(data, width, step)
     end
 end
--- a/src/util/windows.jl
+++ b/src/util/windows.jl
@@ -58,6 +58,16 @@ end
     return window, i + it.step
 end
 
+# Extra next method to account for fact than strings don't have sub method
+# like arrays and Nucleotide sequences do.
+# The operation is an indexing operation, rather than a "proper"
+# substring or subarray operation.
+@inline function next{T <: AbstractString}(it::EachWindowIterator{T}, state::Integer)
+    i = state
+    window = it.data[i:i + it.width - 1]
+    return window
+end
+
 @inline function done(it::EachWindowIterator, state::Integer)
     return (state + it.width - 1) > length(it.data)
 end
--- a/src/util/windows.jl
+++ b/src/util/windows.jl
@@ -8,7 +8,7 @@ An iterator which moves across a container, as a sliding window.
 Constructor requires the data to move the window across, the width of the
 window, and the step of the window.
 """
-immutable EachWindowIterator{T <: arrayOrStringOrSeq}
+immutable EachWindowIterator{T <: ArrayOrStringOrSeq}
     "A reference to the collection to iterate over."
     data::T
     "The width of the sliding window."
@@ -56,7 +56,7 @@ Convienient function for constructing an EachWindowIterator.
 
 Accepts the same arguments as the EachWindowIterator constructor.
 """
-function eachwindow{T <: arrayOrStringOrSeq}(data::T, width::Int, step::Int = 1)
+function eachwindow{T <: ArrayOrStringOrSeq}(data::T, width::Int, step::Int = 1)
     EachWindowIterator{T}(data, width, step)
 end
 
--- a/src/util/windows.jl
+++ b/src/util/windows.jl
@@ -19,18 +19,18 @@ function eachwindow{T}(K::Integer, vec::Vector{T}, step::Integer=1)
     return EachWindowIterator{T}(vec, K, step)
 end
 
-function start{T}(it::EachWindowIterator{T})
+@inline function start{T}(it::EachWindowIterator{T})
     i = 1
     npos = length(it.vec) - it.K + it.step
     return i
 end
 
-function next{T}(it::EachWindowIterator{T}, state::Integer)
+@inline function next{T}(it::EachWindowIterator{T}, state::Integer)
     i = state
     window = sub(it.vec, i:i+it.K - 1)
     return window, i + it.step
 end
 
-function done{T}(it::EachWindowIterator{T}, state::Integer)
+@inline function done{T}(it::EachWindowIterator{T}, state::Integer)
     return state > it.npos
 end
--- a/src/util/windows.jl
+++ b/src/util/windows.jl
@@ -1,4 +1,10 @@
-# Iterate through every window of size K in vec
+
+import Base:
+    start,
+    next,
+    done
+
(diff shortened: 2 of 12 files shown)


src/util/windows.jl
```diff
@@ -0,0 +1,36 @@
...
+immutable EachWindowIterator{T}
+    data::T
+    K::Int
+    step::Int
+    nwin::Int  # Number of windows, length(vec) - K + step
+
+    EachWindowIterator(data, K, step = 1) = new(data, K, step, length(data) - K + 1)
+end
+
+immutable EachWindowIteratorState{T}
```
bicycle1885: This type is not used anywhere.

bicycle1885:
Looks good, but we need tests for it :+1:

kdmurray91:
@Ward9250 Thanks!

src/util/windows.jl
Ward9250: Oh yes I see, I'll remove it.

Ward9250:
@bicycle1885 Thanks! Ok so the two items to do before merge then are to add some docstrings, and then some simple tests to ensure the iterator works as intended. I'll see if I can get both of those things finished tonight so as we can merge tomorrow.

Ward9250:
I think this is ready to merge if people are happy to ok it: I added some tests and docs.

vchuravy:
This looks good, but could you clean up the git history a bit?

Ward9250:
I'll do a rebase and squash some stuff.

Ward9250:
Did a squash - is this ok or do you want me to go back furthur?

src/util/windows.jl
```diff
@@ -0,0 +1,53 @@
...
+    width::Int
+    "How many positions the sliding window moves along."
+    step::Int
+    "The number of windows that will result from iterating across the container."
+    nwin::Int
+    "The number of elements in the container that are missed. Typically because
+    of the width and step size of the window: some elements near the end of the
+    container are missed if window sizes and step sizes are too large."
+    nmissed::Int
+end
```
bicycle1885: Do we really need the `nwin` and `nmissed` fields? I think the `width` and `step` fields are enough to iterate over `data`.

bicycle1885: Plus, I thought docstrings cannot be attached to the fields of a type; comments would be better.

src/util/windows.jl
```diff
@@ -0,0 +1,53 @@
...
+
+"""
+Convienient function for constructing an EachWindowIterator.
+
+Accepts the same arguments as the EachWindowIterator constructor.
+"""
+eachwindow{T <: AbstractArray}(data::T, width::Integer, step::Integer = 1) = EachWindowIterator(data, width, step)
+
+@inline function start{T <: AbstractArray}(it::EachWindowIterator{T})
```
bicycle1885: This type parameter constraint can be safely removed if you don't access the type parameter in the body of a function. (e.g. `function start(it::EachWindowIterator)`)

src/seq/nucleotide.jl
```diff
@@ -679,6 +679,7 @@ end
 # Construct a subesequence
 getindex{T}(seq::NucleotideSequence{T}, r::UnitRange) = NucleotideSequence{T}(seq, r)
 
+sub{T}(seq::NucleotideSequence{T}, r::UnitRange) = NucleotideSequence{T}(seq, r)
```
bicycle1885: Missing `sub` for amino acid sequence?

Ward9250: Thanks for reminding me! AA sequences sometimes slip my mind because I tend to work with DNA.

src/util/windows.jl
Ward9250: The Julia documentation says you can attach docstrings to fields:
"
```
"..."
type T
    "x"
    x
    "y"
    y
end
```
Adds docstring "..." to type T, "x" to field T.x and "y" to field T.y. Also applicable to immutable types."

Ward9250: I thought nmissed would be useful for some purposes. In the past I've wanted to know how many positions are not covered by a sliding window run before.

src/util/windows.jl
Ward9250: Ahhh I have wondered about this in the past i.e. when to use MyType instead of MyType{T}. Thanks!

src/util/windows.jl
bicycle1885: > The Julia documentation says you can attach docstrings to fields:

Wow! [You're right!](http://julia.readthedocs.org/en/latest/manual/documentation/#types).

> I thought nmissed would be useful for some purposes. In the past I've wanted to know how many positions are not covered by a sliding window run before. Likewise, nwin might be useful to know in advance how many windows are going to be generated.

Agreed, but in general I prefer "minimal fields" to avoid complexity. Since these values can be computed on demand with cheap cost, I think creating functions for such values would be a better approach.

Ward9250: That sounds like a good alternative. I'll implement that.

src/seq/nucleotide.jl
Ward9250: This has been added now.

Ward9250:
Ok, sub method added for amino acid sequences, and moved the two fields from the iterator to their own methods.

src/util/windows.jl
```diff
@@ -0,0 +1,66 @@
+"""
+An iterator which moves across a container, as a sliding window.
+"""
+immutable EachWindowIterator{T <: AbstractArray}
```
bicycle1885: This type parameter constraint is too strict to apply the window iterator to biological sequences.

test/util/TestUtil.jl
```diff
@@ -0,0 +1,54 @@
...
+                itr = eachwindow(seq, winsize, stepsize)
+                res = collect(itr)
+                @fact length(res) --> size(itr)
+                for win in res
+                    @fact length(res) --> winsize
+                end
+            end
+        end
+    end
+end
```
bicycle1885: I want to see realistic examples like `eachwindow(dna"ACGTAT", 2)` and `eachwindow(aa"APQTLL", 2)` in test cases.

src/util/windows.jl
Ward9250: Ah yeah I forgot to ask about this - what do people think of making the Sequence abstract type inherit from AbstractArray? I think it makes some sense, a sequence can be thought of as an ordered array of nucleotides, it has getindex, setindex, length, start, next, done and most of the other methods set out in the documentation on the [Array Interface](http://docs.julialang.org/en/release-0.4/manual/interfaces/)

bicycle1885: Yeah, I think we can make the `Sequence` a subtype of `AbstractArray`. But I feel the `Sequence` is more like a subtype of `AbstractString` than `AbstractArray`, though I never see clear semantic distinction between them. At the moment, this type constraint would be too stringent unless the Julia community reaches a consensus about semantics of these abstract types.

Ward9250: I see what you mean, because you can argue that a string is an ordered array of characters of a given alphabet, and in C this really is the case. Can type unions be used to sort this? I would have any T so long as it inherits from either Sequence, AbstractArray, or AbstractString?

bicycle1885: Yes, you can use `T<:Union{Sequence,AbstractArray,AbstractString}` as a type constraint.

Ward9250:
@bicycle1885 Tests have been added for DNA, RNA, and AA sequences as suggested, and all tests now pass for OSX and Linux. AppVeryor is a different story, but I believe the reason it fails is independent of this PR. 

Tested is the constructors of the iterators for correct construction, the number and size of the windows produced, and finally the content of the windows themselves is tested for correctness.

test/intervals/TestIntervals.jl
```diff
@@ -183,7 +181,9 @@ facts("IntervalCollection") do
 
 
     context("Show") do
-        nullout = open("/dev/null", "w")
+        nullstream = @windows ? "NUL" : "/dev/null"
+        nullout = open(nullstream, "w")
```
bicycle1885: Here I think you can use the `DevNull` object, which was broken on v0.4.1 but is now available on v0.4.2.

src/util/windows.jl
```diff
@@ -0,0 +1,92 @@
+
+arrayOrStringOrSeq = Union{AbstractArray, AbstractString, Sequence}
```
bicycle1885: Please capitalize and use `typealias`.

bicycle1885:
I commented a few thing I noticed; otherwise looks good to me :+1:

test/intervals/TestIntervals.jl
Ward9250: I can't write anything to DevNull:
```
julia> write(DevNull, "Hi")
ERROR: Base.DevNullStream does not support byte I/O
 in write at io.jl:70
 in write at ascii.jl:99
 in eval at ./boot.jl:264

julia> show(DevNull, 15)
ERROR: Base.DevNullStream does not support byte I/O
 in write at io.jl:70
 [inlined code] from ascii.jl:99
 in show at show.jl:143
 in eval at ./boot.jl:264
```
The docs only show it as used in the run command, e.g. `run(`cat test.txt` |> DevNull)`

bicycle1885:
```
julia> write(DevNull, "Hi")
2

```

Do you use Julia v0.4.2? It works on my computer. The behavior you saw is a bug I fixed :)

Ward9250:
```
Bens-MacBook-Pro:~ axolotlfan9250$ julia
               _
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: http://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.5.0-dev+1259 (2015-11-12 19:47 UTC)
 _/ |\__'_|_|_|\__'_|  |  Commit af36998 (45 days old master)
|__/                   |  x86_64-apple-darwin14.5.0

```
I generally pull master and run make on it when I install Julia, as a result I think I'm on v0.5.

src/util/windows.jl
Ward9250: Fixed :)

test/intervals/TestIntervals.jl
bicycle1885: You're using too old master to use `DevNull`; my patch was merged on Nov 20. (https://github.com/JuliaLang/julia/pull/13758).

Ward9250: Changed the test to DevNull.

Ward9250:
Thanks @bicycle1885, I'd like to merge this as soon as the Travis completes without errors. I'll wait until tomorrow and do it if there are no further objections.

bicycle1885:
:shipit: 

AppVeyor fails only on 32-bit platforms; maybe due to hitting a limit of integers since it throws `InexactError` (IMHO, we can drop the support of 32-bit platforms because nobody uses them nowadays). Anyway, we can ignore this error for now.

PR #118, opened 2015-11-08
Genotype Count Arrays for Population Genetics. 

Ward9250:
This PR is to track the development of an array type that can be used for population genetics work.

It should be able to be used much like a regular vector/matrix/array, but with additional behaviour that makes it easy to do population genetics work with it. It should also borrow from @dcjones work in the Seq module in regards to how we should do subsetting - allowing the arrays to be immutable by convention, allowing efficient iteration and working on subsets or parts of the whole data, without having to copy data.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -4,3 +4,4 @@ Colors
 IntervalTrees
 Iterators
 Libz
+DataArrays
--- a/src/Bio.jl
+++ b/src/Bio.jl
@@ -8,5 +8,6 @@ include("Ragel.jl")
 include("seq/Seq.jl")
 include("intervals/Intervals.jl")
 include("align/Align.jl")
+include("var/Var.jl")
 
 end # module Bio
--- a/src/var/Var.jl
+++ b/src/var/Var.jl
@@ -1,0 +1,13 @@
+module Var
+
+export GenotypeMatrix,
+
+import Base:
+    convert,
+    show
+
+using DataArrays
+
+include("genotype_array.jl")
+
+end
--- a/src/var/genotype_array.jl
+++ b/src/var/genotype_array.jl
@@ -1,0 +1,123 @@
+
+
+# Genotype Array, a type designed to store genotype counts.
+# Behaves like a named array, and wraps a PooledDataArray.
+# Allows subsetting by Loci names, and by individual names.
+
+# Data is a sparse array containing allele counts.
+# AllelesPerLocus tells how many alleles are at each locus.
+# LocusFactor describes which columns of the matrix, are for each locus.
+# LocusAlleleNames has the names of the alleles for each locus.
+
+type GenotypeArray{S <: AbstractString, N}
+    Individuals::Vector{S}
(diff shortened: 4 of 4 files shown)


kdmurray91:
One thing I mentioned on gitter is:

I'd love to be able to magically translate the locus and sample IDs to some external metadata. I'll illustrate with a use case or two:

- Slice out all loci on chromosome 2 to a new matrix
- Slice away all samples whose age was <=  x
- Give me all loci from any sample in populations A, C or F


All these are *very* common operations where I work, and are currently done using an ungodly fusion of R's `%in%` and `match()`, often with some regex thrown in to deal with free text matching. So, a saner and cleaner version of these operations would be pretty useful...

Ward9250:
Absolutely, we want as many convienient behaviours for this array type as possible, but not too many: I think this array type, and another type for storing SNP genotypes will inherit from the same abstract type. And they will probably be fields in a "higher level type" - for sake of argument let's call it Genotypes, and it will be parametric, and could contain either a GenotypeArray or a SNPsArray, and which one it is shouldn't matter too much to the high-level user. So population and chromosome annotations for example probably belong in the higher type. But this is all stuff I just thought of over the weekend and its all up for discussion as we go - I don't want to be sole dictator of the final result. If you have specific features or things you do in R, and want to do them in Julia, in a more Julian and nicer way, mention them and maybe give a few R lines as an example, and we can talk about how we can do it better with a Julia implementation.

bicycle1885:
Things are still unclear to me; what is different from data frames?
I often use [pandas](http://pandas.pydata.org/) and [GRanges](https://bioconductor.org/packages/devel/bioc/vignettes/GenomicRanges/inst/doc/GenomicRangesIntroduction.pdf) to handle variant data loaded from VCF files. Data type you propose looks similar to these types.

Ward9250:
The types proposed here are more like the classes in the R package 'adegenet' dev version, which is quite popular in my experience, and features in Paradis' book on phylogenetic and evolutionary analysis with R. They allow storage and rapid computation of counts of alleles and allele frequencies. The SNP data type associated with files like VCF stores each SNP using only one bit per ploidy level and allows working on much more data with the available RAM.

It is true that you can represent SNP data with something like GRanges - and I guess using our Intervals/IntervalsTrees type probably works in that respect too. 
Data Frames have been used, the 'pegas' R package extends DataFrames and stores alleles as factors, which I think is more space efficient than adegenet's extended matrix type, but computing frequencies and poulation genetics statistics can require more computation, especially if the statistics require frequency computation. Another limitation of basing the types on DataFrames in julia will probably be that individuals are stored in rows. In addition, the classes in adegenet are highly compatible and operate with classes from phylogenetics, and multivariate stat's packages also by Paradis and similar authors. We'd want the types we use for pop gen to also work well types in Bio.jl related to phylogenetics and so on.

What I want to provide here at first is a bit more specialised than a DataFrame though, and closer to adegenets classes than pegas' classes. In the end though, if a more DataFrame like type is required we could easily have one - it could inherit from the same abstract type as the others.

PR #119, opened 2015-11-11
Modify Interval constructors.

bicycle1885:
Small tweaks. Now we can write `Interval` like the former below:

```julia
Interval("chr1", 10, 20, '+')
Interval("chr1", 10, 20, STRAND_POS)
```

Also, I added the `strand` accessor like `first` and `last`.

One thing I want to add is a sequence name accessor.  `seqname` or just `name`?


bicycle1885:
I'd like to merge this small patch today if no objections.

Blahah:
Looks good to me. In the context of an interval, `name` suggests the name of the interval itself, while `seqname` unambiguously refers to the reference sequence name.

bicycle1885:
Thanks. I added `seqname`, too.

Ward9250:
:shipit:

Blahah:
:shipit:

PR #111, opened 2015-10-11
Pairwise-alignment algorithms

bicycle1885:
I've started to move my [PairwiseAlignment.jl](https://github.com/bicycle1885/PairwiseAlignment.jl) package to the `Bio.Align` module on the top of #110 .
I've finished to relocate the global alignment algorithm, and other data types like substitution matrices.
This is still work in progress, no need to review the source code at the moment, but let me show off you the performance a little bit.

I aligned two long amino acid sequences (titin protein, human and mouse): [Q8WZ42](http://www.uniprot.org/uniprot/Q8WZ42) and [A2ASS6](http://www.uniprot.org/uniprot/A2ASS6).
These have 34,350 and 35,213, respectively. This is the benchmark script:

```julia
using Bio.Seq
using Bio.Align

let
    submat = Bio.Align.BLOSUM62
    gap_open_penalty = 10
    gap_extend_penalty = 1
    affinegap = Bio.Align.AffineGapScoreModel(submat, gap_open_penalty, gap_extend_penalty)
    a = first(open("Q8WZ42.fasta", FASTA)).seq
    b = first(open("A2ASS6.fasta", FASTA)).seq
    @time aln = Bio.Align.pairalign(
        Bio.Align.GlobalAlignment(),
        convert(Vector{AminoAcid}, a),
        convert(Vector{AminoAcid}, b),
        affinegap
    )
    @show aln.score
end
```

The competitor is the [Biostrings](http://bioconductor.org/packages/release/bioc/html/Biostrings.html) package in Bioconductor, whose alignment algorithm is written in [C](https://github.com/Bioconductor-mirror/Biostrings/blob/master/src/align_pairwiseAlignment.c):

```R
library(Biostrings, quietly=T)

a = readAAStringSet("Q8WZ42.fasta")[[1]]
b = readAAStringSet("A2ASS6.fasta")[[1]]
t0 = proc.time()
aln = pairwiseAlignment(a, b, substitutionMatrix="BLOSUM62", gapOpening=10, gapExtension=1)
t1 = proc.time()
print(t1 - t0)
print(score(aln))
```

This is the result:

```
~/.j/v/Bio (pairalign|…) $ julia -O bench.jl
  9.916355 seconds (196.67 k allocations: 1.136 GB, 0.06% gc time)
aln.score = 165611
~/.j/v/Bio (pairalign|…) $ Rscript bench.R ^/dev/null
   user  system elapsed
 14.316   1.235  15.714
[1] 165611
```

[...]

PR #120, opened 2015-11-11
define arithmetic on nucleotides and amino acids

bicycle1885:
This patch makes nucleotides and amino acids more like `Char`s. For example, `DNA_A + 1` is evaluated to `DNA_C` like `'a' + 1 == 'b'`. Total orders are defined, too. So now we can sort these symbols.


bicycle1885:
Can anyone give me comments? I'd like to rebase #111 onto this patch.

Ward9250:
It looks like a good idea to me. Changes look simple and painless too, and I don't know of anyway this interferes with other code in the modules. :shipit:

Blahah:
What defines the ordering? It seems to me there is no natural (biological) ordering of the nucleotides. I think I'd like to see some explanation in comments of why this is necessary and useful, and how the ordering is chosen.

Ward9250:
Good point, I didn't think about the ordering much, just that arithmetic ops on nucleotides is useful.

bicycle1885:
There is no natural ordering, indeed. But without any ordering we cannot order DNA/RNA sequences, neither. A lexicographical order is implicitly assumed in many algorithms, for example, a suffix array is a sequence of indexes of sorted suffixes. Arbitrary ordering would work well as long as it is consistent among algorithms that use the order. The ordering I defined here is implied by the encoding of the nucleotides, which would be the most efficient and easiest way to keep consistency.

Ward9250:
That does make sense if the ordering is consistent, but I suppose actually it's possible to have multiple ordering types and dispatch on those, which Is what I did for alignment anchors on my branch - you could order by their srcPos integer, or their alnPos integer. So if a different ordering in wanted or makes sense in another context it wouldn't be a huge problem.

bicycle1885:
@Blahah Please merge this if you're fine :)

bicycle1885:
I added comments in the code and arithmetic for Kmer. Please review it again :pray:

Blahah:
Thanks! lgtm :shipit:

Ward9250:
Still have my support! :shipit:

bicycle1885:
Thank you!
Again, the CI failure on the nightly build is considered harmless.

PR #126, opened 2015-12-08
Fixing tests and code for AppVeyor

Ward9250:
There was an issue with AppVeyor as a test used /dev/null, which doesen't exist on windows, it has to be NUL instead.

Let's get the AppVeyor badge coloured green!

### DIFF
--- a/test/intervals/TestIntervals.jl
+++ b/test/intervals/TestIntervals.jl
@@ -183,7 +183,9 @@ facts("IntervalCollection") do
 
 
     context("Show") do
-        nullout = open("/dev/null", "w")
+        nullstream = @windows ? "NUL" : "/dev/null"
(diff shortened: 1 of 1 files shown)


Ward9250:
Ok, the test is fixed, now another test fails to do with BED now.

Ward9250:
The lines in question seem to be:

```
while interval_end > res_end
    res_sizes[res_try] += 1
    res_ends[res_try] = res_end = res_end + res_scales[res_try]
end
```

In bigbed_write_blocks, in particular it appears to be one of the `setindex!` operations that is triggering this. @dcjones can you see why this might not work on Windows, I think the second line in the while loop is a bit odd with two assignments on one line.

Ward9250:
Tests now work on Appveyor for 64 bit platforms.

PR #131, opened 2016-02-06
Struc module - framework for parsing PDB files and dealing with protein structures

jgreener64:
This is a work-in-progress PR containing a potential new module for base BioJulia - Struc, which deals with PDB files and protein structures. See the [notebook](https://github.com/jgreener64/Bio.jl/blob/strucmin/doc/struc.ipynb) in docs for an overview; the code also has comments.

Would be interested to hear thoughts on the overall structure of the module - is it doing the right type of things in the right way? The overall approach is heavily based on the Biopython PDB parser.

I am currently working on improving parts of the code, particularly the `organise` and `read` functions. Speed-wise the collect functions and iterations are fast but `organise` and `read` take longer. Also working on writing tests.

Thanks,
Joe


src/struc/model.jl
```diff
@@ -0,0 +1,826 @@
...
+# For example atom names 1H and H1 would be hydrogens but NH1 would not
+"""Determines if an `AbstractAtom` represents hydrogen. Uses the element field
+where possible, otherwise uses the atom name."""
+hydrogenselector(atom::AbstractAtom) = getelement(atom) == "H" || (getelement(atom) == "" && 'H' in getatomname(atom) && !ismatch(r"[a-zA-Z]", getatomname(atom)[:findfirst(getatomname(atom), 'H')-1]))
+
+
+function show(io::IO, struc::Structure)
+    model = struc[1]
+    println(io, "Name                 -  ", getstrucname(struc))
+    println(io, "No models            -  ", countmodels(struc))
```
diegozea: I prefer `N` or `N°` because `No` is confuse.

Blahah: Since there is plenty of whitespace after, how about simply `Number of models`

bicycle1885:
Great work! I really want to use this module when it is merged.

Some general comments about the overall design:

* Which is a better module name: `Bio.Struc` or `Bio.Struct`? I think `Struc` stands for "structure" and the most common abbreviation would be "struct" like C language.
* In Bio Julia, we don't use the `get` prefix for getters. So, `getserial`, `getatomname`, etc. can be renamed to `serial`, `atomname`, etc.
* In PDB format, the character set is limited to non-control ASCII characters. So, I think we can use `ASCIIString`s instead of `AbstractString`s. This will greatly improve the performance thanks to the type stability.

src/struc/model.jl
```diff
@@ -0,0 +1,826 @@
...
+end
+
+Structure(name::AbstractString) = Structure(name, Dict())
+
+
+"""An `Array{AbstractAtom,1}`."""
+typealias AtomList Array{AbstractAtom,1}
+
+"""An `Array{AbstractResidue,1}`."""
+typealias ResidueList Array{AbstractResidue,1}
```
bicycle1885: These type aliases would be useless in many cases because `Vector{Atom}` objects are not an instance of `AtomList`. For example, `isa([1,2,3], Vector{Integer})` is `false` due to invariance of type parameters (http://julia.readthedocs.org/en/latest/manual/types/#parametric-composite-types). In general, using abstract types as type parameters should be avoided.

src/struc/model.jl
```diff
@@ -0,0 +1,826 @@
...
+
+"""An `Array{Chain,1}`."""
+typealias ChainList Array{Chain,1}
+
+"""An `Array{Model,1}`."""
+typealias ModelList Array{Model,1}
```
bicycle1885: I think using `Vector{Chain}` and `Vector{Model}` is more explicit than using these type aliases.

src/struc/model.jl
```diff
@@ -0,0 +1,826 @@
...
+immutable Atom <: AbstractAtom
+    het_atom::Bool
+    serial::Int
+    name::AbstractString
+    alt_loc_id::Char
+    res_name::AbstractString
+    chain_id::Char
+    res_no::Int
+    ins_code::Char
+    coords::Array{Float64,1}
```
bicycle1885: `coords::NTuple{3,Float64}` is generally faster because less memory allocation happens.

src/struc/model.jl
```diff
@@ -0,0 +1,826 @@
...
+which case removes all but one location for disordered atoms."""
+function formatomlist(atoms::AtomList; remove_disorder::Bool=false)
+    # Key is (residue ID, residue name, atom name)
+    # Remove this to a separate function getatomid?
+    atom_dic = Dict{Tuple{AbstractString, AbstractString, AbstractString}, AbstractAtom}()
+    for atom in atoms
+        #@assert typeof(atom) != DisorderedAtom "There is already a DisorderedAtom container in the AtomList"
+        atom_id = (getresid(atom; full=true), getresname(atom), getatomname(atom))
+        # The atom does not exist so we can create it
+        if !(atom_id in keys(atom_dic))
```
bicycle1885: You can use `haskey(atom_dict, atom_id)` here.

src/struc/model.jl
```diff
@@ -0,0 +1,826 @@
...
+            atom_dic[atom_id] = atom
+        # A disordered atom container already exists and the alt loc ID is not taken
+        elseif typeof(atom_dic[atom_id]) == DisorderedAtom && !(getaltlocid(atom) in getaltlocids(atom_dic[atom_id]))
```
bicycle1885: `isa(object, type)` is better than `typeof(object) == type`.

src/struc/model.jl
```diff
@@ -0,0 +1,826 @@
...
+"""Determines if an `AbstractAtom` is a non-hetero atom, i.e. came from an ATOM
+record."""
+stdatomselector(atom::AbstractAtom) = !ishetatom(atom)
+
+"""Determines if an `AbstractAtom` is a hetero atom, i.e. came from a HETATM
+record."""
+hetatomselector(atom::AbstractAtom) = ishetatom(atom)
+
+"""Determines if an `AbstractAtom` has its atom name in the given `Array`."""
+atomnameselector(atom::AbstractAtom, atom_names::Array{AbstractString,1}) = getatomname(atom) in atom_names
```
bicycle1885: This membership checking is linear time. You can use the `Set` type to make it constant time.

src/struc/model.jl
jgreener64: I agree it is ambiguous as stands - will implement a change. Perhaps it is okay to omit N/number entirely? e.g.

```
Name              -  1ABC.pdb
Models            -  1
Chain(s)          -  AB
Residues          -  500
Point mutations   -  4
Other molecules   -  2
```
etc.

jgreener64:
Thanks for great comments @bicycle1885.

In response to general comments:
- Yes I agree `Struct` is probably better, will change.
- Will change getters and setters throughout the code.
- Will switch to `ASCIIString` throughout the code.

src/struc/model.jl
jgreener64: Yeah, will change throughout.

src/struc/model.jl
jgreener64: Great, will change throughout.

src/struc/model.jl
jgreener64: Great, will change throughout.

src/struc/model.jl
jgreener64: Ah interesting. So am I right in thinking `atom_names` should be of type `Set{ASCIIString}` (and then change `backbone_atom_names` to `Set`s)? Or should the function take an array and do `getatomname(atom) in Set(atom_names)`?

src/struc/model.jl
jgreener64: Ah, okay. So I should remove these definitions and have `Vector{AbstractAtom}` for the type of function arguments where required? As it stands I think many functions would have to retain `Vector{AbstractAtom}` as the input type because a chain could have some ordered and some disordered atoms on it.

src/struc/model.jl
jgreener64: Cool. Am I right in thinking that making it an `NTuple` would prevent changing the coordinates without re-instantiating, as the type is immutable?
With the coords as an `Array` I was conveniently just changing the elements for `setx!` etc., though this could be hacky and un-Julian, I don't know.
I guess it depends whether we want people to be able to change the atomic coordinates quickly - I suppose if you're doing something like MD simulation you might have a new type, something like:
```
type MDAtom
    atom::Atom
    current_coords::NTuple{3,Float64}
end
```
In this case the `coords` field in the `Atom` type would represent the (invariant) crystal coordinates, and so could happily remain immutable.

src/struc/model.jl
bicycle1885: It is better to use `function func{T<:AbstractAtom}(atoms::Vector{T}) ... end` than directly using `Vector{AbstractAtom}` as argument types. This can achieve type stability and hence performance will be greatly improved.

src/struc/model.jl
bicycle1885: Yes, you are right. So, it may be better to keep this type as a mutable Vector if you may change the coordinates. +1 for your opinion.

src/struc/model.jl
bicycle1885: The `atom_name` argument should be `Set{ASCIIString}`. Constructing a set from an array every time will require more time than the current implementation. Of course, you can keep this method for the ease of use and add `atomnameselector(atom::AbstractAtom, atom_names::Set{ASCIIString})` method for performance.

Blahah:
I'm going to go one further than @bicycle1885 and say we should just call it `Structure`. It's a few extra letters which will be types a few times in any given project, and in return for that price users get to use a real word.

Ward9250:
:+1: for `Structure` @Blahah . :+1: For changes recommended by @bicycle1885 and a big :+1: for the great work you've done here @jgreener64 :)

jgreener64:
Thanks @Ward9250 !

Happy to call it `Structure`, though this means I would need to rename the inner `Structure` type.

This isn't necessarily a bad thing - this type represents a PDB file and contains one or multiple models (conformations of a protein), so `Structure` is potentially ambiguous anyway. Any ideas for new names? I was thinking `ModelGroup`, `PDBEntry`, `PDBFile`, `PDBRecord` or just `Protein`.

Another possibility is `Ensemble` but I would be tempted to steer clear of that as long-term we may want to create a different ensemble type for PCA etc..

src/struc/model.jl
jgreener64: Ah yes, will look into introducing this where possible throughout the code.

Ward9250:
I would err more on the side of naming types that express generic ideas like a DNA molecule or Protein structure such that they are not named based on a format, and so are format agnostic e.g. `NucleotideSequence` is preferable to `FASTASequence` or `FASTQSequence` since a nucleotide sequence is an idea or concept that is not tied to a specific file format. So it depends on the purpose of your types: are you using them to represent a general idea or concept like a Protein structure, or are you using them to essentially be an embodiment of a specific file format?

jgreener64:
Yes, an interesting point. I have encountered that elsewhere in the code and attempted to err on the side of representing the underlying protein rather than the file format. This would suggest `Protein` as a solution.

jgreener64:
Or `ProteinStructure` as you suggest.

bicycle1885:
+1 for `ProteinStructure`. I feel `Protein` is a vague name to represent the coordinates of a protein. In this respect, `ProteinCoordinates` is more specific and may be a better name than `ProteinStructure`.

jgreener64:
So I have a question about setters I was wondering if anyone could advise on.

The `DisorderedAtom` type is immutable but the user may occasionally want to change the `default` parameter, which is a `Char`. I think I am right in saying that it is not possible to write an in-place function `defaultaltlocid!` (or whatever) to change this, as the type is immutable.

In this case, should I have a function `changedefaultaltlocid(disordered_atom::DisorderedAtom, default::Char)` that returns a new `DisorderedAtom` which is the same apart from the new value?

Ward9250:
I'd lean towards defining a new outer constructor for the type: 
```
function DisorderedAtom(atom::DisorderedAtom, default::Char)
  Stuff and things!
end
```

Which constructs a new disordered atom from an old one, replacing the old value with the new one.

jgreener64:
Great thanks.

jgreener64:
So I've had a go at implementing the changes above and have written tests and function docs. Does anyone have any more suggestions or comments on the code, or can we look toward a merge?

Also, re docs. What would be the best way to have a tutorial sort of page, similar to the current docs here:
http://biojulia.github.io/Bio.jl/seq.html

I could relatively easily change my IJulia notebook into this Markdown formatting if the best way to go is another section on the above site.

bicycle1885:
Great work! I will thoroughly check it tomorrow. The current documentation is based on [Judo.jl](https://github.com/dcjones/Judo.jl), but I'm afraid it is no longer maintained. Does anybody have suggestions about documentation?

We maintain data files for tests at https://github.com/BioJulia/BioFmtSpecimens. So, instead of including PDB files in the Bio.jl repository, could you move PDB files to BioFmtSpecimens and create a pull request there?  Creating a PDB directory and then placing the PDB files and a `index.yml` file like https://github.com/BioJulia/BioFmtSpecimens/blob/master/FASTA/index.yml would be enough.

Ward9250:
As to your docs question, this is hopefully something which will be
addressed as part of @blahah's next sprint, refining out documentation,
resources, and ease of getting involved.

Ward9250:
I should clarify - sprint as in sprint of his Mozilla Fellowship, which is
currently ongoing.

Blahah:
that's right about the docs - I've got a couple of other projects that come first in the list as I need them for events coming up, but around late march I should be working on BioJulia docs. @jgreener64 If you document your functions and provide a nice Jupyter notebook and/or a markdown file with code blocks explaining how to do common things, I will convert that into the docs site and loop you in when I get to more elaborate things like interactive examples.

jgreener64:
Okay cool. @bicycle1885 I will create a PR into BioFmtSpecimens for the PDB files.

@Blahah I will look to improve the existing Jupyter notebook and am happy to help when the docs get attention around late March.

src/structure/model.jl
```diff
@@ -0,0 +1,1089 @@
...
+# e.g. allows you to do res[atom_name] rather than res.atoms[atom_name]
+# setindex! should be used with caution as it can lead to inconsistencies
+# e.g. adding an atom to a residue atom dict does not update the atom list
+
+# Accessing a DisorderedAtom with a character returns the Atom with that alt
+#   loc ID
+Base.getindex(disordered_atom::DisorderedAtom, alt_loc_id::Char) = disordered_atom.alt_loc_ids[alt_loc_id]
+function Base.setindex!(disordered_atom::DisorderedAtom, atom::Atom, alt_loc_id::Char)
+    disordered_atom.alt_loc_ids[alt_loc_id] = atom
+end
```
bicycle1885: It would be better to return the modified object in `setindex!`, which is a convention in Julia.

```julia
function Base.setindex!(disordered_atom::DisorderedAtom, atom::Atom, alt_loc_id::Char)
    disordered_atom.alt_loc_ids[alt_loc_id] = atom
    return disordered_atom
end

File: src/structure/model.jl [hunk 9]
```diff
@@ -0,0 +1,1089 @@
...
+Set the coordinates of an `AbstractAtom` to a `Vector{Float64}` of length 3. For
+`DisorderedAtom`s only the default atom is updated.
+"""
+function coords!(atom::Atom, coords::Vector{Float64})
+    @assert length(coords) == 3 "3 coordinates must be given"
+    x!(atom, coords[1])
+    y!(atom, coords[2])
+    z!(atom, coords[3])
+    return nothing
+end
```
bicycle1885: These functions (`x!`, `y!`, `z!`, and `coords!`) also should return the modified object (i.e. `atom`).

src/structure/model.jl
```diff
@@ -0,0 +1,1089 @@
...
+function resid(element::Union{AbstractResidue, AbstractAtom}; full::Bool=false)
+    if ishetero(element)
+        if full
+            inscode(element) == ' ' ? "H_$(resnumber(element)):$(chainid(element))" : "H_$(resnumber(element))$(inscode(element)):$(chainid(element))"
+        else
+            inscode(element) == ' ' ? "H_$(resnumber(element))" : "H_$(resnumber(element))$(inscode(element))"
+        end
+    else
+        if full
+            inscode(element) == ' ' ? "$(resnumber(element)):$(chainid(element))" : "$(resnumber(element))$(inscode(element)):$(chainid(element))"
```
bicycle1885: These lines are too long. Using `if inscode(element) == ' ' ... else ... end` instead of the ternary operator will break long lines here.

src/structure/model.jl
```diff
@@ -0,0 +1,1089 @@
...
+"""
+defaultmodel(struc::ProteinStructure) = struc.models[modelnumbers(struc)[1]]
+
+chainids(struc::ProteinStructure) = countmodels(struc) > 0 ? chainids(defaultmodel(struc)) : Char[]
+
+
+# Sorting functions
+
+# Sort atoms by serial
+Base.sort!{T <: AbstractAtom}(atoms::Vector{T}) = sort!(atoms, by=serial)
```
bicycle1885: It is not a good manner to override the `sort!` method because people will not expect `sort!` for `Vector` is overridden. Moreover, if one passes `rev=true` to sort atoms, this will result in unexpected result. The best way to sort elements by type-specific order is defining `Base.isless` method for that type. So, add `Base.isless(a1::AbstractAtom, a2::AbstractAtom) = isless(serial(a1), serial(a2))` somewhere.

src/structure/model.jl
```diff
@@ -0,0 +1,1089 @@
...
+    applyselectors!(new_list, selector_functions...)
+    return new_list
+end
+
+"Runs `applyselectors` in place."
+function applyselectors!{T <: Union{AbstractResidue, AbstractAtom}}(element_list::Vector{T}, selector_functions::Function...)
+    for selector_function in selector_functions
+        filter!(selector_function, element_list)
+    end
+end
```
bicycle1885: Return `element_list`.

src/structure/model.jl
```diff
@@ -0,0 +1,1089 @@
...
+
+applyselectors{T <: Union{AbstractResidue, AbstractAtom}}(element_list::Vector{T}) = copy(element_list)
+applyselectors!{T <: Union{AbstractResidue, AbstractAtom}}(element_list::Vector{T}) = element_list
```
bicycle1885: I think these two methods can be automatically derived from the above definitions. The `selector_functions...` argument accepts zero arguments.

src/structure/model.jl
```diff
@@ -0,0 +1,1089 @@
...
+protein backbone atom.
+"""
+backboneselector(atom::AbstractAtom) = stdatomselector(atom) && atomnameselector(atom, backbone_atom_names)
+
+"`Set` of protein standard heavy atom names."
+const heavy_atom_names = Set(["C", "CA", "CB", "CD", "CD1", "CD2",
+                    "CE", "CE1", "CE2", "CE3", "CG", "CG1", "CG2", "CH2", "CZ",
+                    "CZ2", "CZ3", "N", "ND1", "ND2", "NE", "NE1", "NE2", "NH1",
+                    "NH2", "NZ", "O", "OD1", "OD2", "OE1", "OE2", "OG", "OG1",
+                    "OH", "SG"])
```
bicycle1885: I want to know where you got this heavy atom list from. Inserting a comment about source like https://github.com/BioJulia/Bio.jl/blob/master/src/seq/geneticcode.jl#L35-L36.

src/structure/pdb.jl
```diff
@@ -0,0 +1,274 @@
...
+            push!(atom_lists[curr_model], parseatomrecord(rstrip(line, '\n'), line_number))
+        # Read MODEL record
+        elseif startswith(line, "MODEL ")
+            try
+                curr_model = parse(Int, line[11:14])
+            catch ArgumentError
+                throw(PDBParseError("Could not read model serial number", line_number, line))
+            end
+            # Create model if required
+            !haskey(atom_lists, curr_model) ? atom_lists[curr_model] = AbstractAtom[] : nothing
```
bicycle1885: `if !haskey(atom_lists, curr_model)` is easier to read.

src/structure/pdb.jl
```diff
@@ -0,0 +1,274 @@
...
+        error("Could not parse to desired type")
+    end
+end
+
+
+"Form a string of a certain length from a value."
+function spacestring(val_in, new_length::Int)
+    string_out = string(val_in)
+    if length(string_out) > new_length
+        return string_out[1:new_length]
```
bicycle1885: Is that safe to truncate `string_out`? It may be better to throw an error.

src/structure/spatial.jl
```diff
@@ -0,0 +1,71 @@
...
+    sq_dists = zeros(count_one, count_two)
+    for i in 1:count_one, j in 1:count_two
+        @inbounds sq_dists[i,j] = (coords_one[1,i] - coords_two[1,j]) ^ 2 + (coords_one[2,i] - coords_two[2,j]) ^ 2 + (coords_one[3,i] - coords_two[3,j]) ^ 2
+    end
+    return sqrt(minimum(sq_dists))
+end
+
+distance(atom_one::AbstractAtom, atom_two::AbstractAtom) = sqrt((x(atom_one) - x(atom_two)) ^ 2 + (y(atom_one) - y(atom_two)) ^ 2 + (z(atom_one) - z(atom_two)) ^ 2)
+
+Base.(:-)(element_one::StructuralElementOrList, element_two::StructuralElementOrList) = distance(element_one, element_two)
```
bicycle1885: I'm not sure whether this is conventional in structural biology softwares, but I think using `-` is an abuse of notation. For example, `residue - atom` looks removing `atom` from `residue` to me.

src/structure/pdb.jl
```diff
@@ -0,0 +1,274 @@
...
+    else
+        return "$(repeat(" ", new_length - length(string_out)))$string_out"
```
bicycle1885: You can use `lpad` here: `lpad(string_out, new_length)`.

bicycle1885:
The overall design looks good. Once comments I added are addressed and PDB files are moved, I think this pull request is almost ready to be merged. Small tweak can be added later as separated pull requests.

src/structure/model.jl
jgreener64: Okay will change throughout.

src/structure/model.jl
jgreener64: Yep will change

src/structure/model.jl
jgreener64: Agreed

src/structure/model.jl
jgreener64: Ah okay that is interesting. Will implement this.

src/structure/model.jl
jgreener64: Yep

src/structure/model.jl
jgreener64: Ah okay

src/structure/model.jl
jgreener64: Okay will do

src/structure/pdb.jl
jgreener64: Agreed

src/structure/spatial.jl
jgreener64: On second thought it is ambiguous - may be better to leave out `-` and just retain `distance`.

src/structure/pdb.jl
jgreener64: I did suspect this function might exist somewhere, thanks.

src/structure/pdb.jl
jgreener64: Yes this would prevent silent errors e.g. if the serial overflows 5 characters.

src/structure/model.jl
jgreener64: Turns out the most common definition is non-hydrogen atoms, so I have implemented it like that.

jgreener64:
Thanks for the comments @bicycle1885 . Let me know if I have made the fixes you had in mind.

bicycle1885:
@jgreener64 I made a pull request to this pull request at https://github.com/jgreener64/Bio.jl/pull/1, which fixes the filepaths of PDB files. Everything else looks good to me :+1: . Is there any comment from others?

Ward9250:
This is not the kind of data I'm used to dealing with, or programming with. So I'm happy to accept both of your recommendations on this one, I am happy to read it and test it on my laptop however.

src/structure/model.jl
```diff
@@ -0,0 +1,1104 @@
...
+    chain_id::Char
+    number::Int
+    ins_code::Char
+    het_res::Bool # Does the residue consist of hetatoms?
+    atom_list::Vector{ASCIIString}
+    atoms::Dict{ASCIIString, AbstractAtom}
+end
+
+# Constructor without atoms
+Residue(name::ASCIIString, chain_id::Char, number::Int, ins_code::Char, het_res::Bool) = Residue(
```
bicycle1885: I missed a problem about argument types. Generally speaking, broad (abstract) argument types are better than specific (concrete) ones. This is specified in the style guide of Julia: http://docs.julialang.org/en/latest/manual/style-guide/?highlight=argument#avoid-writing-overly-specific-types. So for example, this `ASCIIString` type can be replaced with `AbstractString` type, which accepts more various argument values like `SubString{ASCIIString}`, which you can see when you `split` strings. Likewise, `Integer` is better than `Int` in this case.

src/structure/pdb.jl
```diff
@@ -0,0 +1,273 @@
...
+        end
+    # If there is only one model, do not write out MODEL/ENDMDL lines
+    else
+        writepdblines(output, element, selector_functions...)
+    end
+end
+
+writepdb(output::IO, element::StructuralElementOrList, selector_functions::Function...) = writepdblines(output, element, selector_functions...)
+
+function writepdb(filepath::ASCIIString, element::StructuralElementOrList, selector_functions::Function...)
```
bicycle1885: This overly-specific type of `filepath` caused [CI failure](https://ci.appveyor.com/project/Ward9250/bio-jl/build/1.0.108/job/j08aoa6obdmxo7y5) on Windows. Of course, you can replace this type with `AbstractString`.

bicycle1885:
I've added a few comments about argument types. Sorry for missing it.

src/structure/model.jl
jgreener64: Ah yes. I can change in the functions throughout.

Should the types of `Residue.name` and `Residue.number` stay as `ASCIIString` and `Int` respectively, or should they change to the more general type along with the constructor? I guess they can stay specific as values will be converted (e.g. from any `AbstractString` to `ASCIIString`) when the constructor is called?

bicycle1885: Yes, types of fields in `Residue` etc. should stay as-is. Default constructors will convert argument values to match types of fields. So, I think it is enough to change argument types only.

jgreener64:
Have changed argument types as suggested.

If anyone has any other comments or suggestions let me know :+1:

bicycle1885:
I believe it's ready to merge it! :shipit: 

I'm going to merge this one if no objections for a few days.

jgreener64:
Great, thanks.

Ward9250:
LGTM :shipit:

Issue #134, opened 2016-02-22
Stop codons in amino acid sequences


ianpgm:
Is it correct that stop codons (represented as '*') are not considered a valid part of amino acid sequences? See:
```
julia> Bio.Seq.AminoAcidSequence("PLCFRHHQSRYTGKLL*")
ERROR: * is not a valid amino acid
 in convert at /Users/ianmarshall/.julia/v0.4/Bio/src/seq/aminoacid.jl:111
 in AminoAcidSequence at /Users/ianmarshall/.julia/v0.4/Bio/src/seq/aminoacid.jl:214
 in AminoAcidSequence at /Users/ianmarshall/.julia/v0.4/Bio/src/seq/aminoacid.jl:211
 in call at essentials.jl:56
```
Is this intentional for some reason or could it be added to the list of allowed "amino acids"?

bicycle1885:
Yes. Stop codons are not in the list, but we're going to add more amino acids (stop codon and gap) and ambiguous nucleotides. See related issue #122.

Ward9250:
Stop codons are not in Open Reading Frame's if I recall, and so are not really part of the protein, they just represent the end of it. So I can see some sense in the argument, that - of course - * is not a valid AA, but surely the desired behaviour when parsing an AminoAcidSequence from the string above, should just make it from PLCFRHHQSRYTGKLL, and when it hit's *, just stop. Either way, @bicycle1885 is right, the BioSequence overhaul we have planned will cater to this and several other issues.

ianpgm:
Thanks for the responses! True, for the example I gave it might be best to just leave off the stop codon. The program I'm building is exactly as you describe - importing short metagenomic reads and translating them in six reading frames to find out which of the reading frames is most likely to encode a protein. So I would expect to encounter a number of stop codons within the sequence that I would want to ignore at first.

Ward9250:
Oh cool I did something similar with an R script a while back. I wish I could have done it in BioJulia then.

Ward9250:
newseq has been merged, and allows for additional amino acids.

Issue #122, opened 2015-11-13
[RFC] More nucleotides


bicycle1885:
I'd like to share and discuss ideas about DNA or RNA sequences, partially mentioned in #55.

Currently, only five nucleotides (A/C/G/T/N) are defined in this package and a DNA sequence cannot store other ambiguous nucleotides like M (A or C). This is because we use 2-bit encoding for A/C/G/T and 1-bit ambiguity mask for N. This limitation will definitely cause problems in real-world bioinformatics. Since we are creating a general purpose infrastructure for bioinformatics, this would be an intolerable limitation.

I think we should, at least, support all nucleotide base codes defined by IUPAC, which includes all possible combinations of four nucleotides (http://www.ncbi.nlm.nih.gov/Class/MLACourse/Modules/MolBioReview/iupac_nt_abbreviations.html).
If we use 4-bit encoding, we can store any nucleotides listed here. These ambiguous nucleotides are, of course, allowed in the Biostrings package of Bioconductor. An apparent drawback of 4-bit encoding is that it requires more space than the current (2+1)-bit encoding: the data size of a DNA sequence will become 4/3 = 1.333 times larger. But I think this will not be a serious problem in most cases because it is very rare to handle such a large sequence that hits the RAM limit except eukaryotic reference sequences. For such long sequences, we can prepare a reference sequence type that adopts 2-bit encoding for A/C/G/T + compressed N mask, which is already implemented in my FMM.jl package (https://github.com/bicycle1885/FMM.jl/blob/master/src/genomicseq.jl).

One advantage of this 4-bit encoding is that it will improve the performance of random access on DNA sequences. This is because the current implementation stores the N mask in a separated bit vector object, and hence it has a branch (`if seq.ns[i]) ... else ... end`) to access an element in a sequence:

```julia
function getindex{T}(seq::NucleotideSequence{T}, i::Integer)
    if i > length(seq) || i < 1
        throw(BoundsError())
    end
    i += seq.part.start - 1
    if seq.ns[i]
        return nnucleotide(T)
    else
        return getnuc(T, seq.data, i)
    end
```

But using 4-bit encoding, we can store all information in a vector so that any branch is not need any more. That would lead to a significant performance improvement in intensive-sequence-access algorithms like pairwise alignment.

Ward9250:
[...]

Issue #132, opened 2016-02-12
Sequences gets parsed as AminoAcidSequence instead of DNASequence


vchuravy:
I was taking a look at http://www.ncbi.nlm.nih.gov/nuccore/NC_010004 and I was surprised that parsing the FASTA file the record was of type AminoAcidSequence instead of DNASequence.

Is there a way to tell the FASTAParser which encoding to use (or to convert a AminoAcidSequence to a DNASequence?)

````julia
using Bio
using Bio.Seq

nc010004 = open("NC_010004.1.fasta", Bio.Seq.FASTA)

for rec in nc010004
  @show typeof(rec.seq)
end

# typeof(rec.seq) = Bio.Seq.AminoAcidSequence
````

vchuravy:
In line 1989 row 54 there is an `'R'` in the input file (although it is labelled as a DNA sequence on ncbi....)

vchuravy:
According to this http://www.bioinformatics.org/sms2/iupac.html, the 'R' encodes an uncertain read. It would be nice for Bio.jl to support that, but that is a different issue then.

vchuravy:
Also see #122

bicycle1885:
@vchuravy Thank you for reporting. This is really what I worried about in #122; DNA sequences in real world are sometimes ambiguous and we often encounter them when handing a large set of sequences. This is the highest priority issue to me and I will tackle it soon when I have time.

vchuravy:
Thank you. Is there anything specific that I could help with?

bicycle1885:
Well, I'd like to know more examples of ambiguous DNA sequences that contain ambiguous nucleotides except 'N'. If you already know or find other examples, can you report them here? I will add them to the test set: https://github.com/BioJulia/BioFmtSpecimens.

breichholf:
Wikipedia lists a total of 16 abbreviations that could occur and still be considered valid DNA code. Realistically, however, in a real world FASTA R and Y will more commonly (and this is already a rare event) occur, and B/D/H/V will rarely (if ever) occur.
You're more likely to see these ambiguous base codes when aligning Sanger sequencing data.

https://en.m.wikipedia.org/wiki/Nucleotide

Ward9250:
Relabelling as enhancement, as bug implies incorrect result in present functionality, whereas this falls more into the category as being a request for additional functionality in Seq.

Issue #129, opened 2016-01-08
Additional array operations for Sequences (insert, delete, push, pop)


kescobo:
Re: [discussion from gitter](https://gitter.im/BioJulia/Bio.jl?at=5681690e0171d1791d0265b4)

Should we have methods to insert and/or delete one or more nucleotides/amino acids to mutable sequences? @Ward9250 [said](https://gitter.im/BioJulia/Bio.jl?at=568be9f38aafa1e4459c419f):

>as a sequence may be thought of as a specialized string, which can be thought of as an array, insert and delete (and push and pop?) methods are something we should consider implementing as efficiently as possible. 

Suggestions?

PR #135, opened 2016-02-25
replace sequence types with BioSequence

bicycle1885:
This pull request introduces a new sequence type: `BioSequence{A<:Alphabet}`. The main purpose of this is to store more various nucleotides in a sequence (see #122 for discussion). The `BioSequence` type is indexed by a subtype of `Alphabet`, which indicates the domain (or set) of biological characters. For example, the old `DNASequence` is now `BioSequence{DNAAlphabet{4}}` (here `DNAAalphabet{4}` is a subtype of `Alphabet` and its type parameter, `4`, indicates the number of encoding bits). At the moment, this pull request is really premature and the tasks below should be done before merge.

- [x] Finish implementation of `BioSequence`.
  - [x] Basic operations
  - [x] Parsers
- [x] Pass tests.
- [x] Add more ambiguous nucleotides (#122) and stop codon (#134).
- [x] Check performance regression.

---
EDIT 2016-03-06

- [x] Add tests.


bicycle1885:
I'm wondering whether `seq[i:j]` and `sub(seq, i:i)` should return a view or copy. The current master branch makes a view for immutable sequences and a copy for mutable sequences, but people may expect `sub` to return a view even for mutable sequences like vectors.

**master**

| mutability | `seq[i:j]` | `sub(seq, i:j)` |
|---|---|---|
| immutable | view | view |
| mutable | copy | copy |

**my suggestion**

| mutability | `seq[i:j]` | `sub(seq, i:j)` |
|---|---|---|
| immutable | view | view |
| mutable | copy | view |

Ward9250:
I think there may be a case to change this, to have sub return a view even
of a mutable sequence. But only if we are clear that, because the user is
calling sub, they are explicitly asking for a view, and if it's a view of a
mutable sequence - well that's their lookout, they should expect changes
may affect things. However, if you use the indexing operation, if the
sequence is immutable, you can safely make the efficiency gains made
possible by using a view instead of an actual copied subset.

Ward9250:
This is great! I've had to spend yesterday and today working on thesis chapter, but over the weekend I will do some work on this and phylo.

Ward9250:
I've been looking at what you've done with Alphabet, and have just about worked out the idea for a macro, which allows people to create new alphabets for BioSequence, without having to code all the code for the type, with all the bitsof's and so on.

bicycle1885:
Cool. It's almost finished to pass tests except a failing one and ones that touch internals I've changed. When all tests passed, I'll squash all commits into one and start to work on ambiguous nucleotides and performance problems.

bicycle1885:
All tests have passed (with minimal fix)! Now I'm going to work on adding more nucleotides.

Ward9250:
Wohooo! I've nearly finished the alphabet builder macro too. Just a few
more tweaks and it will be done tonight.

bicycle1885:
I made a simple benchmark script to check the regression of this branch vs. master: https://gist.github.com/bicycle1885/cf8a09e2bf0f21d563b9.

And the following is the result of current commit (88ae5ee) vs. master branch. A few of them (`access_all` and `fill_A!`) are slightly faster, some are comparable (`convert_asciistring`, `each_3mer_step1`, `each_3mer_step3`), and others are tremendously slower. This would be because I removed some complicated algorithms that are not directly applicable to the new sequence representation, so I'm going to revise them.

benchseq

Ward9250:
Can they be used but just for the given sequence type to which they apply? I would have thought performance of the 2-bit encoding wouldn't change much.

Ward9250:
The macro for helping people create alphabets works for me on my machine now #136 - tests still fail, but it's the same tests as fail for this PR. I think this is going to be great. If we can come up with a simple and standard way for people to create sets of characters, and then an alphabet. We can say, ok for your alphabet you can expect BioSequence{*your alphabet*} to do *list of things here - get, set, indexing, match/mismatch etc* in an efficient manner.

bicycle1885:
> Can they be used but just for the given sequence type to which they apply?

Yes, but now I don't run any test for 2-bit sequences and there may be silly bugs. I'll try it later.

bicycle1885:
> The macro for helping people create alphabets works for me on my machine now #136

It looks cool! I'll check it once this pull request settles down.

Here is the progress of performance improvement:

benchseq

bicycle1885:
OK, now I removed most of performance regressions:

benchseq


I guess 2x slow downs of `count_nucs` and `make_complement` would be almost inevitable because we are handling more types of nucleotides in 4-bit encoding of DNA.

Ward9250:
I think that is fair to expect, more nucleotides, more bits, more time. If
people want faster counting they can use 2-bit encoding. This is really
fantastic stuff.

On Sun, Mar 6, 2016 at 2:15 PM, Kenta Sato (佐藤 建太) <notifications@github.com

bicycle1885:
Thank you :). 
This pull request wil solve #18, #132, #134 in one shot.

I will add more tests for 2-bit sequences and make examples like gaps and stop codons to https://github.com/BioJulia/BioFmtSpecimens.

bicycle1885:
I've added many operations addressed in #129: `resize!`, `push!`, `pop!`, `shift!`, `unshift!`, `append`, `insert!` and `deleteat!` for `BioSequence`. Now sequences behave more like vectors.

Further, I've found that removing mutable/immutable state of sequences simplifies stuff without losing performance. `seq[i:j]` always creates a lightweight view of `seq`. Copying underlying data happens only when we try to modify a sequence whose data are shared with other sequences. This is something like a [copy-on-write](https://en.wikipedia.org/wiki/Copy-on-write) strategy and it is used when [forking a child process in Linux Kernel](http://linux.die.net/man/2/fork).

Ward9250:
Copy on write is also the mechanism at work in R. It works well unless you
want to modify something many many times, in which case it sucks, but our
seq types are different, because it's only copy on write when data is
shared. I like this!

On Tue, Mar 8, 2016 at 6:52 PM, Kenta Sato (佐藤 建太) <notifications@github.com

bicycle1885:
I've added the [MurmurHash3](https://en.wikipedia.org/wiki/MurmurHash)-like function. Though written in pure Julia, this is as fast as builtin the `hash` function of Base (written in C) for enough long sequence (benchmark script is here: https://gist.github.com/bicycle1885/78a5fb65dea6b3ac4a4d):

```
~/.j/v/Bio (newseq|✚3…) $ julia benchhash.jl
ASCIIString
  0.000765 seconds
  0.000923 seconds
  0.000936 seconds

DNASequence (aligned)
  0.000493 seconds
  0.000468 seconds
  0.000497 seconds

DNASequence (unaligned)
  0.000941 seconds
  0.000960 seconds
  0.000913 seconds
```

This enables to create dictionaries whose keys are our sequences.

Now I've finished to implement functionality I wanted in the new sequence type, so I'll start to clean up the commits.

bicycle1885:
OK, this is ready to be reviewed.

Ward9250:
That's awesome, I approve. Let me know what you think of #136 and whether it's ok to add into this.

bicycle1885:
I'd like to add a few more commits here.

kescobo:
Just skimming comments, and all of it sounds great. I'm not sure if there's policy on this, but should there be a requirement to update the docs when a major change like this occurs before merging? Seems like it could cause a lot of confusion if the docs lag too far behind alterations to core functionality.

bicycle1885:
Yes, documentation should come with this major change. One obstacle against writing docs in Julia is there is no definitive documentation system. I think we should select one as soon as possible. Judo.jl, which is used to generate [our current document](http://biojulia.github.io/Bio.jl/), is not maintained and I've failed to use it when I tried last time. The most promising one I think is [Lapidary.jl](https://github.com/MichaelHatherly/Lapidary.jl), but it's still under active development. You can view its self-hosted document here: http://michaelhatherly.github.io/Lapidary.jl/latest/. cc: @Ward9250 @Blahah

Ward9250:
Judo.jl is not maintained? I thought this was @dcjones package? To be clear, Julia does have an inbuilt system for documenting API's, types and methods, just not one for generating docs sites or PDFs and that kind thing a'la R's PDF/HTML documentation and vignettes via ROxygen.

Ward9250:
#139

Ward9250:
We have a Lapidary.jl system working now with travis.ci automatic building and hosting on gh-pages through mkdocs :) So that works, it's just a case of updating the documentation.

bicycle1885:
I've added docs about new sequence data structure and extended symbols. We still need to brush up descriptions and styles, but it's not so urgent.

Ward9250:
So long as it is documented, we can brush up documentation and make it consistent later.

Ward9250:
Are there any more commits to come?

bicycle1885:
No. Let's merge it!

PR #138, opened 2016-03-12
add GFF3 parser

bicycle1885:
This adds GFF3 (http://www.sequenceontology.org/gff3.shtml) parser.

- [x] Parse *attributes* (to `Dict`?)
- [x] Percent-encoding support
- [x] "###" directive
- [x] "##FASTA" directive
- [x] Add tests.


Ward9250:
I think passing attributes to a dict is a fairly natural way to do it. It's what I ended up doing for phylogenetic node attributes.

dcjones:
Hey Kenta, is it ok if I take this over and try to get it ready to merge? I'm in urgent need of gff parser for a project, and would like to keep using julia for it.

bicycle1885:
Of course. Are you going to use Ragel?

dcjones:
Yeah, based off the work you've already done. I have it working and passing some tests, I just need to figure out what to do with attributes (especially with those with multiple values) and `##FASTA` sequences.

dcjones:
I suppose this is done.

Attributes are stored in a structure that acts like a dictionary but can be overwritten efficiently.

FASTA in the GFF3 file is accessible by calling `getfasta` on the gff reader after the input has been exhausted, which will return a fasta reader. There's also a `hasfasta` function to check if there is trailing fasta data. This is ugly, but it's an ugly part of the format, and it's the best solution I could think of.

Directives can be optionally saved and when a gff entry is read, you can check for directives occurring since the last entry by calling `directives(reader)`.

Percent encodings are unescaped, which adds a dependency on URIParser.

As an added bonus, I fixed a bug where StringFields always allocate way too much space.

dcjones:
This PR also includes general parsing performance improvements now. Parsing FASTQ is about 2.5x faster, and FASTA 2x faster now.

bdeonovic:
Is this getting merged soon? I have some interest in it.

dcjones:
I've been using it without issue for the past couple weeks. @bicycle1885 do you have any objections to merging?

bicycle1885:
I'll take a look soon, hopefully today.

src/intervals/gff3/reader.jl
```diff
@@ -0,0 +1,70 @@
...
+function GFF3Reader(input::IO, save_directives::Bool=false)
+    return GFF3Reader(BufferedInputStream(input), save_directives)
+end
+
+function IntervalCollection(interval_stream::GFF3Reader)
+    intervals = collect(GFF3Interval, interval_stream)
+    return IntervalCollection{GFF3Metadata}(intervals, true)
+end
+
+function directives(reader::GFF3Reader)
```
bicycle1885: Add docstrings for exported functions.

test/intervals/runtests.jl
```diff
@@ -501,6 +501,45 @@ end
         end
 
     end
+
+    @testset "GFF3 Parsing" begin
```
bicycle1885: Tests for `directives`, `hasfasta` and `getfasta`.

src/intervals/gff3/reader.jl
```diff
@@ -0,0 +1,70 @@
...
+
+function Intervals.metadatatype(::GFF3Reader)
+    return GFF3Metadata
+end
+
+function Base.eltype(::Type{GFF3Reader})
+    return GFF3Interval
+end
+
+function GFF3Reader(input::IO, save_directives::Bool=false)
```
bicycle1885: I think it would be better to define `save_directive` as a keyword argument. Also, need a docstring here.

src/intervals/gff3/gff3.jl
```diff
@@ -0,0 +1,158 @@
...
+        return attrs.data[i]
+    end
+end
+
+function Base.getindex(attrs::GFFAttributes, key_::String)
+    key = StringField(key_.data, 1:length(key_.data))
+    return getindex(attrs, key)
+end
+
+function pushindex!(attrs::GFFAttributes, key::StringField,
```
dcjones: I'm not sure what to call this function. `GFFAttributes` is basically a multi-map, so `pushindex!` was meant to imply something between `setindex!` and `push!`.

dcjones:
Thanks for the review. I addressed everything except for the `pushindex!` thing. I'm not sure what a better name for that function is, but I'll change it if you suggest something.

bicycle1885:
I'll merge this if no further comments. I'm going to squash all commits into one and it will be attributed to @dcjones.

Ward9250:
There still aren't any manpages added to cover the new functionality?

bicycle1885:
It seems to be. I will add follow-up commits soon.

bicycle1885:
I've rebased commits without squashing them because it may remove contributions.

Ward9250:
Should we make that an official policy we all follow when a PR has multiple authors? :)

PR #140, opened 2016-03-18
Refactor ragel

bicycle1885:
This is refactoring of Ragel code. src/Ragel.jl is shorter and `State` has less fields. I also added `@float64_from_anchor!` and `@ascii_from_anchor!` macros, which will be used in #138. 

Now `read!(input, output)` returns `output`, rather `true`/`false`. I think this is a better practice in Julia because `Base.read!` always returns the second argument. Instead, you can check the existence of remaining data by calling `eof` method. So you can write in-place parsing like:

```julia
stream = open("mini.fasta", FASTA)
seq = Seq.FASTASeqRecord()
while !eof(stream)
    read!(stream, seq)
    println(seq)
end
```

(EDIT: now I think the following is a separated issue and reverted the change; never mind)
~~I also changed the method of printing formatted data. On master branch we use `Base.write` method but it is used for writing binary data.~~

```
help?> write
search: write writedlm writecsv writemime readandwrite iswritable mimewritable

  write(stream, x)

  Write the canonical binary representation of a value to the given stream. Returns the
  number of bytes written into the stream.

  You can write multiple values with the same :func:write call. i.e. the following are
  equivalent:

  write(stream, x, y...)
  write(stream, x) + write(stream, y...)
```

~~I think `print` is a suitable method for this purpose when printing human-readable text like FASTA, FASTQ, and BED. So, I changed it.~~


bicycle1885:
Sorry, I've found a case this does not work. I'll fix later.

bicycle1885:
I rebased this onto the master branch, and this is ready to review and merge.

bicycle1885:
I will merge this if nobody blocks me today.

kdmurray91:
Had a cursory glance over it and LGTM.

Ward9250:
:+1: LGTM

bicycle1885:
Thanks!

PR #142, opened 2016-03-22
Add document on pairwise alignment

bicycle1885:
This added document on pairwise alignment and cleaned up some code.


Ward9250:
LGTM :shipit:

PR #143, opened 2016-03-24
add regular expression for BioSequence

### LINKED ISSUE #55: Missing things (before deprecation of BioSeq.jl)
List of things that need to be present on Bio.jl before deprecation of BioSeq.jl

Important:
- [x] **Gap** representation and gapped sequences; replace  BioSeq's aminoacid('-')
- [x] **Alignment** representations: in order to replace BioSeq's Array{AminoAcid,2} 

Optional:
- [ ] Matching functions (search, replace and others):
  - [ ] IUPAC Regex
  - [ ] PROSITE patterns
- [x] 8-bit Bit-Level Coding Scheme for Nucleotides

bicycle1885:
This implements regular expression for bio sequences.

This is still at very early stage, so it is not placed under the Bio module. But some interesting code work:

```julia
julia> include("re.jl")
WARNING: replacing module RE
RE

julia> using Bio.Seq


julia> re = RE.Regex{DNANucleotide}("^A*(C+G{3,})TT")
RE.Regex{Bio.Seq.DNANucleotide}("^A*(C+G{3,})TT")

julia> RE.match(re, dna"AAACGGGTTTT")
Nullable(RegexMatch("AAACGGGTT", 1="CGGG"))

```

This will close #55 since the last missing things are regular expression. At the moment I've implemented only PCRE-like syntax, but adding PROSITE pattern would be straightforward because just implementing a frontend that converts a PROSITE pattern to a `SyntaxTree` object.

Also note that I've not yet measured the performance. So, it is possible that this approach is unacceptably slow.


diegozea:
It looks very awesome @bicycle1885 !

bicycle1885:
Thank you :)

The performance difference versus PCRE used in Julia seems to be not so small. Simple pattern search like `r"T(A[AG]|GA)"` is about 9-10x slower compared to PCRE without considering conversion cost of `BioSequence` => `ASCIIString`. But this result is somewhat reasonable because PCRE does JIT compiling for pattern matching while I'm using a virtual machine written in Julia. There may be a way to hack Julia's JIT compiler to generate native code but I have no idea.

I still need to add some parsimonious operators like "A+?" so that we can get minimum matching.

bicycle1885:
This is replaced with #233.

PR #145, opened 2016-03-26
add constructors of AlignedSequence from BioSequences

bicycle1885:
This adds constructors that create `AlignedSequence` from `BioSequence`s with gap symbols.
I still need to add tests and docs, but it's time to go to bed :sleeping:.

@kescobo Does this meet your demand?

### DIFF
--- a/src/align/anchors.jl
+++ b/src/align/anchors.jl
@@ -255,17 +255,54 @@ immutable AlignedSequence{S}
     aln::Alignment
 end
 
-
-function AlignedSequence{S}(seq::S, aln::Alignment)
-    return AlignedSequence{S}(seq, aln)
-end
-
-
 function AlignedSequence{S}(seq::S, anchors::Vector{AlignmentAnchor},
                             check::Bool=true)
     return AlignedSequence(seq, Alignment(anchors, check))
 end
 
+function AlignedSequence(seq::Sequence, ref::Sequence)
+    return AlignedSequence(seq, 1, ref, 1)
+end
+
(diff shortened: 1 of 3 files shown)


bicycle1885:
I added tests and docs, so it's ready to review and merge.

src/align/anchors.jl
```diff
@@ -255,17 +255,54 @@ immutable AlignedSequence{S}
...
+        throw(ArgumentError("two sequences must be the same length"))
+    end
+    seqpos -= 1
+    refpos -= 1
+    op = OP_START
+    newseq = similar(seq, 0)  # sequence without gap symbols
+    anchors = AlignmentAnchor[]
+    for (x, y) in zip(seq, ref)
+        if x == gap(eltype(seq)) && y == gap(eltype(ref))
+            throw(ArgumentError("two sequences must not have gaps at the same position"))
```
Ward9250: It is perfectly possible for a FASTA file to have two sequences with gaps at the same position - typically if there is a multiple alignment, in which two sequences both have an deletion relative to some other sequence. For now I think this is fine, but may need alteration when I figure out how we will best do MSA, I think I have a solution, just need some spare time to put some code together.

bicycle1885: Exactly. This constructor is intended for pairwise alignment, but, yes, we need to care about MSA so that we can create an `AlignedSequence` for each aligned sequence even when gaps occur at the same positions.

kescobo:
@bicycle1885 Yes, perfect!

bicycle1885:
Great. Merged.

PR #146, opened 2016-03-27
Documentation for Bio.Structure

jgreener64:
Hi all,

This PR adds documentation for `Bio.Structure` in the Lapidary.jl format. It also expands a few in-code docs.

Let me know if it looks okay. Thanks,
Joe

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -1,6 +1,6 @@
 # Project information
 site_name:   Bio.jl
-site_description: Bioinformatics and Computational Biology Infastructure for Julia
+site_description: Bioinformatics and Computational Biology Infrastructure for Julia
 site_author: The BioJulia Organisation
 site_url: https://biojulia.github.io/Bio.jl/
 
@@ -46,6 +46,7 @@ pages:
   - Biological Sequences: man/seq.md
   - Alignments: man/alignments.md
   - Intervals: man/intervals.md
+  - Structure: man/structure.md
 - Reference:
   - Seq: lib/seq.md
 - Contributing: man/contributing.md
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -5,7 +5,8 @@
         "man/install.md",
         "man/seq.md",
         "man/alignments.md",
-        "man/intervals.md"
+        "man/intervals.md",
+        "man/structure.md"
     ]
     Depth = 2
 
--- a/docs/src/man/structure.md
+++ b/docs/src/man/structure.md
@@ -4,6 +4,11 @@
     CurrentModule = Bio.Structure
     DocTestSetup = quote
         using Bio.Structure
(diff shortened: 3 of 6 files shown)


bicycle1885:
Great @jgreener64! Just skimmed it, LGTM.

Ward9250:
The doc builder is failing for some reason in the Travis Build Log.

jgreener64:
Ah, yes.
As part of the doc tests I download a PDB file, which runs the `download` Julia function. This seems to have problems with the autobuild.
Any suggestions here, e.g. could the docs access the `BioFmtSpecimens` in any way? Or I could store one file somewhere in the docs directory?

bicycle1885:
Hmm, I'm not sure but this may be due to a resource leak of opened files in Lapidary.jl (it has been renamed to Documenter.jl) since the error message says:
```
ERROR: LoadError: could not spawn `which curl`: too many open files (EMFILE)
```

jgreener64:
I removed the doc tests to see if it fixed the error. The `LoadError` above goes but there still seem to be some doc build problems.

Ward9250:
Actually it looks ok according to the travis log. The reason the docs look like they build fine, but they skip deployment. This behaviour is intended, as only builds on master deploy docs - Because docs deployment directly manipulates the a folder subtree of the gh-pages Branch, and we only want docs online of content that's actually in master or a release.

jgreener64:
Ah yeah makes sense.
In that case this is ready for review I think.

Ward9250:
Do you think it's possible to include the file that is downloaded in the package or better yet in BioFmtSpecimens? We have tests which access BioFmtSpecimens so if you add the files you want to BioFmtSpecimens we can take a look at getting it to use them and see if it works :).

jgreener64:
Yeah that would be ideal - the file is already in `BioFmtSpecimens` as it happens at `PDB/1EN2.pdb`.

I guess, similar to in the tests, I would run `get_bio_fmt_specimens()` during setup then access a file using `Pkg.dir("Bio", "test", "BioFmtSpecimens", "PDB", "1EN2.pdb")`?

I will attempt this now.

jgreener64:
Have attempted with `BioFmtSpecimens`. Let's see how the build goes.

Ward9250:
Hey, this looks like it worked! :tada:

Ward9250:
I'm happy to merge this. There's currently no struct guidelines for the manual pages. API docstrings are covered in the code guidelines. Maybe we'll make manual guidelines in the near future, but this looks good! :+1:

jgreener64:
:tada: Yeah seemed to work! Great, thanks.

Ward9250:
Now it has merged, we'll be able to see if biojulia.github.io/Bio.jl/latest/ gets updated in about 15 minutes!

Ward9250:
It's up! :tada:

PR #154, opened 2016-04-02
Phylo revival

Ward9250:
A PR to revive and migrate the work being done on the Phylo module from my fork, over the BioJulia repo, and get it merged into master ASAP. At present this does not include work done on plotting phylogenies as there is some issue with Compose I don't have time to look at, and the plotting stuff is only half way complete. So the plan is, correct and edit any code here that currently needs editing, get tests finished, get some manual written and merge.

* [x] Fix code so build and using Bio.Phylo works.
* [ ] Build test suite.
* [ ] Make sure Phylo passes its test suite.
* [ ] Make some manual documentation for upload to web.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -4,3 +4,4 @@ Colors
 IntervalTrees
 Iterators
 Libz
+DataStructures
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -16,13 +16,13 @@ docs_dir: 'build'
 
 # Options
 extra:
-  logo: images/biojl.png 
+  logo: images/biojl.png
   author:
     github: BioJulia
   palette:
     primary: 'indigo'
     accent:  'blue'
-  
+
 extra_css:
   - assets/Documenter.css
 
@@ -47,6 +47,7 @@ pages:
   - Alignments: man/alignments.md
   - Intervals: man/intervals.md
   - Structure: man/structure.md
+  - Phylogenetics: man/phylo.md
 - Reference:
   - Seq: lib/seq.md
 - Contributing: man/contributing.md
--- a/docs/src/man/phylo.md
+++ b/docs/src/man/phylo.md
@@ -1,0 +1,36 @@
+# Bio.Phylo: Phylogenetic Trees
+
+    {meta}
+    CurrentModule = Bio.Phylo
+    DocTestSetup = quote
+        using Bio.Phylo
+    end
+
+The `Bio.Phylo` module provides several data types for handling phylogenetic
+trees and models of ancestry/evolutionary history.
+
+Phylogeneies are inferred models of the inferred evolutionary relationships of a
+set of species, or strains, or populations. The inference is made usually based
+on physical or genetic similarity.
+
+Phylogenies can be rooted, or unrooted, and
+phylogenies can also be vary varied as to the information they contain.
+For example, a cladogram is a phylogeny formed using cladistic methods.
+Cladograms describe the branching pattern of the history of a set of taxa.
+Cladograms do not contain any branch lengths which represent the amount of time
+passed, or amount of genetic change or evolutionary distance. To provide an
+example a cladogram would tell us that we, branch closely with chimps, but will not
+us how much evolutionary distance is between us, or how much time has passed since
+the split.
+
+A Phylogram is like a cladogram, but it _does_ include branch lengths that tell
+you how much genetic change has occurred (typically as an evolutionary distance).
+
+A Chronogram is like a Phylogram, but it's branch lengths indicate evolutionary
+time.
+
+Phylogenies may also have their nodes annotated with any number of pieces of extra
+information and metadata.
+
+
+## PhyNodes: The building blocks of phylogenies.
--- a/src/Bio.jl
+++ b/src/Bio.jl
@@ -11,5 +11,6 @@ include("intervals/Intervals.jl")
 include("align/Align.jl")
 include("structure/Structure.jl")
 include("phylo/Phylo.jl")
+include("tools/Tools.jl")
 
 end # module Bio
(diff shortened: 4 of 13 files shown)


Ward9250:
Interesting tidbit: This branch and PR just triggered the 1000th Travis CI build.

Blahah:
> 1000th Travis CI build.

:tada: :balloon: :birthday:

Ward9250:
It would be cool to get a comment on the functions `cachenodes!` and `sortdescendents` I've implemented them in the most reasonable way I could think of at the time, but if improvements can be made to make them more efficient that would be awesome.

bicycle1885:
This is a general discussion/question about the design of a phylogenetic tree. If you may already have discussed it, I'm sorry for not caching it up. The current `PhyNode` type has several optional fields: `name`, `branchlength`, and `confidence`. If these fields are metadata attached to a tree node, why don't you factor out them as a type parameter?

```julia
type PhyNode{T}
    parent::PhyNode{T}
    children::Vector{PhyNode{T}}
    metadata::T
end
```

This style is already been done in the `Bio.Intervals.Interval{T}` type and seems to be successful.

If `name` is a semi-mandatory field, it would be reasonable to add it to the fields, too. If `name` is intended to be an identifier for a node, users may want to use another type like integer. In such a case, adding the `id` field with one more type parameter would be a sensible choice.

The point I mean is that data structures should be defined as abstractly as possible. When we think a specialization is common and useful, then we can provide it as a type alias.

Ward9250:
I did think about this for a bit, and it occurred to me at the time, to not be the best idea. 

My reasoning was the following:

You might want to annotate a node with all sorts of different types of data, from a colour for plotting to an estimation of a divergence date, to maybe even a sequence, or metadata pulled from a database. 

It's also often the case that you might have different data for different nodes, (in particular you'll have more data for tips and extant things than ones built and inferred by a tree inference program).

This makes what you propose complicated. E.g. Some nodes will be a different type to others in the same tree because of their differing metadata contents. Unless you enforce the metadata field to be instantiated for all nodes but contain no data, in which case what's the difference between that, and what is already done with name, confidence, and branch lengths? Also, changes in type caused by the changing of data would make nodes hard to work with, especially when it would change their {T}, and since their links to one another are also specified with a {T}... 

I also don't think it's the job of the phylogeny to do this. The job of a phylogeny is to describe the structure of an ancestral relationship between OTU's, not to try to be a super flexible container of any kind of data the user might want to pin on it. The minimum amount of information needed to create such a structure was some measure of confidence, an id/name, and a branch length, these are _usually_ output by most tree building programs, except where something like a cladogram without branch-lengths is produced.

Now I've just spent 30 minutes thinking about it again, and I think something like what you propose is possible, but we need to think about it. What does this mean we can and can't do with trees, what happens when the type of one of these parametric nodes in the tree changes, does this create some ripple effect, because of the types of the fields connecting nodes to each other? And so on.

Ward9250:
@bicycle1885 Do you think for the metadata we might be able to take advantage of tuples, the fact that the type of a tuple is a tuple of the types, and generated functions?

Ward9250:
Another alternative of course is to have an abstract node type, and let people define many node types.

bicycle1885:
> I also don't think it's the job of the phylogeny to do this. The job of a phylogeny is to describe the structure of an ancestral relationship between OTU's, not to try to be a super flexible container of any kind of data the user might want to pin on it.

Yes, that's true, absolutely. But trees included in Bio.jl should be as general as possible so that people can build their tools on top of it. I still think there is a simple but extensible way to implement the type for phylogenic tree. An approach may be to separate the tree structure and metadata. For example, we can use LightGraphs.jl to represent a phylogenic tree and branch length and confidence of nodes can be stored in arrays.

Ward9250:
@bicycle1885 I've changed PhyNodes now so they are parameterised and can be made to hold any type as metadata, any kind of support value, and any kind of branch value, so you can create many kinds of nodes, and by extension trees to suit exact purposes. I just need to make the rest of the code compatible, then I can get back to the tests and docs.

I've seen LightGraphs.jl, it looks very nice, and I am quite attracted by it! I've been thinking about it and phylo quite a bit recently. The old (very first!) version of Phylo did have phylogenies represented by a graph, like an edge list. I've seen phylogenies represented in this way in R packages, and I also think in Matlab, because a phylogeny can be thought of as a very specific case of network or graph. But, whilst an adjacency list is efficient for graphs, for some things specific to phylogeny that Evo people want to do, it's not so great, such as pruning trees or doing lots of tree manipulation. Which in the linked node implementation we have is as simple as removing a child ref from an array, and pointing the pruned node's parent field at itself. In an adjacency list this task involves finding all vertices or edges and is O(abs(E)). Using linked nodes is also how it done for the more popular phylogenetics tools out there, and I think also lends itself well to parallel tasks - it's easy to copy different subtrees over to different process.
So we made a decision some time ago to go with this linked node implementation.

That being said, I would love to increase compatibility with LightGraphs and other graph formats, as each is typically very good for one or a few purposes, and julia's type system is ideal for us to create Phylogenies with different internal representation, like a Phylogeny with nodes, and a phylogeny with an adjacency list. One thing I'd like to see brought to LightGraphs which would make it particularly compatible is the ability to store arbitrary types as vertices! But I think this kind of improvement is for a future PR(s), as this PR is just to get finished what is already started, but it's a fantastic suggestion :)

vchuravy:
I think a graph is preferable, but it might be worthwhile to look at https://github.com/Keno/AbstractTrees.jl

Ward9250:
Ok guys, I'm closing this PR, and will open up a new one in a bit of time. I've been thinking about design a lot recently based on everyone's constructive comments. And I've come to a decision, that involves sufficient changes that will mean this becomes a massive PR again. So it's easier to close this, do what I gotta do an open the revamped and more elegant Phylo.

PR #155, opened 2016-04-14
support .2bit format

### LINKED ISSUE #128: Add .2bit file format support
It would be nice to have .2bit file format support:
 * http://genome.cse.ucsc.edu/FAQ/FAQformat.html#format7
 * http://genome-source.cse.ucsc.edu/gitweb/?p=kent.git;a=blob_plain;f=src/inc/twoBit.h;hb=HEAD

bicycle1885:
This will close #128.

### DIFF
--- a/src/seq/Seq.jl
+++ b/src/seq/Seq.jl
@@ -188,6 +188,7 @@ include("seqrecord.jl")
 include("predict.jl")
 include("fasta.jl")
 include("fastq.jl")
+include("twobit.jl")
 
 include("search/exact.jl")
 include("search/approx.jl")
--- a/src/seq/Seq.jl
+++ b/src/seq/Seq.jl
@@ -128,7 +128,8 @@ export
     approxsearchindex,
     approxrsearch,
     approxrsearchindex,
-    ReferenceSequence
+    ReferenceSequence,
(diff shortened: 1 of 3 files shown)


Ward9250:
Looks good! IS 2bit format something that will ragel?

bicycle1885:
No. The .2bit format is a binary format and there is no need to look ahead while parsing.

Ward9250:
:+1: Ok cool, I'm thinking we'll have to add something to the docs for beginners that help them know if something should be done with Ragel or not - so binary formats can be added to the No list of "Will it Ragel?"

bicycle1885:
Binary formats are not necessarily unsuitable for using Ragel. A good criterion is whether the parser needs to do look-ahead parsing. If it is needed, Ragel will be helpful because Ragel does the dirty work of parsing. Human beings are good at look-ahead parsing while machines are not. Since binary formats are designed for machines, parsers can be written without look-ahead process and the speed is often much faster than text formats.

bicycle1885:
Now this is in a good shape and ready to be reviwed.

bicycle1885:
I will merge this one tomorrow if no objections.

PR #157, opened 2016-04-24
Phylo Rework

Ward9250:
Ok, here we have a rework of Phylo to build on top of recent advances in graph packages in Julia - In particular LightGraphs.jl.

The graph types do the job of encoding the structure of the phylogeny, whilst our phylogeny types do the job of encoding the extra information phylogenies require, and make it easy for people to manipulate them.

### DIFF
--- a/src/phylo/Phylo.jl
+++ b/src/phylo/Phylo.jl
@@ -1,0 +1,63 @@
+# The Bio.Phylo module
+# ====================
+#
+# Types and methods for phylogenetic trees.
+#
+# Part of the Bio.Phylo module.
+#
+# This file is a part of BioJulia. License is MIT: https://github.com/BioJulia/Bio.jl/blob/master/LICENSE.md
+
+
+
+module Phylo
+
+using Bio.Tokenize
+
+using LightGraphs
+
+using Bio.Indexers
+
+import Bio.FileFormat
+
+export
+
+    # Types
+    Phylogeny,
+
+    # Phylogeny Methods
+
+    ## Roots & Clades
+    isrooted,
+    isrerootable,
+    root,
+    isroot,
+    clades,
+    isclade,
+
+    ## Leaves / Taxa
+    leaves,
+    isleaf,
+
+    ## Structure manipulation
+    hasparent,
+    parent,
+    haschildren,
+    nchildren,
+    haschild,
+    children,
+    ispreterminal,
+    issemipreterminal,
+
+    ## Misc
+    n_possible_rooted,
+    n_possible_unrooted
+
+
+
+include("phylogeny.jl")
+include("node_basics.jl")
+include("branch_basics.jl")
+include("manipulation.jl")
+
+
+end # module Phylo
--- a/src/phylo/branch_basics.jl
+++ b/src/phylo/branch_basics.jl
@@ -1,0 +1,41 @@
+# phylo/branch_basics.jl
+# ======================
+#
+# Types and methods for phylogenetic trees.
+#
+# Part of the Bio.Phylo module.
+#
+# This file contains methods for querying and setting attributes on branches,
+# as well as adding and removing branches.
+#
+# This file is a part of BioJulia. License is MIT: https://github.com/BioJulia/Bio.jl/blob/master/LICENSE.md
+
+function branchlength{C}(tree::Phylogeny{C, Float64}, edge::Edge)
+    return tree.edgedata[edge]
+end
+
+function branchlength!{C}(tree::Phylogeny{C, Float64}, edge::Edge, value::Float64)
+    tree.edgedata[edge] = value
+    return tree
+end
+
+function rem_branch!(tree::Phylogeny, branch::Edge)
+    rem_edge!(tree.graph, src(branch), dst(branch))
+    pop!(tree.edgedata, branch)
+    return tree
+end
+
+function add_branch!(tree::Phylogeny, branch::Edge, branchlength::Float64)
+    add_edge!(tree.graph, branch)
+    branchlength!(tree, branch, branchlength)
+    return tree
+end
+
+function parent_branch(tree::Phylogeny, vertex::Int)
+    @assert has_parent(tree, vertex) error("Vertex is not connected to a parent.")
+    return in_edges(tree.graph, vertex)[1]
+end
+
+function child_branches(tree::Phylogeny, vertex::Int)
+    return out_edges(tree.graph, vertex)
+end
--- a/src/phylo/manipulation.jl
+++ b/src/phylo/manipulation.jl
@@ -1,0 +1,78 @@
+# phylo/manipulation.jl
+# =====================
+#
+# Types and methods for phylogenetic trees.
+#
+# Part of the Bio.Phylo module.
+#
+# This file contains methods for manipulating the structure of phylogenies.
+#
+# This file is a part of BioJulia. License is MIT: https://github.com/BioJulia/Bio.jl/blob/master/LICENSE.md
+
+
+
+# Internal functions
+
+"""
+    unconnected_clades(tree::Phylogeny)
+
+**Unexported function for internal use.**
+
+Gets the internal clade vertices which are not connected to either parents or
+children.
+"""
+function unconnected_clades(tree::Phylogeny)
+    cv = clades(tree)
+    idx = find(degree(tree.graph, cv) .== 0)
+    return cv[idx]
+end
+
+"""
+    subtree_roots(tree::Phylogeny)
+
+**Unexported function for internal use.**
+
+Gets clade vertices from the phylogeny which are roots of detached subphylogenies.
+Such vertices have children, but no parent, and are not THE root of the
+phylogeny (n + 1). Such subtrees are often created after pruning of trees.
+"""
+function subtree_roots(tree::Phylogeny)
+    cv = clades(tree)
+    idx = find(haschildren(tree, cv) & !hasparent(tree, cv))
+    return cv[idx]
+end
+
+"""
+    disconnect_root(tree::Phylogeny)
+
+**Unexported function for internal use.**
+
+Unconnects the root vertex from all of its children.
+"""
+function disconnect_root!(tree::Phylogeny)
+    r = root(tree)
+    for i in out_neighbours(tree.graph, r)
+        rem_branch!(tree, Edge(r, i))
+    end
+    return tree
+end
+
+"""
+    delete!(tree::Phylogeny, vertex::Int, preserve_bl::Bool = false)
+
+Delete a node from a phylogenetic tree.
+"""
+function delete!(tree::Phylogeny, vertex::Int, preserve_bl::Bool = false)
+    parent = parent(tree, vertex)
+    # Delete the connection to parent but remember the branchlength of the
+    # deleted branch.
+    parentedge = Edge(parent, vertex)
+    lentoparent = preserve_bl ? branchlength(tree, parentedge) : 0.0
+    rem_branch!(tree, parentedge)
+    for branch in child_branches(tree, vertex)
+        newbl = branchlength(tree, branch) + lentoparent
+        add_branch!(tree, Edge(parent, dst(branch)), newbl)
+        rem_branch!(tree, Edge(vertex, dst(branch)))
+    end
+    return tree
+end
--- a/src/phylo/node_basics.jl
+++ b/src/phylo/node_basics.jl
@@ -1,0 +1,270 @@
+# phylo/node_basics.jl
+# ====================
+#
+# Types and methods for phylogenetic trees.
+#
+# Part of the Bio.Phylo module.
+#
+# This file contains the methods for updating or querying characteristics of
+# nodes/clades in a phylogeny. In the Phylo representation these are
+# vertices of a LightGraphs DiGraph, which are numbered higher than the
+# number of taxa + 1.
+#
+# This file is a part of BioJulia. License is MIT: https://github.com/BioJulia/Bio.jl/blob/master/LICENSE.md
+
+"""
+    isroot(tree::Phylogeny, vertex::Int)
+
+Test whether a vertex is the root.
+
+# Examples:
+```julia
+isroot(my_tree, 9)
+```
+"""
+function isroot(tree::Phylogeny, vertex::Int)
+    return vertex == root(tree)
+end
+
+"""
+    isroot{T <: SymString}(tree::Phylogeny, vertex::T)
+
+Test whether a vertex is the root.
+
+# Examples:
+```julia
+isroot(my_tree, :Root)
+isroot(my_tree, :A)
+```
+"""
+function isroot(tree::Phylogeny, vertex::SymString)
+    isroot(tree, tree.vertexindex[name])
+end
+
+"""
+    isclade(tree::Phylogeny, vertex::Int)
+
+Test whether a node is internal clade, i.e. has a parent and one or more children.
+
+# Examples:
+```julia
+isclade(my_tree, 6)
+```
+"""
+function isclade(tree::Phylogeny, vertex::Int)
+    return vertex >= root(tree)
+end
+
+"""
+    isclade(tree::Phylogeny, vertex::SymString)
+
+Test whether a node is internal, i.e. has a parent and one or more children.
+
+# Examples:
+```julia
+isclade(my_tree, :Node7)
+isclade(my_tree, "Homo Sapiens")
+isclade(my_tree, Symbol("Homo Sapiens"))
+```
+"""
+function isclade(tree::Phylogeny, name::SymString)
+    return isclade(tree, tree.vertexindex[name])
+end
+
+"""
+    isleaf(tree::Phylogeny, vertex::Int)
+
+Test whether a vertex is a leaf.
+
+A vertex is a leaf, when it has a parent node, but no children nodes.
+
+# Examples:
+```julia
+isleaf(my_phylogeny, 1)
+```
+"""
+function isleaf(tree::Phylogeny, vertex::Int)
+    return vertex <= tree.ntaxa
+end
+
+"""
+    isleaf(tree::Phylogeny, name::SymString)
+
+Test whether a vertex is a leaf.
+
+A vertex is a leaf, when it has a parent node, but no children nodes.
+
+# Examples:
+```julia
+isleaf(my_phylogeny, :A)
+isleaf(my_phylogeny, "A")
+```
+"""
+function isleaf(tree::Phylogeny, name::SymString)
+    return isleaf(tree, tree.vertexindex[name])
+end
+
+function isclade(tree::Phylogeny, vertex::Int)
+    return vertex > tree.ntaxa
+end
+
+"""
+    hasparent(tree::Phylogeny, vertex::Int)
+
+Check if a vertex in a phylogeny is linked to a parent.
+"""
+function hasparent(tree::Phylogeny, vertex::Int)
+    return indegree(tree.graph, vertex) >= 1
+end
+
+"""
+    hasparent(tree::Phylogeny, name::SymString)
+
+Check if the vertex in a phylogeny is linked to a parent.
+"""
+function hasparent(tree::Phylogeny, name::SymString)
+    return hasparent(tree, tree.vertexindex[name])
+end
+
+"""
+    hasparent(tree::Phylogeny, vertices::AbstractVector{Int})
+
+Check some vertices in a phylogeny and test if each each is linked to a parent.
+"""
+function hasparent(tree::Phylogeny, vertices::AbstractVector{Int})
+    return indegree(tree.graph, vertices) .>= 1
+end
+
+"""
+    parent(x::PhyNode, vertex::Int)
+
+Get the parent of a node. In a phylogeny it is not possible
+for any one node/vertex to have more than one parent.
+
+# Examples:
+```julia
+parent(my_tree, 2)
+```
+"""
+function parent(tree::Phylogeny, vertex::Int)
+    @assert hasparent(tree, vertex) error("Vertex does not have a parent.")
+    return in_neighbours(tree.graph, vertex)[1]
+end
+
+"""
+    hasparent(tree::Phylogeny, child::Int, parent::Int)
+
+Test if a vertex of a phylogeny is the parent of another vertex.
+"""
+function hasparent(tree::Phylogeny, parent::Int, child::Int)
+    return has_edge(tree.graph, parent, child)
+end
+
+"""
+    haschildren(tree::Phylogeny, vertex::Int)
+
+Test whether a vertex in a phylogeny has children.
+"""
+function haschildren(tree::Phylogeny, vertex::Int)
+    return nchildren(tree, vertex) > 0
+end
+
+"""
+    haschildren(tree::Phylogeny, vertices::AbstractVector{Int})
+
+Test vertices in a phylogeny to see if it has children.
+"""
+function haschildren(tree::Phylogeny, vertices::AbstractVector{Int})
+    return nchildren(tree, vertices) .> 0
+end
+
+"""
+    nchildren(tree::Phylogeny, vertex::Int)
+
+Count the number of clades that are direct children of a clade.
+"""
+function nchildren(tree::Phylogeny, vertex::Int)
+    return outdegree(tree.graph, vertex)
+end
+
+"""
+    nchildren(tree::Phylogeny, vertices::AbstractVector{Int})
+
+Count the number of clades that are direct children for a selection of clades.
+"""
+function nchildren(tree::Phylogeny, vertices::AbstractVector{Int})
+    return outdegree(tree.graph, vertices)
+end
+
+"""
+    nchildren(tree::Phylogeny, name::T)
+
+Count the number of clades that are direct children of a clade.
+"""
+function nchildren(tree::Phylogeny, name::SymString)
+    return nchildren(tree, tree.vertexindex[name])
+end
+
+"""
+    haschild(tree::Phylogeny, child::Int, parent::Int)
+
+Test if a vertex of a phylogeny is the child of another vertex.
+"""
+function haschild(tree::Phylogeny, child::Int, parent::Int)
+    return has_edge(tree, parent, child)
+end
+
+"""
+    children(tree::Phylogeny, vertex::Int)
+
+Get the children vertices of a vertex in the phylogeny.
+"""
+function children(tree::Phylogeny, vertex::Int)
+    return copy(out_neighbours(tree.graph, vertex))
+end
+
+"""
+    ispreterminal(tree::Phylogeny)
+
+Test whether a node is preterminal i.e. It's children are all leaves.
+"""
+function ispreterminal(tree::Phylogeny, vertex::Int)
+    if isleaf(tree, vertex)
+        return false
+    end
+    return all([isleaf(tree, i) for i in out_neighbours(tree.graph, vertex)])
+end
+
+"""
+    issemipreterminal(tree::Phylogeny, vertex::Int)
+
+Test whether a node is semi-preterminal i.e. Some of it's children are leaves,
+but not all are.
+"""
+function issemipreterminal(tree::Phylogeny, vertex::Int)
+    if isleaf(tree, vertex)
+        return false
+    end
+    areleaves = [isleaf(tree, i) for i in out_neighbours(tree.graph, vertex)]
+    return any(areleaves) && !all(areleaves)
+end
+
+"""
+    isredundant(tree::Phylogeny, vertex::Int)
+
+Test if a vertex in a phylogeny is redundant or not.
+A vertex is considered redundant is it has a parent and only one child.
+"""
+function isredundant(tree::Phylogeny, vertex::Int)
+    return hasparent(tree, vertex) && nchildren(tree, vertex) == 1
+end
+
+"""
+    isredundant(tree::Phylogeny, vertices::AbstractArray{Int})
+
+Test if vertices in a phylogeny to see if each is redundant or not.
+A vertex is considered redundant is it has a parent and only one child.
+"""
+function isredundant(tree::Phylogeny, vertices::AbstractArray{Int})
+    return hasparent(tree, vertices) & nchildren(tree, vertices) .== 1
+end
--- a/src/phylo/phylogeny.jl
+++ b/src/phylo/phylogeny.jl
@@ -1,0 +1,223 @@
+# phylo/phylogeny.jl
+# ==================
+#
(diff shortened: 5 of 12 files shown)


Ward9250:
Would people be ok with writing tests for the functionality that I currently have present in this PR and then merge it? Basic functions for tips, clades, and the addition and removal of branches is complete. There is of course loads more functionality there can be, including io and larger scale tree rearrangements, but I'm about to hit a hefty period of thesis writing and don't want this to dangle again.

bicycle1885:
I will take a look ASAP, today.

src/phylo/Phylo.jl
```diff
@@ -0,0 +1,73 @@
...
+#
+# Part of the Bio.Phylo module.
+#
+# This file is a part of BioJulia. License is MIT: https://github.com/BioJulia/Bio.jl/blob/master/LICENSE.md
+
+
+
+module Phylo
+
+using Base.Intrinsics
```
bicycle1885: Is this really used?

src/phylo/Phylo.jl
```diff
@@ -0,0 +1,73 @@
...
+
+using Bio.Tokenize
+
+using LightGraphs
+
+using Bio.Indexing
+
+#import DataStructures: Queue, enqueue!, dequeue!, Stack, Queue, Deque
```
bicycle1885: Remove this comment.

src/phylo/Phylo.jl
```diff
@@ -0,0 +1,73 @@
...
+include("phylogeny.jl")
+include("node_basics.jl")
+include("branch_basics.jl")
+include("manipulation.jl")
+
+
+
+#include("treeio.jl")
+#include("iteration.jl")
+#include("annotation.jl")
```
bicycle1885: Remove these comments.

src/phylo/branch_basics.jl
```diff
@@ -0,0 +1,47 @@
...
+# as well as adding and removing branches.
+#
+# This file is a part of BioJulia. License is MIT: https://github.com/BioJulia/Bio.jl/blob/master/LICENSE.md
+
+
+function branchlength(tree::Phylogeny, edge::Edge)
+    error("No method for branchlength for this kind of phylogeny.")
+end
+
+function branchlength{C}(tree::Phylogeny{C, Float64}, edge::Edge, value::Float64)
```
bicycle1885: The forth argument is not used in this function. Is that really needed?

src/phylo/branch_basics.jl
```diff
@@ -0,0 +1,47 @@
...
+    return tree.edgedata[edge]
+end
+
+function branchlength!(tree::Phylogeny, edge::Edge, value::Float64)
+    error("No method for branchlength! for this kind of phylogeny.")
+end
+
+function branchlength!{C}(tree::Phylogeny{C, Float64}, edge::Edge, value::Float64)
+    tree.edgedata[edge] = value
+end
```
bicycle1885: Return `tree`. Generally speaking, I think mutating functions should return the mutated argument like `push!` and `pop!`.

src/phylo/manipulation.jl
```diff
@@ -0,0 +1,79 @@
...
+
+
+"""
+    disconnect_root(tree::Phylogeny)
+
+**Unexported function for internal use.**
+
+Unconnects the root vertex from all of its children.
+"""
+function disconnect_root(tree::Phylogeny)
```
bicycle1885: Maybe the name should be `disconnect_root!`?

src/phylo/node_basics.jl
```diff
@@ -0,0 +1,335 @@
...
+
+Test whether a vertex is the root.
+
+# Examples:
+```julia
+is_root(my_tree, :Root)
+is_root(my_tree, :A)
+```
+"""
+function is_root{T <: SymString}(tree::Phylogeny, vertex::T)
```
bicycle1885: No need to use the type variable. Just `is_root(tree::Phylogeny, vertex::SymString)` would be fine.

src/phylo/node_basics.jl
```diff
@@ -0,0 +1,335 @@
...
+    is_root(tree, tree.vertexindex[name])
```
bicycle1885: Add `return`.

bicycle1885: Also, I can't see `name` variable. Does this really work?

src/phylo/node_basics.jl
```diff
@@ -0,0 +1,335 @@
...
+    return outdegree(tree.graph, vertices)
+end
+
+
+"""
+    n_children(tree::Phylogeny, name::T)
+
+Count the number of clades that are direct children of a clade.
+"""
+function n_children{T <: SymString}(tree::Phylogeny, name::T)
```
bicycle1885: `n_children(tree::Phylogeny, name::SymString)`.

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,230 @@
...
+
+# The Phylogeny Type
+# -------------------
+
+"""
+Phylogeny represents a phylogenetic tree.
+
+Fields:
+
+    - graph: A LightGraphs directional graph.
```
bicycle1885: `LightGraphs` will not make sense to those who don't know the LightGraphs.jl package.

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,230 @@
...
+
+    - ntaxa: A field to keep track of the number of tips in the phylogeny.
+
+    - rooted: A boolean field to keep track of whether the phylogeny is considered
+    rooted or not.
+
+    - rerootable: A boolean field to keep track of whether the phylogeny may be
+    rerooted or not.
+"""
+type Phylogeny{C, B}
```
bicycle1885: I want to know what `C` and `B` type variables mean in comments or docstrings.

src/util/tokenize.jl
```diff
@@ -0,0 +1,67 @@
+# ========================
+# A Tokenizer for Strings
+# ========================
+
+# Ben J. Ward, 2016
```
bicycle1885: Not a good header style.

src/util/tokenize.jl
```diff
@@ -0,0 +1,67 @@
...
+Tokenizer type that is responsible for splitting a string into token according to a
+regex specification. The regex specification is stored in `dict`.
+
+**Fields:**
+
+* `dict`:      A Dictionary of String keys and Regex values.
+* `tokenizer`: A regex that is generated from the specification.
+"""
+type Tokenizer
+    dict::Dict{ASCIIString, Regex}
```
bicycle1885: `ASCIIString` is deprecated in v0.5. Consider to use `Compat.ASCIIString` or `String`.

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,230 @@
...
+    - graph: A LightGraphs directional graph.
+
+    - ntaxa: A field to keep track of the number of tips in the phylogeny.
+
+    - rooted: A boolean field to keep track of whether the phylogeny is considered
+    rooted or not.
+
+    - rerootable: A boolean field to keep track of whether the phylogeny may be
+    rerooted or not.
+"""
```
bicycle1885: I think docstrings do not need to mention the internal fields. These docstrings may send a wrong signal that these fields are accessible to the user. If you intended to use them internally, comments rather docstrings are suitable for developers.

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,230 @@
...
+# Phylogeny Constructors
+# -----------------------
+
+"""
+    Phylogeny{C, B}(tips::Vector{Symbol}, ::Type{C}, ::Type{B})
+
+Construct a completely unresolved phylogeny, with no annotated data.
+
+The phylogeny will store annotation data of type C for tree nodes
+(clades and tips). It will also store data of type B associated with branches.
```
bicycle1885: `C` and `B` should be wrapped by backquotes.

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,230 @@
...
+
+# Examples
+```julia
+Phylogeny([:Human, :Chimp, :Dog, :Cat], Void, Void)
+Phylogeny([:Human, :Chimp, :Dog, :Cat], Float64, Float64)
+```
+"""
+function Phylogeny{C, B}(taxa::Vector{Symbol}, ::Type{C}, ::Type{B})
```
bicycle1885: I think type variables should come first in the arguments (`Phylogeny(::Type{C}, ::Type{B}, taxa::Vector{Symbol}`). The standard library seems to follow the convention.

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,230 @@
...
+    v = copy(tree.vertexdata)
+    d = copy(tree.edgedata)
+    return Phylogeny{C, B}(g, i, v, d, tree.ntaxa, tree.rooted, tree.rerootable)
+end
+
+
+# Miscellaneous Functions
+# ------------------------
+
+function n_possible_rooted(n::Int)
```
bicycle1885: No docstrings?

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,230 @@
...
+    @assert n >= 2 "n must be greater or equal to 2."
+    num = factorial(2 * n - 3)
```
bicycle1885: This will overflow with `n = 12`. Is that cover common use cases?

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,230 @@
...
+is_rerootable(my_phylogeny)
+```
+"""
+function is_rerootable(x::Phylogeny)
+    return x.rerootable
+end
+
+
+"Get the vertex of the tree which represents the root of the tree."
+root_vertex(tree::Phylogeny) = tree.ntaxa + 1
```
bicycle1885: This will return a vertex even when `tree` is not rooted (i.e. `tree.rooted == false`). Is that intended?

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,230 @@
...
+
+
+"Get the range of vertices which represent the internal clades of the tree."
+clade_vertices(tree::Phylogeny) = (root_vertex(tree) + 1):nv(tree.graph)
```
bicycle1885: This kind of internal design of `Phylogeny` is the most important thing to those who are not familiar with the implementation. I think this should be commented just before the definition of the `Phylogeny` type like:

    # Vertex ranges:
    # * 1:tree.ntaxa => leaf vertices
    # * tree.ntaxa + 1 => root vertex
    # * tree.ntaxa + 1:<#nodes> => clade vertices

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,230 @@
...
+
+
+"""
+    no_children(tree::Phylogeny)
+
+Test each vertices in a phylogeny to see if it lacks children.
+"""
+function no_children(tree::Phylogeny)
+    return n_children(tree) .== 0
+end
```
bicycle1885: `has_parent`, `no_parent`, `has_children`, and `no_children` for the `Phylogeny` type are not tested.

test/phylo/runtests.jl
```diff
@@ -1,2 +1,56 @@
...
+@testset "Phylogenies" begin
+
+    @testset "Creation" begin
+        # Create an unresolved star phylogeny.
+        tips = [:A, :B, :C]
+        tree = Phylogeny(tips)
+
+        @testset "Roots" begin
+            @test is_rooted(tree) == false
+            @test is_rerootable(tree) == true
```
bicycle1885: `@test is_rerootable(tree)` is enough.

test/phylo/runtests.jl
```diff
@@ -1,2 +1,56 @@
...
+        @testset "Leaves" begin
+            @test leaf_vertices(tree) == 1:3
+            @test is_leaf(tree, 1) == true
+            @test is_leaf(tree, 2) == true
+            @test is_leaf(tree, 3) == true
+        end
+
+        @testset "Clades" begin
+            @test is_clade(tree, 4) == true
+            @test internal_vertices(tree) == 4:5
```
bicycle1885: I cannot find the `internal_vertices` method. Where does it come from?

src/phylo/branch_basics.jl
```diff
@@ -0,0 +1,47 @@
...
+# Part of the Bio.Phylo module.
+#
+# This file contains methods for querying and setting attributes on branches,
+# as well as adding and removing branches.
+#
+# This file is a part of BioJulia. License is MIT: https://github.com/BioJulia/Bio.jl/blob/master/LICENSE.md
+
+
+function branchlength(tree::Phylogeny, edge::Edge)
+    error("No method for branchlength for this kind of phylogeny.")
```
bicycle1885: Julia will automatically throw `MethodError` if there is no matching method. Why this definition is needed?

src/phylo/node_basics.jl
```diff
@@ -0,0 +1,335 @@
...
+    is_root(tree::Phylogeny, vertex::Int)
+
+Test whether a vertex is the root.
+
+# Examples:
+```julia
+is_root(my_tree, 9)
+```
+"""
+function is_root(tree::Phylogeny, vertex::Int)
```
bicycle1885: Using underscores is discouraged in Julia (http://julia.readthedocs.io/en/latest/manual/variables/#stylistic-conventions).  I think `isroot` is readable enough.

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,230 @@
...
+"Get the vertex of the tree which represents the root of the tree."
+root_vertex(tree::Phylogeny) = tree.ntaxa + 1
+
+
+"Get the range of vertices which represent the internal clades of the tree."
+clade_vertices(tree::Phylogeny) = (root_vertex(tree) + 1):nv(tree.graph)
+
+
+"Get the range of vertices which represent the leaves of the tree."
+leaf_vertices(tree::Phylogeny) = 1:tree.ntaxa
```
bicycle1885: `root`, `clades`, and `leaves` would be more concise names.

src/phylo/phylogeny.jl
Ward9250: n is often over twelve, trees can have quite a large number of taxa now.

src/phylo/phylogeny.jl
Ward9250: It is intended as it is common in all phylogenetic software that the trees have a "root" even if the tree is considered an unrooted tree, typically for practical reasons. A rooted and unrooted tree will always have a vertex that is reserved for a root (ntaxa + 1). So this function can be seen as getting the vertex that is reserved for the root of the tree, regardless of whether or not the tree is rooted and the vertex is connected. Similar to how clade_vertices returns the vertices that the tree has to use as clade nodes, regardless of whether or not they are used or connected up to the tree structure.

src/phylo/phylogeny.jl
Ward9250: That's a good idea :) I will do this.

src/phylo/Phylo.jl
Ward9250: Probably not, it was taken from another modules example. I'll remove this.

src/phylo/phylogeny.jl
Ward9250: Added a clarification to the doctoring that the graph field is a directional graph. And for phylogenies we use the directional graph as implemented in LightGraphs as the DiGraph type. This should be clearer to readers that this is a type imported from a different package.

test/phylo/runtests.jl
```diff
@@ -1,2 +1,70 @@
...
+                nchild = i == 4 ? 3 : 0
+                hasc = i == 4 ? true : false
+                noc = i == 4 ? false : true
+                @test hasparent(tree, i) == hasp
+                @test noparent(tree, i) == nop
+                @test nchildren(tree, i) == nchild
+                @test haschildren(tree, i) == hasc
+                @test nochildren(tree, i) == noc
+            end
+            @test hasparent(tree, [1:5]) == [true, true, true, false, false]
```
bicycle1885: `[1:4]` is deprecated in Julia 0.4, and removed in 0.5.

src/phylo/node_basics.jl
```diff
@@ -0,0 +1,335 @@
...
+end
+
+
+
+"""
+    noparent(tree::Phylogeny, vertex::Int)
+
+Check if a vertex in a phylogeny is not linked to a parent.
+"""
+function noparent(tree::Phylogeny, vertex::Int)
```
bicycle1885: Since `hasparent` is defined, I think this function is a little bit redundant because we can write it as `!hasparent(tree, vertex)`.

src/phylo/node_basics.jl
```diff
@@ -0,0 +1,335 @@
...
+end
+
+
+"""
+    noparent(tree::Phylogeny, vertices::AbstractVector{Int})
+
+Check some vertices in a phylogeny and test if each each is not linked to a
+parent.
+"""
+function noparent(tree::Phylogeny, vertices::AbstractVector{Int})
```
bicycle1885: `!hasparent`.

src/phylo/node_basics.jl
```diff
@@ -0,0 +1,335 @@
...
+    return nchildren(tree, vertices) .> 0
+end
+
+
+"""
+    nochildren(tree::Phylogeny, vertex::Int)
+
+Test a vertex in a phylogeny to see if it lacks children.
+"""
+function nochildren(tree::Phylogeny, vertex::Int)
```
bicycle1885: `!haschildren`

src/phylo/node_basics.jl
```diff
@@ -0,0 +1,335 @@
...
+    return nchildren(tree, vertex) == 0
+end
+
+
+"""
+    nochildren(tree::Phylogeny, vertices::AbstractVector{Int})
+
+Test vertices in a phylogeny to see if they lack children.
+"""
+function nochildren(tree::Phylogeny, vertices::AbstractVector{Int})
```
bicycle1885: `!haschindren`

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,260 @@
...
+Returns the number of possible rooted tree topoligies for n taxa.
+
+**Warning:** This method currently overflows when n > 12.
+"""
+function n_possible_rooted(n::Int)
+    @assert n >= 2 "n must be greater or equal to 2."
+    num = factorial(2 * n - 3)
+    den = (2^(n - 2)) * factorial(n - 2)
+    return Int(num / den)
+end
```
bicycle1885: You can compute `(2n-3)!!` as `doublefactorial(2n-3)` using the [Combinatorics.jl](https://github.com/JuliaLang/Combinatorics.jl) package.

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,260 @@
...
+
+"""
+    n_possible_rooted(n::Int)
+
+Returns the number of possible rooted tree topoligies for n taxa.
+
+**Warning:** This method currently overflows when n > 12.
+"""
+function n_possible_rooted(n::Int)
+    @assert n >= 2 "n must be greater or equal to 2."
```
bicycle1885: `DomainError` would be the most suitable exception here.

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,260 @@
...
+
+"""
+    n_possible_unrooted(n::Int)
+
+Returns the number of possible rooted tree topoligies for n taxa.
+
+**Warning:** This method currently overflows when n > 13.
+"""
+function n_possible_unrooted(n::Int)
+    @assert n >= 3 "n must be greater of equal to 3."
```
bicycle1885: `DomainError`

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,260 @@
...
+end
+
+
+# Miscellaneous Functions
+# ------------------------
+
+"""
+    n_possible_rooted(n::Int)
+
+Returns the number of possible rooted tree topoligies for n taxa.
```
bicycle1885: Typo: topologies

bicycle1885: Also, backquote `n`.

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,260 @@
...
+    num = factorial(2 * n - 3)
+    den = (2^(n - 2)) * factorial(n - 2)
+    return Int(num / den)
+end
+
+
+"""
+    n_possible_unrooted(n::Int)
+
+Returns the number of possible rooted tree topoligies for n taxa.
```
bicycle1885: Fix a typo: "topologies", and backquote `n`.

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,260 @@
...
+    return indegree(tree.graph) .== 1
+end
+
+
+"""
+    noparent(tree::Phylogeny)
+
+Check every vertex in a phylogeny and test if each is not linked to a parent.
+"""
+function noparent(tree::Phylogeny)
```
bicycle1885: `!hasparent`.

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,260 @@
...
+    return nchildren(tree) .> 0
+end
+
+
+"""
+    nochildren(tree::Phylogeny)
+
+Test each vertices in a phylogeny to see if it lacks children.
+"""
+function nochildren(tree::Phylogeny)
```
bicycle1885: `!haschildren`.

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,260 @@
...
+
+Construct a completely unresolved phylogeny, with no annotated data.
+
+The phylogeny will store annotation data of type `C` for tree nodes
+(clades and tips). It will also store data of type `B` associated with branches.
+
+# Examples
+```julia
+Phylogeny([:Human, :Chimp, :Dog, :Cat], Void, Void)
+Phylogeny([:Human, :Chimp, :Dog, :Cat], Float64, Float64)
```
bicycle1885: Update docstrings to reflect the change of the argument order.

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,260 @@
...
+```
+"""
+function Phylogeny(taxa::Vector{Symbol})
+    return Phylogeny(Float64, Float64, taxa)
+end
+
+
+"""
+    Phylogeny{T <: AbstractString}(taxa::Vector{T})
+Construct a completely unresolved phylogeny, with no annotated data.
```
bicycle1885: Insert a blank line before this description.

src/phylo/phylogeny.jl
```diff
@@ -0,0 +1,260 @@
...
+
+This constructor assumes you want a phylogeny where the only data associated
+with it is branch lengths (Float64) for the edges, and confidence values
```
bicycle1885: Backquote `Float64`.

src/phylo/node_basics.jl
```diff
@@ -0,0 +1,335 @@
...
+
+
+
+"""
+    is_parent_of(tree::Phylogeny, child::Int, parent::Int)
+
+Test if a vertex of a phylogeny is the parent of another vertex.
+"""
+function is_parent_of(tree::Phylogeny, parent::Int, child::Int)
+    return has_edge(tree, parent, child)
```
bicycle1885: Does this really work? I think you mean `has_edge(tree.graph, parent, child)`.

src/phylo/node_basics.jl
```diff
@@ -0,0 +1,335 @@
...
+end
+
+
+
+"""
+    is_parent_of(tree::Phylogeny, child::Int, parent::Int)
+
+Test if a vertex of a phylogeny is the parent of another vertex.
+"""
+function is_parent_of(tree::Phylogeny, parent::Int, child::Int)
```
bicycle1885: Overloading the `hasparent` function may be better (`hasparent(tree, child, parent)`).

src/phylo/node_basics.jl
```diff
@@ -0,0 +1,335 @@
...
+
+Test whether a vertex is the root.
+
+# Examples:
+```julia
+isroot(my_tree, :Root)
+isroot(my_tree, :A)
+```
+"""
+function isroot{T <: SymString}(tree::Phylogeny, vertex::T)
```
bicycle1885: `isroot(tree::Phylogeny, vertex::SymString)`.

src/phylo/node_basics.jl
```diff
@@ -0,0 +1,335 @@
...
+    return outdegree(tree.graph, vertices)
+end
+
+
+"""
+    nchildren(tree::Phylogeny, name::T)
+
+Count the number of clades that are direct children of a clade.
+"""
+function nchildren{T <: SymString}(tree::Phylogeny, name::T)
```
bicycle1885: `nchildren(tree::Phylogeny, name::SymString)`.

src/phylo/manipulation.jl
```diff
@@ -0,0 +1,80 @@
...
+    # deleted branch.
+    parentedge = Edge(parent, vertex)
+    lentoparent = preserve_bl ? branchlength(tree, parentedge) : 0.0
+    rem_branch!(tree, parentedge)
+    for branch in child_branches(tree, vertex)
+        newbl = branchlength(tree, branch) + lentoparent
+        add_branch!(tree, Edge(parent, dst(branch)), newbl)
+        rem_branch!(tree, Edge(vertex, dst(branch)))
+    end
+end
```
bicycle1885: Return the mutated argument: `tree`.

src/phylo/node_basics.jl
```diff
@@ -0,0 +1,335 @@
...
+end
+
+
+
+function isclade(tree::Phylogeny, vertex::Int)
+    return vertex > tree.ntaxa
+end
+
+
+
```
bicycle1885: Too many blank lines. I think just one blank line is enough.

src/phylo/node_basics.jl
Ward9250: That would also make it return true if the node had more than one parent - which should never happen, but coparent would return true instead of false if it ever did happen by mistake.

bicycle1885: If so, why didn't you define `hasparent(tree, vertex)` as `indegree(tree.graph, vertex) >= 1` rather `indegree(tree.graph, vertex) == 1`? I think `noparent` should be the negation of `hasparent`. That is, `hasparent` and `noparent` should not return the same value (true/false) at the same time. If a vertex accidentally has two parents, `hasparent` should be ture and `noparent` should be false.

src/phylo/phylogeny.jl
Ward9250: Ah that's really cool! I was doing a fair amount of reading last night about how to program numerical operations that are far too big, even bigger than bigger integer types.

src/util/tokenize.jl
Ward9250: Will using String work in 0.4 though? I understand that String used to be AbstractString or something like that.

```
julia> String
WARNING: Base.String is deprecated, use AbstractString instead.
  likely near no file:0
AbstractString
```

bicycle1885: Sorry, I mean `Compat.String`, which is defined in Compat.jl.

src/phylo/node_basics.jl
Ward9250: I agree, has parent by definition should probably only tell you if a vertex has a parent, if you have two parents you do technically still have a parent. hasparent should be defined as you suggest, and maybe some other method to catch cases where a vertex specifically has more than one parent.

bicycle1885: Yes, so I believe there is no chance to use `noparent` since we already has `haspanret`.

src/util/tokenize.jl
Ward9250: Awesome, I've made them Compat.String

Ward9250:
Tests pass on 0.4 and I think 0.5 fails for an unrelated issue. All review suggestions have been acted on. I'd like to merge tonight if nobody blocks me.

src/phylo/node_basics.jl
```diff
@@ -0,0 +1,293 @@
...
+function hasparent(tree::Phylogeny, vertices::AbstractVector{Int})
+    return indegree(tree.graph, vertices) .>= 1
+end
+
+"""
+    noparent(tree::Phylogeny, vertex::Int)
+
+Check if a vertex in a phylogeny is not linked to a parent.
+"""
+noparent(tree::Phylogeny, vertex::Int) = !hasparent(tree, vertex)
```
bicycle1885: I still think this function is redundant. It is just an alias of `!hasparent` but doesn't make code simpler at all. Why do you think this is needed?

src/phylo/branch_basics.jl
```diff
@@ -0,0 +1,41 @@
...
+# Types and methods for phylogenetic trees.
+#
+# Part of the Bio.Phylo module.
+#
+# This file contains methods for querying and setting attributes on branches,
+# as well as adding and removing branches.
+#
+# This file is a part of BioJulia. License is MIT: https://github.com/BioJulia/Bio.jl/blob/master/LICENSE.md
+
+function branchlength{C}(tree::Phylogeny{C, Float64}, edge::Edge, value::Float64)
```
bicycle1885: The third argument (`value`) is not used in the definition.

src/phylo/Phylo.jl
```diff
@@ -0,0 +1,66 @@
...
+
+    ## Leaves / Taxa
+    leaves,
+    isleaf,
+
+    ## Structure manipulation
+    hasparent,
+    noparent,
+    parent,
+    is_parent_of,
```
bicycle1885: This function no longer exists.

src/phylo/Phylo.jl
```diff
@@ -0,0 +1,66 @@
...
+    haschildren,
+    nochildren,
+    nchildren,
+    is_child_of,
```
bicycle1885: This function no longer exists.

src/phylo/branch_basics.jl
Ward9250: Fixed

src/phylo/node_basics.jl
Ward9250: This is now fixed with next commit.

src/phylo/Phylo.jl
Ward9250: Module exports fixed in next commit.

bicycle1885:
Okay, :shipit: I think it would be better to squash the long series of commits so that people can easily figure out what happened there.

Ward9250:
Thanks @bicycle1885 :D Yes I will make sure to squash commits before I merge.

PR #158, opened 2016-04-24
add documentation for blast module

kescobo:
Documentation for the blast command line wrapper. Please let me know if I should add anything else.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -6,7 +6,7 @@
         "man/seq.md",
         "man/alignments.md",
         "man/intervals.md",
-        "man/structure.md"
+        "man/structure.md",
+        "man/tools.md"
     ]
     Depth = 2
-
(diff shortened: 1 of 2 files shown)


Ward9250:
This LGTM 👍 the modules make sense. Looks like some nice simple examples! I'm happy to merge this. On merge it should update the web - although the build can take some minutes.

jgreener64:
Looks good! Am I missing something, or have the online docs not been updated?
Does `mkdocs.yml` need to be updated?

Ward9250:
The docs build system won't kick in unless tests pass on julia v0.4. Because there's no point in building documentation for things that don't work as expected and adding a whole ton of commits to the gh-pages while people are tinkering or developing. That's the current settings anyway. The idea being when we make tagged releases, they will have a successful build and passing test suite, and so the website will have a "latest" docs which will be docs of the last time the in development stuff passed it's tests, then docs for v0.0.1, v0.5 and so on. #159 is the current sticking point here that's stopping the docs build system from generating the pages for blast.

jgreener64:
Ah okay that makes sense :+1:

Issue #160, opened 2016-04-25
Genotype Count Matrices for Population Genetics


BenJWard:
In population genetics we desire matrix-like types which allow us to store genotypic data in a convenient format which permits efficient computation of population genetics summaries or other statistics. E.g. The frequency of an allele in a population and so on. Here we discuss implementation and design.
# Basic Design

The basic design of the matrix will be a matrix, where rows represent individuals and columns represent the alleles, such that one gene might have 3 columns if it has 3 distinct alleles of that gene. 

Furthermore, individuals may be grouped into populations. So you should be able to do things like:

`my_matrix[:, :GeneA]` or `my_matrix[Ind(:Mouse), :]` or `my_matrix[Pop(:Norwich), :]`

To select entire genes, individuals, or populations.

To this end, indexing data structures which create Integer indexes based on names (symbols) have been implemented, and will be merged into Bio in Bio.Indexing as a separate PR.
# TODO
- [x] Merge Indexing datastructures.
- [ ] Create basic type for the count matrices.
- [ ] IO for the most common formats.

Ward9250:
# Basic Design

The basic design of the matrix will be a matrix, where rows represent individuals and columns represent the alleles, such that one gene might have 3 columns if it has 3 distinct alleles of that gene. 

Furthermore, individuals may be grouped into populations. So you should be able to do things like:

`my_matrix[:, :GeneA]` or `my_matrix[Ind(:Mouse), :]` or `my_matrix[Pop(:Norwich), :]`

To select entire genes, individuals, or populations.

To this end, indexing data structures which create Integer indexes based on names (symbols) have been implemented, and will be merged into Bio in Bio.Indexing as a separate PR.

Ward9250:
#161 has been created. Comments on it are welcome. There is some functionality I would like to add to the indexes (mostly for phylo) to do with pushing new label-value associations. or removing specific ones, or reassigning them (removing tree nodes, adding tree nodes, and pruning and grafting trees and such! - but that's a separate issue to this).

Ward9250:
#161 is now merged. Indexers can be used with the types planned here and in Phylo.

PR #161, opened 2016-04-25
Indexes

Ward9250:
Datatypes in planned genotype count arrays #160 and the phylogenetic module rework #157 require data structures to index names or labels to integers or a set of integers. This PR provides them.


src/util/indexing.jl
```diff
@@ -0,0 +1,163 @@
...
+function SingleIndex(names::Vector{Symbol})
+    u = make_unique(names)
+    lookup = Dict{Symbol, Int}(zip(u, 1:length(u)))
+    SingleIndex(lookup, u)
+end
+SingleIndex{S <: AbstractString}(names::Vector{S}) = SingleIndex(convert(Vector{Symbol}, names))
+SingleIndex() = SingleIndex(Dict{Symbol, Int}(), Symbol[])
+
+
+type GroupIndex{T <: rangeOrVect} <: AbstractIndex
```
bicycle1885: Why not `T<:AbstractVector{Int}` rather `T<:rangeOrVect`. `UnitRange{Int}` actually is a subtype of `AbstractVector{Int}`.

Ward9250: I didn't know UnitRange was an abstract vector, although it does now make sense when I think LightGraphs makes use of UnitRange for it's vertex storage! Thanks @bicycle1885!

src/util/indexing.jl
```diff
@@ -0,0 +1,150 @@
...
+                    AbstractVector{UInt128}}
+
+# The indexer Type
+type Indexer{T <: intsAndVecs}
+    lookup::Dict{Symbol, T}
+    names::Vector{Symbol}
+end
+
+# Conveinience constructors for creating Indexer that associate one name,
+# with in integer value.
```
kdmurray91: Pedantry: `s/in/an/`

src/util/indexing.jl
```diff
@@ -0,0 +1,150 @@
...
+end
+function Indexer{T <: intsAndVecs}(::Type{T})
+    Indexer(Dict{Symbol, T}(), Symbol[])
+end
+
+function make_unique(names::Vector{Symbol})
+    seen = Set{Symbol}()
+    names = copy(names)
+    dups = Int[]
+    for i in 1:length(names)
```
kdmurray91: Pedantry: I find `for (i, name) in enumerate(names) ... end` more concise. (Damn pythonistas :smile:)

Ward9250: Good catch, although I think one second thoughts, the i can be done away with in both loops in the make_unique function and just name be used instead.

Ward9250:
Tests fail because of the BED thing, if people say :shipit: then I'll make sure to rebase to get rid of that error before merging.

src/util/indexing.jl
```diff
@@ -0,0 +1,249 @@
...
+names[1] -> 1, names[2] -> 2, ..., names[length(names)] -> length(names).
+"""
+function Indexer{S <: AbstractString, T <: Integer}(names::Vector{S}, ::Type{T})
+    return Indexer(convert(Vector{Symbol}, names), T)
+end
+
+
+"""
+Construct an Indexer that maps a Vector of Symbols (names), to a set of
+values of type Int, where Int is the default integer encoding on you machine
```
kdmurray91: s/you/your/

src/util/indexing.jl
```diff
@@ -0,0 +1,249 @@
...
+"""
+Rename a indexer according to some function, f, that accepts one argument.
+
+Each name in the Indexer x, will be input to, f, and the output will be
+what that name will be renamed to.
+
+Therefore the function, f, will need to return Symbols or Strings.
+"""
+rename!(x::Indexer, f::Function) = rename!(x, [(x,f(x)) for x in x.names])
+rename!(f::Function, x::Indexer) = rename!(x, f)
```
kdmurray91: Having both these `rename!`s (L207-208)  seems a little funny to me. I'd prefer to pick whichever is more Julian.

src/util/indexing.jl
```diff
@@ -0,0 +1,249 @@
...
+rename(x::Indexer, args...) = rename!(copy(x), args...)
+rename(f::Function, x::Indexer) = rename(copy(x), f)
```
kdmurray91: This also feels a little weird in light of L207. I'm not sure which argument order it should be, but i think it should be *one* of them :)

src/util/indexing.jl
```diff
@@ -0,0 +1,249 @@
...
+end
+
+
+"Completely replace the names in an Indexer with a new Vector of Strings."
+function names!{S <: AbstractString}(x::Indexer, names::Vector{S})
+    names!(x, convert(Vector{Symbol}, names))
+end
+
+
+function rename!(x::Indexer, names)
```
kdmurray91: Is it an issue that all the `rename` functions only work with `Symbol`? I'm going to assume not, but other operations are defined for both

test/runtests.jl
```diff
@@ -23,6 +22,10 @@ end
 
 include("TestFunctions.jl")
 
+for target in readdir("util")
+    include("util/$target")
+end
+
```
kdmurray91: Hmmn. Would be good if these could be added to the target list as `util/XXXX` so they behave just like seq, intervals etc on the command line. Would negate this for loop too.

kdmurray91: To clarify, I mean moving the test sources to e.g. `test/util/indexing/runtests.jl`. Though you could just as well do away with the `util` subfolder, if you wanted (I'm +0 on that).

test/util/indexing.jl
```diff
@@ -0,0 +1,181 @@
...
+        @test names(g_index_three) == symbolnames
+        @test names(empty_index) == Symbol[]
+        @test isequal(copy(int_index_one), int_index_one) && !is(copy(int_index_one), int_index_one)
+        @test isequal(copy(int_index_two), int_index_two) && !is(copy(int_index_two), int_index_two)
+        @test isequal(copy(int_index_three), int_index_three) && !is(copy(int_index_three), int_index_three)
+        @test isequal(copy(g_index_one), g_index_one) && !is(copy(g_index_one), g_index_one)
+        @test isequal(copy(g_index_two), g_index_two) && !is(copy(g_index_two), g_index_two)
+        @test isequal(copy(g_index_three), g_index_three) && !is(copy(g_index_three), g_index_three)
+        @test isequal(copy(empty_index), empty_index) && !is(copy(empty_index), empty_index)
+        for n in symbolnames
```
kdmurray91: Pedantry: This seems like it could use a refactor

```julia
for indexset in (symbolnames, textnames, 1:5)
     for n in indexset
           @test all_the_things(n)
     end
end
```

test/util/indexing.jl
```diff
@@ -0,0 +1,181 @@
...
+        @test typeof(int_index_one) == Indexer{Int64}
+        @test typeof(int_index_two) == Indexer{UInt64}
+        @test typeof(int_index_three) == Indexer{UInt8}
+        @test typeof(g_index_one) == Indexer{UnitRange{Int64}}
+        @test typeof(g_index_two) == Indexer{UnitRange{Int64}}
+        @test typeof(g_index_three) == Indexer{UnitRange{UInt8}}
+        @test typeof(empty_index) == Indexer{UInt64}
+    end
+
+    @testset "Basic Operators" begin
```
kdmurray91: Most of these test sets could be refactored to use the "for loop" mode of `@testset` (examples in [docs](http://docs.julialang.org/en/latest/stdlib/test/#working-with-test-sets))

```julia
@testset "Foo the Bar with Baz" for idxer in (int_index_one, int_index_two, ...) do
     @test the_thing_with(idxer) == whatever
end
```

kdmurray91:
:+1: 

There's nothing but pedantry in the above comments @Ward9250, so ship away when you're happy!

src/util/indexing.jl
Ward9250: It's basically if people put the two arguments the other way round. So the history of this type is I took it verbatim from - I think - DataFrames.jl, and modified it a bit, and then recently modified it a lot, and so this second rename! is a revenant of when it was more like DataFrames.jl. I don't know which is more julian, but since the second function is the odd one out of all the code, we can get rid of that one.

test/runtests.jl
Ward9250: I agree that perhaps some re-org needs to be done. I think I did this as a quick and dirty workaround, so I could see the test outputs before the BigBED bug came along threw errors and stopped the testing.

test/util/indexing.jl
Ward9250: I was not aware you could so that with @testset! That's freaking sweet! :)

src/util/indexing.jl
kdmurray91: OK, in that case L210 should have the args swapped??

src/util/indexing.jl
Ward9250: Fixed!

src/util/indexing.jl
Ward9250: Second rename! removed.

src/util/indexing.jl
Ward9250: Arguments swapped to match the corresponding rename!

src/util/indexing.jl
Ward9250: methods of rename! have been added for `S <: AbstractString`

Ward9250:
Just found a horrible flaw.

Constructors that accept vectors of symbols / strings will create indexers containing references to the array of names (arrays are mutable composite types and so in julia are passed by reference).

Meaning the indexer references that array of names, but so could other stuff since presumably references to the array existed in other scopes in order to be passed to the constructor. Meaning the names of the indexer could be altered without the Indexer being the one doing the altering, resulting in incorrect and undefined behaviour of the indexer.

To solve this, constructors will take copies of arrays so as each indexer points to its own array of names, nothing else can / should unless someone does something stupid with the dot operator, which we could alter the behaviour of anyway and prevent if features I've seen suggested in julia issues work out.

Ward9250:
Ah false alarm, the make_unique function copies the names array. But this is a good test to include to make sure at no point does this change.

src/util/indexing.jl
```diff
@@ -0,0 +1,263 @@
...
+            k += 1
+        end
+    end
+    return names
+end
+
+
+Base.length(x::Indexer) = length(x.names)
+Base.names(x::Indexer) = copy(x.names)
+_names(x::Indexer) = x.names
```
bicycle1885: What purpose is this function intended for?

src/util/indexing.jl
```diff
@@ -0,0 +1,263 @@
...
+Base.copy(x::Indexer) = Indexer(copy(x.lookup), copy(x.names))
+Base.deepcopy(x::Indexer) = copy(x) # all eltypes immutable
```
bicycle1885: Is this needed?

src/util/indexing.jl
```diff
@@ -0,0 +1,263 @@
...
+Base.isequal(x::Indexer, y::Indexer) = isequal(x.lookup, y.lookup) && isequal(x.names, y.names)
```
bicycle1885: Basically, you don't need to defined the `Base.isequal` method if you have `Base.(:(==))` because `Base.isequal` falls back to `Base.(:(==))`.

src/util/indexing.jl
```diff
@@ -0,0 +1,263 @@
...
+Base.(:(==))(x::Indexer, y::Indexer) = isequal(x, y)
+Base.haskey(x::Indexer, key::Symbol) = haskey(x.lookup, key)
+Base.haskey{S <: AbstractString}(x::Indexer, key::S) = haskey(x, convert(Symbol, key))
```
bicycle1885: `Base.haskey(x::Indexer, key::AbstractString) = ...` will be fine here.

src/util/indexing.jl
```diff
@@ -0,0 +1,263 @@
...
+Base.haskey(x::Indexer, key::Real) = 1 <= key <= length(x.names)
```
bicycle1885: I think `key::Integer` is a better signature because `key::Real` suggests `key = 3.1415` is a valid argument.

src/util/indexing.jl
```diff
@@ -0,0 +1,263 @@
...
+
+"""
+Rename a indexer according to some function, f, that accepts one argument.
+
+Each name in the Indexer x, will be input to, f, and the output will be
+what that name will be renamed to.
+
+Therefore the function, f, will need to return Symbols or Strings.
+"""
+rename!(x::Indexer, f::Function) = rename!(x, [(x,f(x)) for x in x.names])
```
bicycle1885: I think `rename!` should take a function as the first argument for consistency with other standard functions: `map`, `foldl`, `foldr`, etc.

src/util/indexing.jl
```diff
@@ -0,0 +1,263 @@
...
+rename(x::Indexer, args...) = rename!(copy(x), args...)
+
+
+# Indexing into an index mapping single names to single numbers
+
+## Indexing with single name
+Base.getindex(x::Indexer, idx::Symbol) = x.lookup[idx]
+Base.getindex{S <: AbstractString}(x::Indexer, idx::S) = x.lookup[convert(Symbol, idx)]
```
bicycle1885: `Base.getindex(x::Indexer, idx::AbstractString) = ...`

src/util/indexing.jl
```diff
@@ -0,0 +1,263 @@
...
+Therefore the function, f, will need to return Symbols or Strings.
+"""
+rename!(x::Indexer, f::Function) = rename!(x, [(x,f(x)) for x in x.names])
+rename(x::Indexer, args...) = rename!(copy(x), args...)
+
+
+# Indexing into an index mapping single names to single numbers
+
+## Indexing with single name
+Base.getindex(x::Indexer, idx::Symbol) = x.lookup[idx]
```
bicycle1885: Why is the second argument name `idx`? I thought this is a key and the first argument is corresponding to an index (`x` is an abbreviation for `index`, right?).

src/util/indexing.jl
```diff
@@ -0,0 +1,263 @@
...
+end
+
+## Indexing with a vector of intergers
+function Base.getindex{I <: Integer}(x::Indexer, idx::Vector{I})
+    ns = x.names[idx]
+    return x[ns]
+end
+
+## Indexing with a range of integers
+function Base.getindex{I <: Integer}(x::Indexer, idx::UnitRange{I})
```
bicycle1885: I think `Base.getindex{I <: Integer}(x::Indexer, idx::Vector{I})` and `Base.getindex{I <: Integer}(x::Indexer, idx::UnitRange{I})` can be merged in to a single definition as `Base.getindex{I<:Integer}(x::Indexer, keys::AbstractVector{I})`.

bicycle1885:
I'm still not so clear about how this indexer will be used in your coming work. Can you clarify it a little bit more for me?

bicycle1885:
Also, there is a guideline for docstrings: http://julia.readthedocs.io/en/latest/manual/documentation/#documentation. I think it's time for us to follow this guideline.

Ward9250:
It's used in both Phylogenies to map node and tip names to the vertices, which in LightGraphs.jl are only ever stored as integers. It's further to be used in genotype count matrices to map names of individuals / samples to rows of the matrix, as well as names of populations / groups to multiple rows of the matrix, in addition to mapping names of alleles to columns. Being used for the same job in multiple places, it makes sense to encapsulate that data and functionality here, rather than having to define it all twice in two separate places.

src/util/indexing.jl
Ward9250: Fixed :+1:

src/util/indexing.jl
Ward9250: Fixed :+1:

src/util/indexing.jl
Ward9250: Fixed :+1:

src/util/indexing.jl
Ward9250: Fixed :+1:

src/util/indexing.jl
Ward9250: Fixed :+1:

src/util/indexing.jl
Ward9250: Fixed :+1:

src/util/indexing.jl
Ward9250: Fixed :+1:

src/util/indexing.jl
Ward9250: Was a throwback to when the code represented some code I used from DataFrames.jl, Fixed :+1:

src/util/indexing.jl
Ward9250: Also a throwback to some code I used that used to be in DataFrame.jl, before I overhauled the Indexer type. Fixed :+1:

Ward9250:
Thanks for review guys, I will implement Julia docstrings guidelines and do any remaining tweaks and merge tomorrow.

Ward9250:
With docstrings updated now, that is all fixes and suggestions from two reviewers implemented. I shall merge this today. Looking forward to using it in then Phylo and population genetics work. Thanks for looking over this guys I really appreciate it.

PR #162, opened 2016-04-26
Add a bunch of temp files to gitignore

kdmurray91:
- Cloned BioFmtSpecimens (created under test/ by test suite)
- KDE directory files
- Vim swap & misc temp files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,16 @@
 docs/build/
 docs/site/
+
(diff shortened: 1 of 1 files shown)


Ward9250:
Looks good to me, this is only the .gitignore, no BioJulia codebase is edited so I don't think there's need of review for something so simple.

PR #163, opened 2016-05-02
fix BigBed parser

bicycle1885:
This will fix #159 after https://github.com/BioJulia/BufferedStreams.jl/pull/11 and https://github.com/BioJulia/Libz.jl/pull/21. So, we need to tag and release new versions of BufferedStreams.jl and Libz.jl to make TravisCI happy.

Issue #159, opened 2016-04-25
BigBed Tests Fail


Ward9250:
The final two tests in the Intervals module fail.

```
BED → BigBed → BED round-trip: Error During Test
  Got an exception of type ArgumentError outside of a @test
  ArgumentError: ensureroom failed, IOBuffer is not writeable
   in ensureroom at ./iobuffer.jl:175
   in write at ./iobuffer.jl:319
   in write at io.jl:54
   in write at /home/travis/.julia/v0.4/Bio/src/intervals/bigbed.jl:300
   in bigbed_write_index at /home/travis/.julia/v0.4/Bio/src/intervals/bigbed.jl:1443
   in write_bigbed_bigwig at /home/travis/.julia/v0.4/Bio/src/intervals/bigbed.jl:1811
   in write at /home/travis/.julia/v0.4/Bio/src/intervals/bigbed.jl:1725
   [inlined code] from /home/travis/.julia/v0.4/Bio/test/intervals/runtests.jl:460
   in anonymous at no file:0
   in include at ./boot.jl:261
   in include_from_node1 at ./loading.jl:320
   [inlined code] from /home/travis/.julia/v0.4/Bio/test/runtests.jl:27
   in anonymous at no file:0
   in include at ./boot.jl:261
   in include_from_node1 at ./loading.jl:320
   in process_options at ./client.jl:280
   in _start at ./client.jl:378
BigBed Intersection: Error During Test
  Got an exception of type ArgumentError outside of a @test
  ArgumentError: ensureroom failed, IOBuffer is not writeable
   in ensureroom at ./iobuffer.jl:175
   in write_sub at ./iobuffer.jl:299
   in write at ./iobuffer.jl:316
   in writebytes at /home/travis/.julia/v0.4/BufferedStreams/src/sources.jl:154
   in writebytes at /home/travis/.julia/v0.4/BufferedStreams/src/sources.jl:9
   in flushbuffer! at /home/travis/.julia/v0.4/BufferedStreams/src/bufferedoutputstream.jl:39
   in close at /home/travis/.julia/v0.4/BufferedStreams/src/bufferedoutputstream.jl:121
   in writebytes at /home/travis/.julia/v0.4/Libz/src/sink.jl:144
   in flushbuffer! at /home/travis/.julia/v0.4/BufferedStreams/src/bufferedoutputstream.jl:39
   in close at /home/travis/.julia/v0.4/BufferedStreams/src/bufferedoutputstream.jl:121
[...]

PR #168, opened 2016-05-05
install NCBI BLAST+ before running Travis CI

bicycle1885:
This would be necessary to pass tests of the BLAST wrapper.


bicycle1885:
The Travis CI failure is due to #159, which will be fixed by #163. I'll merge this one since BLAST installation succeeded on Linux and OS X.

PR #169, opened 2016-05-06
remove encode_seq macro

bicycle1885:
This removes `@encode_seq` macro, which is not used anymore. I will merge this soon.

### DIFF
--- a/src/seq/bioseq.jl
+++ b/src/seq/bioseq.jl
@@ -82,34 +82,6 @@ end
 # Constructors
 # ------------
 
-# This is the body of the BioSequence constructor below. It's separated into a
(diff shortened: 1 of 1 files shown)


Ward9250:
How does it work now? I guess there's just two methods, a safe one and unsafe one?

bicycle1885:
It is now a function called [`encode_copy!`](https://github.com/BioJulia/Bio.jl/blob/505c41793465a1b2b612e5fe2f8b390d845f31f8/src/seq/bioseq.jl#L526), and actually I removed "unsafe" version because it makes the interface simpler and I couldn't see a dramatic performance improvement (though I didn't record the exact value).

PR #173, opened 2016-05-08
Seqtrie: a Trie over BioSequences

kdmurray91:
This is a wholesale looting of DataStructures.jl's Trie. I've modified it to work for BioSequences. Saves BioSequence -> string -> BIosequence round trip when using Tries.

At the moment something is broken in the prefix iterator, but all other tests pass.

Cheers,
K

### DIFF
--- a/src/seq/Seq.jl
+++ b/src/seq/Seq.jl
@@ -110,7 +110,21 @@ export
     DNAAlphabet,
     RNAAlphabet,
     AminoAcidAlphabet,
-    CharAlphabet
+    CharAlphabet,
+    SeqTrie,
+    DNATrie,
+    RNATrie,
+    AminoAcidTrie,
+    keys,
+    keys_with_prefix,
+    setindex!,
+    getindex,
+    get,
+    haskey,
+    hasprefix,
+    path
+
+
 
 using
     Compat,
--- a/src/seq/Seq.jl
+++ b/src/seq/Seq.jl
@@ -117,8 +117,6 @@ export
     AminoAcidTrie,
     keys,
     keys_with_prefix,
-    setindex!,
-    getindex,
     get,
     haskey,
     hasprefix,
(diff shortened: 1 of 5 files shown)


bicycle1885:
Having the Trie data structure would be nice, but I'm skeptical about the naive implementation in terms of performance and space efficiency.  As far as I know, the most efficient data structure for dynamic Trie is a double-array trie. You can find a benchmark here: http://www.tkl.iis.u-tokyo.ac.jp/~ynaga/cedar/. I'd like to implement a proof-of-concept code next week if I have time.

kdmurray91:
@bicycle1885 wow, you read my mind. I was eventually going to write a `Datrie.jl` package either implementing or wrapping `libdatrie` (which is what I have used in Axe, my demultiplexer). This is the result of a half-hour hackfest trying to get Tries from DataStructures to work with Biosequences.

I'll be very keen to see how you go with datrie.

Cheers,
K

bicycle1885:
According to the benchmark I posted above, `libdatrie` seems to be extremely slower to build a trie than other implementations (libdatrie: 209955.04 ns/key vs. cedar: 183.57 ns/key, for example). cedar seems to be the fastest in most cases and it is written in ~600 lines of C++. So, I hope it is not so hard to port their code to Julia. Moreover, [their implementation tips](http://www.slideshare.net/s5yata/dsirnlp-04s5yata) are available in Japanese (yes, my native language!).

kdmurray91:
Wow, that's surprising. I guess I never came across that performance issue, as I build a quite small trie once at the start, and nearly all time is spend in prefix lookups (for which datrie is fast enough, though still slower according to the benchmark of cedar). I know @dcjones also has `hat-trie`, which may be worth considering.

kdmurray91:
@bicycle1885 has this been superseded by your work with demultiplexers?

bicycle1885:
Actually not. What I made in #234 is too simple to say it as a trie tree; it cannot traverse a tree upwards. I wanted to implement a real double-array trie but I didn't have enough time to do that.

Issue #178, opened 2016-05-09
FASTAParser Returns Wrong BioSequence Type for RepeatMasked NucleotideSequences


timbitz:
So I was playing a bit with the current `master` branch BioSequences, when I noticed that when loading a big file of repeat masked human genomic regions, 1 out of the first 21 was returned as an AminoAcidAlphabet, and then the 22nd threw an encoding error for RNAAlphabet{4}.

```julia
for s in open( "data/ucgene_genomic.repeatMasked_N.fa", FASTA )
    println(s.name)
    println(typeof(s.seq))
end
uc057aty.1
Bio.Seq.BioSequence{Bio.Seq.DNAAlphabet{4}}
...
uc057awf.1
Bio.Seq.BioSequence{Bio.Seq.DNAAlphabet{4}}
uc057awx.1
Bio.Seq.BioSequence{Bio.Seq.AminoAcidAlphabet}
...
uc057aym.1
Bio.Seq.BioSequence{Bio.Seq.DNAAlphabet{4}}
ERROR: cannot encode Invalid RNA Nucleotide in Bio.Seq.RNAAlphabet{4}
 in enc64 at /cluster/home/timsw/.julia/v0.4/Bio/src/seq/bioseq.jl:643
 in encode_copy! at /cluster/home/timsw/.julia/v0.4/Bio/src/seq/bioseq.jl:549
 in call at /cluster/home/timsw/.julia/v0.4/Bio/src/seq/bioseq.jl:102
 [inlined code] from /cluster/home/timsw/.julia/v0.4/Bio/src/seq/fasta-parser.jl:114
 in read! at /cluster/home/timsw/.julia/v0.4/Bio/src/Ragel.jl:139
 in tryread! at /cluster/home/timsw/.julia/v0.4/Bio/src/Ragel.jl:218
 in next at /cluster/home/timsw/.julia/v0.4/Bio/src/Ragel.jl:278
 in anonymous at no file

```

[...]

PR #179, opened 2016-05-10
FASTA type

bicycle1885:
This adds type-specific parser for FASTA formats and fixes the sequence type inference reported in #178.

I've found that the naive Bayes classifier is too unreliable to classify sequence types and introduced a simple heuristic to improve the accuracy (thank you @timbitz!):
```julia
    if ACGTUN ratio > 0.95
        if #T ≥ #U
            return DNASequence
        else
            return RNASequence
        end
    else
        return AminoAcidSequence
    end
```

However, this heuristic is still not perfect when a sequence is too short or too skewed. So, I made it possible to specify the sequence type when creating a FASTA parser:
```julia
# set a sequence type as the third argument (optional)
for record in open("some.fasta", FASTA, AminoAcidSequence)
   # record.seq is surely AminoAcidSequence
end
```
The advantage of it is that Julia can infer the sequence type and will generate a more optimized program due to the type stability.


vchuravy:
I think that being able to specify the sequence type is really good. The automatic inference is cool, but if I am working on a pipeline I know what data I move through the system and I would rather have the system loudly fail, than silently move on and create invalid results.

bicycle1885:
@vchuravy Thanks :).

This is now ready to review. I'd like to merge this if you like the API.

src/Ragel.jl
```diff
@@ -229,18 +229,18 @@ end
...
 # ----------------------------------
 
-function Base.open{T<:FileFormat}(filename::AbstractString, ::Type{T}; args...)
+function Base.open{T<:FileFormat}(filename::AbstractString, ::Type{T}, args...; kwargs...)
     memory_map = false
     i = 0
-    for arg in args
+    for arg in kwargs
         i += 1
         if arg[1] == :memory_map
```
Ward9250: Is it mandatory that memory_map be the first two kwargs? If so can this be changed? I can imagine users occasionally getting the order wrong.

Ward9250:
LGTM :+1:

src/Ragel.jl
bicycle1885: This does not mean `memory_map` must be the first argument. `arg[1]` is a key of a keyword argument and `arg[2]` will be a value of it. This loop tries to find a pair of `memory_map => true/false` and thus `memory_map` can be located at any position.

Ward9250: Ah I see each iteration `arg` looks like a tuple. I follow you now. Seems mornings are not good to me reading code.

timbitz:
Sweet.

PR #181, opened 2016-05-12
improve style of docs

bicycle1885:
I'm now trying to improve the style of our documents. 

Here are screenshots:

**Current**
screen shot 2016-05-12 at 18 01 39

**New**
screen shot 2016-05-12 at 18 02 03

Pending issues are:
*  Syntax highlighter improvement: https://bitbucket.org/birkenfeld/pygments-main/pull-requests/594/fix-julialexer-and-juliaconsolelexer/diff
* Stylistic docstrings: https://github.com/MichaelHatherly/Documenter.jl/issues/48

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,6 +1,6 @@
 using Documenter, Bio
 
-custom_deps() = run(`pip install --user pygments mkdocs mkdocs-material`)
+custom_deps() = run(`pip install --user pygments mkdocs mkdocs-cinder`)
 @osx? makedocs(doctest = false) : makedocs()
 deploydocs(
            repo = "github.com/BioJulia/Bio.jl.git",
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -11,20 +11,18 @@ repo_url:  https://github.com/BioJulia/Bio.jl
 copyright: Copyright (c) 2016 The BioJulia Organisation
 
 # Documentation and theme
-theme: material
+theme: cinder
 docs_dir: 'build'
 
 # Options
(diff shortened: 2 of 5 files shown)


Ward9250:
I've been thinking about this recently too. I was wondering if we could/should make the docs with the same theme as the official Julia documentation? Which I believe is mkdocs theme with some extra branding bits. Other than the docs - which are automatic, I've been creating the branding and webpage to have a colour scheme and font that is consistent with Julia's. It would be awesome to apply this principle to the docs as well.

bicycle1885:
I'm not sure whether there is a consensus on "Julian color scheme". The logo of Julia may be a such thing. But, at least, I don't think the current coloring looks nice to most people.

I recently found a nice color scheme called "cinder": http://sourcefoundry.org/cinder/. This is very easy to apply to our docs and looks much better than the current one. If you like it, why don't we apply it until we find a better color scheme for branding?

Ward9250:
Cinder looks lovely, I'm happy to go with it. I agree the current docs is seems to be blindly following the google materials design trend, which I'm not that fond of. The branding and logos only use the colours and fonts of the Julia logo presently, which I think makes sense, but actually for the docs maybe we should go with what looks best, regardless of branding.

bicycle1885:
> But actually for the docs maybe we should go with what looks best, regardless of branding, they have to be functional above any stylistic concern.

Exactly.

I've squashed commits into one that uses the cinder theme. Actually, I wanted to include my fork of pygments to make code highlighting better, but it is a separated issue. Please merge this if it looks good to you.

Issue #147, opened 2016-03-27
Omniref for Code Review


Ward9250:
We're going to give a trial run of using Omniref to formalise the code review process a bit more when a PR enters a RTR stage. I've sent invitations to the Org owners and the reviewers team to try the functionality of omniref out. Let's use if for the next few PRs and see how it goes.

Ward9250:
I think the recent review tools for github have made the need for Omniref moot.

PR #182, opened 2016-05-12
Fix coding style

bicycle1885:


Ward9250:
Opening an omniref code review for this, to push ahead with the attempt to make our code reviews more formal #147 - hope that's ok, the initial problems with Omniref that we had appear to have been fixed so thought we should try again :)

Ward9250:
Do you think we should have a consistent header style for all our code files? That say the title of the file, a short description, and then Copyright "The BioJulia Organisation" / "BioJulia" with the MIT licence? I've seen other software projects do this.

bicycle1885:
> Do you think we should have a consistent header style for all our code files? That say the title of the file, a short description, and then Copyright "The BioJulia Organisation" / "BioJulia" with the MIT licence? I've seen other software projects do this.

I like such style, but full description of the license would be too verbose to place in all of the source files. I think one-line description and link are enough like Julia does:
```julia
# This file is a part of Julia. License is MIT: http://julialang.org/license

...
```
<https://github.com/JuliaLang/julia/blob/master/base/bool.jl>

So, the following is my favorite style of the header:
```julia
# Title
# =====
#
# Short description.
#
# [Long description (optional)]
#
# This file is a part of BioJulia. License is MIT: <link to the license file>
```

Ward9250:
That's awesome, I like that header too :) I've added your example to the Code Style page of the Contributing docs.

bicycle1885:
Okay, I will insert headers soon :)

bicycle1885:
Question: Should we stick to a single license statement in all packages developed under BioJulia organization? If so, which file should we point to in our headers?

Ward9250:
@bicycle1885 At present our contribution guidelines state that packages submitted to BioJulia should be licensed under the MIT Licence. I think each package should have it's own LICENCE file, as I know you're supposed to distribute a copy of the licence with software. But we could have a copy of the LICENCE on our website (when I get it finished!) and people can link to that.

bicycle1885:
I see. I also think that each package should have a copy of the original license file.

The current URL, I added in the headers, is somewhat dynamic because it is a link to a file hosted on GitHub. Eventually I think we need to place it in somewhere static.

Ward9250:
I agree that sounds the best!

bicycle1885:
I'd like to merge this one soon.

PR #184, opened 2016-05-16
Added the Dating submodule to Phylo.

Ward9250:
This PR adds a Dating submodule to Phylo. Initially it contains a simple procedure I designed for a larger software project on my PhD, which returns an estimation of the coalescence time between two sequences, based on assumptions formalised by modelling mutation accumulation between sequences as a binomial process.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -8,3 +8,4 @@ Iterators
 Libz 0.1.1
 LightGraphs
 LightXML
+Roots
--- a/src/phylo/Phylo.jl
+++ b/src/phylo/Phylo.jl
@@ -67,6 +67,6 @@ include("phylogeny.jl")
 include("node_basics.jl")
 include("branch_basics.jl")
 include("manipulation.jl")
-
+include("dating.jl")
 
 end # module Phylo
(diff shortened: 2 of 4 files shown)


Ward9250:
This is ready for review. I can see test fail, something to do with failing when precompiling LightGraphs? This didn't happen when testing the Phylo module previously, so not sure what's going on there.

vchuravy:
It would be nice to have docstrings for that function explaining what it does and also (if possible) give a reference.

Ward9250:
Ok sure I can add those.

Ward9250:
Added docstring, and also I now know the cause of the error, it's because a function in StatsBase, imported by LightGraphs no longer exists. This is being accounted for by LightGraphs, but won't change and fix the error until the next release.

bicycle1885:
I can't see any tests related to this module. Did you forget to push them or something?

Ward9250:
I just didn't add them yet whilst I chased up the LightGraphs.jl error business.

Ward9250:
Ok I've added tests. If they pass I'd like to merge if no furthur objections.

bicycle1885:
👍

PR #152, opened 2016-04-01
Exact search/rsearch for sequences.

bicycle1885:
This adds `search` and `rsearch` methods for sequences. These works like `Base.search`/`Base.rsearch` for strings, but match ambiguous symbols, too. In the following example, the target sequence `dna"ACGNATT"` has an ambiguous nucleotide `DNA_N` at the 4th symbol, which matches any character in the queries: 

```julia
julia> using Bio.Seq

julia> search(dna"ACGNATT", dna"ACG")
1:3

julia> search(dna"ACGNATT", dna"CGT")
2:4

julia> search(dna"ACGNATT", dna"GGA")
3:5

```

I'm now planning to implement three kinds of sequence search utils:

1. Exact search (this pull request)
2. Approximate search that allows errors (substitution/insertion/deletion) up to `k` (#153)
3. Regular expression search (#143)

The first one (this pull request) is ready to review. The second one is not yet a pull request, but I already wrote the main algorithm. The third one is also close to be a ready-to-review quality.

PR #185, opened 2016-05-17
define eltype for Kmers

bicycle1885:
(Now I'm splitting #152 into several pull requests).

This will add `eltype` for the `Kmer` type. I believe nobody is against this change, so I will merge this soon.

### DIFF
--- a/src/seq/kmer.jl
+++ b/src/seq/kmer.jl
@@ -201,6 +201,7 @@ Base.endof(x::Kmer) = length(x)
 Base.start(x::Kmer) = 1
 Base.next(x::Kmer, i::Int) = unsafe_getindex(x, i), i + 1
 Base.done(x::Kmer, i::Int) = i > length(x)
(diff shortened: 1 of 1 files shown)


Ward9250:
Yeah I don't think anyone objects to this. @bicycle1885 Do Kmers still have to be short enough to fit into a single integer? I was just wondering as some methods I'd like to implement over the 1.5 years I'm funded for Bio.jl stuff, uses fairly long kmers.

bicycle1885:
What kind of extension do you have in your mind? Up to 64nt using `UInt128` or arbitrary length?

Ward9250:
I was wondering is using bigger Ints is something that is easily done yes.

bicycle1885:
Sounds reasonable but 128bit integers won't fit in registers of modern computers, which may significantly kill the performance of many operations. Anyway it would be worth trying.

PR #186, opened 2016-05-17
add iscompatible

bicycle1885:
(This was a part of #152.)

This pull request adds `iscompatible(x, y)` function for biological symbols. This function returns `true` if and only if two symbols are compatible with each other. For example, `DNA_A == DNA_N` is `false` but `iscompatible(DNA_A, DNA_N)` is `true` because `DNA_N` may be `DNA_A`.

Bikeshedding: Is there more suitable name to native English speakers than `iscompatible`?

### DIFF
--- a/docs/src/man/seq.md
+++ b/docs/src/man/seq.md
@@ -192,6 +192,7 @@ Invalid DNA Nucleotide
     {docs}
     alphabet
     gap
+    iscompatible
 
 
 ## Sequences
--- a/src/seq/Seq.jl
+++ b/src/seq/Seq.jl
@@ -44,6 +44,7 @@ export
     RNA_B,
     RNA_N,
     RNA_Gap,
+    iscompatible,
     Sequence,
     BioSequence,
     DNASequence,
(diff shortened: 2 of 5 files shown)


kdmurray91:
This could be a little dodgy, but in the case of ambiguites, what about using the set membership symbol? I.e. `DNA_A \in DNA_N === true`.


I can't think of a more natural english phrasing than your suggestion of `iscompatible`. 
But I'm Australian, so you may want to refer to the truer English of @Ward9250  et al. :stuck_out_tongue_winking_eye:

bicycle1885:
`iscompatible` is different from the membership operator. Using the notations of set theory, it is defined as: `iscompatible(x, y) := x ∩ y ≠ ∅ `. So for example, `iscompatible(DNA_M, DNA_R)` is `true` in the IUPAC nomenclature ('M' is 'A' or 'C' and 'R' is 'A' or 'G'). But operators like set theory may be useful in some cases.

bicycle1885:
I'm going to merge this.

PR #187, opened 2016-05-17
rename internal macros to avoid overwriting others

bicycle1885:
This overwriting has killed tests on the banded Needleman-Wunsch algorithm on Julia v0.5.


Ward9250:
Does this fix the errors on Julia nightlies tests? The error on those tests is no longer related to Align, but is instead some sort of seg fault.

bicycle1885:
Partially. There is still a pile of problems to run tests on the current nightlies. I think there is nothing we can do to avoid the segfault because it would be due to the compiler's bug. You can work around it by passing the `--compilecache=no` option, then you will see another problem.

PR #188, opened 2016-05-18
fix a docstring of KmerCounts (and potential bug)

bicycle1885:


bicycle1885:
Merged since this is just a doc fix.

PR #190, opened 2016-05-19
SymbolRange, cyclic arithmetic, and more informative `show` method.

bicycle1885:
This pull request includes three topics:

1. Introduce `SymbolRange` type which is similar to `Base.UnitRange` but specialized for our biological symbols.
2. Make symbol arithmetic cyclic; for example, `DNA_A + 16` is `DNA_A`, not `DNA_INVALID`.
3. Make `show(io::IO, Nucleotide|AminoAcid)` more informative; for example `show(AA_A)` shows `AA_A` not `A`.

2 depends on 1, and 3 is a thing I want for a long time.
```
julia> seq = dna"ACGT"
4nt DNA Sequence:
ACGT

julia> seq[1]  # if we show 'A', we cannot see it from RNA and amino acid
DNA_A

```

If you need one-letter representation, you can call `print` instead of `show`.

### DIFF
--- a/docs/src/man/seq.md
+++ b/docs/src/man/seq.md
@@ -179,13 +179,13 @@ julia> DNA_T - DNA_C  # difference
 
 ```
 
-Note that these operations do **not** check bounds of valid range:
+Note that these operations are cyclic:
 ```jlcon
-julia> DNA_C - 2
-Invalid DNA Nucleotide
+julia> DNA_C + 15
+A
 
-julia> DNA_A + 30
-Invalid DNA Nucleotide
+julia> DNA_A - 1
+-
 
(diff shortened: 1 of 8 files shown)


Ward9250:
I like this, so if I understand how 2. works, DNA_A + 4 would also be DNA_A?

bicycle1885:
No, DNA arithmetic is modulo of 16. This is because `DNANucleotide` type has 16 nucleotides including ambiguous nucleotides.

Ward9250:
Oh that's right I forgot there's ambiguities now!

bicycle1885:
I will merge this soon.

PR #193, opened 2016-05-20
Added more tests for phylogeny types.

Ward9250:
Lets bump up the test coverage for Phylo!

### DIFF
--- a/src/phylo/phylogeny.jl
+++ b/src/phylo/phylogeny.jl
@@ -104,11 +104,6 @@ function Phylogeny{T <: AbstractString}(taxa::Vector{T})
     return Phylogeny(convert(Vector{Symbol}, taxa))
 end
 
-"""
-    Base.copy{C, B}(tree::Phylogeny{C, B})
-
-Create a copy of a phylogenetic tree.
-"""
 function Base.copy{C, B}(tree::Phylogeny{C, B})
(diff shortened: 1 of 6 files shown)


src/phylo/phylogeny.jl
```diff
@@ -105,6 +104,11 @@ function Phylogeny{T <: AbstractString}(taxa::Vector{T})
     return Phylogeny(convert(Vector{Symbol}, taxa))
 end
 
+"""
+    Base.copy{C, B}(tree::Phylogeny{C, B})
+
+Create a copy of a phylogenetic tree.
+"""
```
bicycle1885: This docstring is redundant; this method does the same thing as `Base.copy`, which is already documented.

src/phylo/phylogeny.jl
```diff
@@ -113,6 +117,21 @@ function Base.copy{C, B}(tree::Phylogeny{C, B})
     return Phylogeny{C, B}(g, i, v, d, tree.ntaxa, tree.rooted, tree.rerootable)
 end
 
+"""
+    Base.:(==)(x::Phylogeny, y::Phylogeny)
+
+Test if two phylogenies are considered equal.
+"""
```
bicycle1885: Same as `Base.copy`.

PR #194, opened 2016-05-21
clean up mismatches and docs

bicycle1885:
Mostly stylistic changes of `mismatches` and other docstrings.

### DIFF
--- a/src/seq/bioseq.jl
+++ b/src/seq/bioseq.jl
@@ -777,75 +777,54 @@ end
 # -----------------
 
 """
-`mismatches(a::NucleotideSequence, b::NucleotideSequence, [nmatches=false])`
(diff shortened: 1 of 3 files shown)


Ward9250:
I've commented in Omniref (which I prefer as it saves all comments forever, and I think that's useful when people want to see how decisions were made in the code historically), on github specific line comments get rendered outdated.

Ward9250:
Test coverage for Phylo is now nearly 100%. And docstrings have been removed as suggested by @bicycle1885 :+1:

Ward9250:
@bicycle1885 Just catching up on things I forgot about in the thesis writing panic. I'm happy for this to go in if tests pass after rebase :+1:

PR #196, opened 2016-05-21
rename Blast to BLAST

bicycle1885:
This renames the `Blast` module to `BLAST` and places it under the `Bio.Tools` module. So now you need to call `using Bio.Tools.BLAST` instead of `using Bio.Blast` to use this functionality.


Ward9250:
Since this is bike shedding, probably go ahead and merge this whenever you're ready :+1:

bicycle1885:
Yeah, but I just want an approval from @kescobo.

Ward9250:
Is it also possible to use Homebrew WinRPM or BinDeps to install a BLAST along with Bio.jl? I didn't know if those things are for downloading entire programs like that or just like C libraries and such.

bicycle1885:
BLAST+ is installed for Travis CI (https://github.com/BioJulia/Bio.jl/blob/f477d42696e1999616499bb38deed89cb896a41d/.travis.yml#L11-L19). But I don't think it is a good idea to automatically install this kind of CLI tools by default because it may disturb user's environment and these wrappers are not necessary tools to use other functionalities of Bio.jl

Ward9250:
Do you know how to install it for appveyor? I would do it but I'm afraid I haven't used Windows since before I first went to University. So the last Windows box I saw was Vista 😆

bicycle1885:
I used Windows 7 but I don't have access now. So, I opened an issue for that #197. Please help us!

kescobo:
@bicycle1885 no objection here.

bicycle1885:
Thank you. Merged.

PR #199, opened 2016-05-23
rename Blast to BLAST (take 2)

bicycle1885:
Renamed `BlastResult` -> `BLASTResult`.


bicycle1885:
Merged since this is continued from #196 .

PR #200, opened 2016-05-23
Begun docs for Phylo.

Ward9250:
Starting man-pages for Bio.Phylo.


Ward9250:
Closing and deleting branch as this is not going to be done until we know more about whether Phylo will have it's design changed as a result of #263.

PR #202, opened 2016-05-24
Fix kmer

bicycle1885:
This includes following changes:

* fix k-mer arithmetic (e.g. `DNAKmer("AC") + 0x01` threw an exception)
* deprecate `kmer`, `dnakmer` and `rnakmer` in favor of `Kmer`, `DNAKmer` and `RNAKmer` respectively.
* export `Codon` from `Bio.Seq`, which is an alias of `Kmer{RNANucleotide,3}`.


Ward9250:
LGTM :+1: Do you think we could also have a codon type for DNA sequences?

bicycle1885:
Not sure. But the term "codon" means a triplet of **RNA** nucleotides in most cases. If users need a codon type for DNA, it is quite easy to define that.

Ward9250:
It's also common to talk of codons in DNA sequences if the sequence is of a gene or ORF. It can be useful to think of codons in DNA when looking at mutations in intron regions and stuff that causes frameshift and so on.

bicycle1885:
I see. I tried to define `DNACodon` and `RNACodon` types and remove `Codon` in #204, which may be more explicit than `Codon`.

PR #204, opened 2016-05-24
define DNACodon and RNACodon

bicycle1885:
as aliases of DNA and RNA triplet, respectively (https://github.com/BioJulia/Bio.jl/pull/202#issuecomment-221238453). 

`Codon` type is removed since it is a little bit confusing.


Ward9250:
:+1:

PR #205, opened 2016-05-30
Genotype Count Matrices and Binary formatted genotypes

Ward9250:
Here we have a PR for Genotype Count Matrices and a type for storing genotypes (such as SNP data from VCF files) in a binary format.

### DIFF
--- a/src/var/binary_snps.jl
+++ b/src/var/binary_snps.jl
@@ -6,6 +6,29 @@ SNPs are stored in a binary format, and also that the SNPs are binary,
 in that they have two states, a reference state and an alternate state.
 =#
 
+
+# An alphabet for storing SNPs
+
+"""
+SNPs alphabet.
+"""
+immutable SNPAlphabet{n} <: Alphabet
+
+for n in (2, 4, 8)
+    @eval begin
+        bitsof(::Type{SNPAlphabet{$n}}) = $n
+        alphabet(::Type{SNPAlphabet{$n}}) = UInt8(0):UInt8($n)
+    end
+end
+
+Base.eltype{n}(::Type{SNPAlphabet{n}}) = UInt8
+
+
+
+
+
+typealias SNPSequence{n} BioSequence{SNPAlphabet{n}}
+
 type BinarySNPs
     snps::Vector{Vector{UInt8}}
     numberOfLoci::Int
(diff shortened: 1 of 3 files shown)


src/var/binary_snps.jl
```diff
@@ -0,0 +1,167 @@
+
+#=
+Objects of type BinarySNPs store the SNPs of an individual
+The name BinarySNPs is double edged: it refers to both the fact that
+SNPs are stored in a binary format, and also that the SNPs are binary,
+in that they have two states, a reference state and an alternate state.
```
kdmurray91: How are hets handled here? I'm assuming that the rows of this matrix are *alleles* rather than SNPs, and that the Index work is used to select the N alleles for each SNP?

Ward9250: This type stores the snaps of a single individual, rather than a matrix of individuals. SNPs are stored in binary format using vectors. So 0 indicates the position has the reference allele, whilst a 1 indicates the position has the alternative allele. If the individual is haploid, then it will have one binary vector. If it is diploid it may have two binary vectors and so on. So a het in a diploid individual would mean there is a 1 in one of the vectors at a given position, but a 0 in the other vector at the same position. This design is heavily influenced by the class of a similar name in R's adegenet. So anything we can improve on this would be really awesome!

Ward9250: I had considered going a slightly alternate route and rather than having multiple binary vectors for diploids and so on. Going the BioSequence route and making some alphabet, where the number of bits used to store each snp is the ploidy of the individual.

kdmurray91:
Hmm, yes. How about storing states as an `IntArray` with `log_2(number of states)` bits? I guess this is pretty similar to how biosequences are stored, right?

Ward9250:
That could work, I think the best way to encode variant data is going to
depend a lot on the data - kindof as with sequences. Most SNPs in a VCF
file say have only two states, so a compact bit vector, as SNPBin currently
uses is good. But then that starts to be less useful when you are looking
at more highly polymorphic things or anything with more than two states.
Unless three states could be represented in the SNPBin type with two snps
with the same base location.

These datatypes will be the first thing I start working on after my thesis
submission, as they look like they will be very useful for a project I'm
likely expected to do when my postdoc starts. So any discussion about
design or what features people want will be very helpful. What we have
currently are initial ideas.

Ward9250:
I'm going to close this as it's all invalidated and made moot by my current work.

PR #206, opened 2016-05-31
BAM (retake)

bicycle1885:
This is a retake of #101 and will supersede it. This is built on top of @dcjones's work, and still needs additional functionality and refactoring.


bicycle1885:
This will be replaced by #265.

PR #210, opened 2016-06-04
fix FASTA writer interfaces

bicycle1885:
This restores the `FASTASeqRecord` type alias, which I removed without careful thought.

I think `FASTAMetadata` will not be used by users because `FASTASeqRecord` is able to create a FASTA record for serialization. So, I removed it from the exported list (#207).

I will fix interfaces for other formats very soon.

### DIFF
--- a/docs/src/man/reading.md
+++ b/docs/src/man/reading.md
@@ -57,3 +57,25 @@ making it easy to allocate an empty entry for any parser stream.
 entry = eltype(stream)()
 ```
 
+
+## Writing data
+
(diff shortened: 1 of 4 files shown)


Ward9250:
@bicycle1885 Is this RTR? If so I will say :+1:, I'd like t use it in my current scripts to write some fasta alignment files. I'll write docs whilst I'm at it too, to save you time if you'd like.

bicycle1885:
<https://github.com/JuliaLang/METADATA.jl/pull/5324>  is released, which will fix the CI failure. I will merge this soon after I check the tests pass.

PR #211, opened 2016-06-05
fix FASTQ writer interfaces

bicycle1885:
As per `FASTASeqRecord`, this makes it easier to create FASTQ sequence records that can be serialized using `FASTQWriter`.


Ward9250:
LGTM :+1:

bicycle1885:
Thanks :)

PR #212, opened 2016-06-06
Sliding windows return their start and ends.

Ward9250:
This modifies the code for sliding windows slightly, such that in a for loop now you get a tuple of values. The first value is the start and end position of the window, and the second value is the view of the sequence / array itself in that region.

### DIFF
--- a/src/util/windows.jl
+++ b/src/util/windows.jl
@@ -87,9 +87,8 @@ end
 
 @inline function Base.next(it::EachWindowIterator, state::Integer)
     i = state
-    region = i:i + it.width - 1
-    window = sub(it.data, region)
-    return (region, window), i + it.step
+    window = sub(it.data, i:i + it.width - 1)
+    return (i, window), i + it.step
 end
 
 # Extra next method to account for fact than strings don't have sub method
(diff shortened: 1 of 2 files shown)


src/util/windows.jl
```diff
@@ -87,8 +87,9 @@ end
 
 @inline function Base.next(it::EachWindowIterator, state::Integer)
     i = state
-    window = sub(it.data, i:i + it.width - 1)
-    return window, i + it.step
+    region = i:i + it.width - 1
```
bicycle1885: If the window size is fixed, returning the end position is not needed, which is also compatible with our each-kmer iterator.

Ward9250: I think this is what you meant: I altered this so now it returns (i, window).

bicycle1885: Exactly.

test/util/windows.jl
```diff
@@ -104,19 +104,19 @@ end
                 @test missed(ribitr) == expectedMissed
                 @test missed(aasitr) == expectedMissed
                 for win in arrres
-                    @test length(win) == winwidth
+                    @test length(win[2]) == winwidth
```
bicycle1885: Testing `win[1]` is desirable.

Ward9250: Added tests for `win[1]`, will see if it passes suite.

PR #214, opened 2016-06-14
add BEDWriter

bicycle1885:
Like FASTAWriter and FASTQWriter, this adds BEDWriter and its interfaces for serializing intervals in BED format.

### DIFF
--- a/src/intervals/bed.jl
+++ b/src/intervals/bed.jl
@@ -73,6 +73,17 @@ end
 "An `Interval` with associated metadata from a BED file"
 typealias BEDInterval Interval{BEDMetadata}
 
+function Base.open(filepath::AbstractString, mode::AbstractString, ::Type{BED};
+                   n_fields::Integer=-1)
+    io = open(filepath, mode)
+    if mode[1] == 'r'
+        return open(BufferedInputStream(io), BED)
+    elseif mode[1] ∈ ('w', 'o')
+        return BEDWriter(io, n_fields)
+    end
+    error("invalid open mode")
+end
+
 
 # Parser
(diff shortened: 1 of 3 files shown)


src/intervals/bed.jl
```diff
@@ -73,6 +73,17 @@ end
...
 
+function Base.open(filepath::AbstractString, mode::AbstractString, ::Type{BED};
+                   n_fields::Integer=-1)
+    io = open(filepath, mode)
+    if mode[1] == 'r'
+        return open(BufferedInputStream(io), BED)
+    elseif mode[1] ∈ ('w', 'o')
+        return BEDWriter(io, n_fields)
+    end
+    error("invalid open mode")
```
Ward9250: Does this if else create type uncertainty?

Ward9250:
Oooooo! :) The consistent writing interface is wonderful. I was thinking (separate issue but related) we should start some developer documentation pages in our docs explaining the consistent IO interfaces and how to write your own abstract writers, I'm happy to write some of it in July if that's agreeable. I look forward to implementing my tree writers with this new format :+1:

src/intervals/bed.jl
bicycle1885: Yes, but it is unavoidable unless we unify the parser and the writer types into a type.

bicycle1885:
Thank you 😄 . I'm really grateful to you for documentation because writing English is still a hard work to me.

After this change, I'd like to cut a branch for the release of Bio.jl 0.2 if you don't mind.

Ward9250:
Sounds good to me!

PR #216, opened 2016-06-30
Added sidecar to docs

Ward9250:
Thought it would be cool for readers of docs if it was possible to have Gitter sidecar in there, so if they see something and have questions, they can immediately pop it open and talk to someone.

I believe I've done this correctly with mkdocs, _but_ I'm not 100% sure, and we can't see how it affects the page until we merge to master.

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -28,6 +28,8 @@ extra_css:
 extra_javascript:
   - https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS_HTML
   - assets/mathjaxhelper.js
+  - https://sidecar.gitter.im/dist/sidecar.v1.js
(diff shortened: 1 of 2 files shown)


bicycle1885:
Looks cool 👍

Ward9250:
I'm going to try alter a fork of the cinder theme to include the sidecar
and load the JS in asynchronously in a deferred manner, which will speed
the loading of the sidecar and page up, and the template will be able to
use the name of the package/site to just load the correct Gitter room
without package authors or maintainers worrying about it.
Docs in packages can then use the cinder fork as a submodule and so all
docs of all current and future packages can take advantage of this.

On Fri, Jul 1, 2016 at 2:47 PM, Kenta Sato (佐藤 建太) <notifications@github.com

PR #217, opened 2016-07-01
Edited docs to use mkdocs-biojulia

Ward9250:
Ok, here we have Bio.jl docs using the mkdocs-biojulia theme adapted from cinder.
Presently mkdocs-biojulia differs from cinder in the following ways:

1). The Gitter sidecar is incorporated into the html template and js is loaded in async deferred manner.

2). The custom css and js that was in the assets folder of docs (friendly.css, custom.css, and sidecar.js) have been incorporated into the the css of mkdocs-biojulia.

In theory then, mkdocs.yml of BioJulia packages should just need to set theme: mkdocs-biojulia, and this should work.

Note we still need additional CSS and javascript fields in mkdocs.yml for Documenter.css and the mathjax stuff. But it might be possible to eliminate the need for those fields too.


bicycle1885:
If the `mkdocs-biojulia` package is not registered in PyPI, you cannot install it by its name. You may need to specify the URL of our fork of the cinder theme like this: http://stackoverflow.com/questions/8247605/configuring-so-that-pip-install-can-work-from-github.

Ward9250:
I have registered it on PyPI.

On Fri, Jul 1, 2016 at 5:54 PM, Kenta Sato (佐藤 建太) <notifications@github.com

Ward9250:
Ok, everything looks as though it tested and build docs correctly - but of course skipped deployment so again we can't see the result until merge, but i guess Documenter did all the pip install of mkdocs-biojulia from PyPI correctly, because there was no error thrown.

bicycle1885:
Nice.

Ward9250:
Hopefully this will allow us to just say biojulia as the theme, and then that makes for simpler mkdocs.yml files for packages, and man pages that are completely consistent.

Ward9250:
This works now :+1:  So just using `theme: biojulia` in `mkdocs.yml` works now.

PR #218, opened 2016-07-02
Incoroprate an FOSGAA global alignment

BenJWard:
So we've had the BioJulia/fogsaa repo for a while, which just works on strings and implements a fast B&B global alignment algorithm, and with recently using Bio.Align pairwise global alignments in my scripts I was impressed, although speed was not the quickest - given I was aligning very large contigs of a FALCON assembly. So I thought hey why not get fogsaa incorporated as an option in the global pairalign.

There are three steps I currently identify, I've already made a commit doing a bit of step 1:
- [ ] Go through code, make variable names more meaningful, refactor any pieces that could be more simply expressed, and make sure it's julian, and add comments for the lay reader to understand what the algorithm does.
- [ ] Refactor code so it is of the form currently used in Bio.Align pairwise: There should be a FOGSAA type just as there is a NeedlemanWunsch type and so on, that stores data. The alignment algorithm should then be factored so as it fits with the current framework i.e. there is a run method for the algorithm type and so on.
- [ ] A further refactor of the code so as it works not just with strings but all our sequence types, and makes use of Anchors and Operations and Trace, instead of defining its own, which is what it currently does.

### DIFF
--- a/src/align/pairwise/algorithms/fogsaa.jl
+++ b/src/align/pairwise/algorithms/fogsaa.jl
@@ -7,6 +7,62 @@
 # This file is a part of BioJulia.
 # License is MIT: https://github.com/BioJulia/Bio.jl/blob/master/LICENSE.md
 
+type FOGSAA
+    alignments::Matrix
+    queue::Vector{Vector{FOGSAABranch}}
+    Mi
+    mi
+
+    function FOGSAA(m, n, s)
+        g0 = m - n
+    	gapMatch = g0 == 0 ? 0 : s.gap_start + abs(g0) * s.gap
+        m0 = g0 < 0 ? m : n
+    	gapMiss = gapMat + m0 * s.mismatch # gm
+    	gapMatch += m0 * s.match
+        alignments = zeros(UInt8, m + 1, n + 1)
+        scores = fill(gapMiss, m + 1, m + 1)
+    	scores[1, 1] = 0
+        Mi = mi = gapMatch - gapMiss + 1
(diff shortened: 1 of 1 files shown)


bicycle1885:
I worry the code you based on is not distributed under any OSS license, so I think we need to write it from scratch.

Ward9250:
The code I'm using is from BioJulia/fogsaa which is a fork of kjenova/fogsaa, there's no license assigned to it. Other than a reference to the nature paper they do not state where it comes from, I assumed they wrote it themselves based on a reading of the nature paper.

bicycle1885:
If so, the original author holds all rights and we cannot redistribute it without their permission.

Ward9250:
I've just done some digging, and the original FOGSAA is a C++ program which also, like the julia version of FOGSAA is available for download, and as in all science, you are told to download and check out the source code and it is freely available. However, here again, there is not license, and so technically all rights are reserved, and since downloading implicitly means making a copy, it's a grey area as to whether or not a scientist is even allowed to download, compile, and use the software, even though it has been published in Nature, with the intent of it being used! Therefore I feel the lack of license has been an oversight on behalf of the authors. I will chase this up and report back.

Ward9250:
Ok I've sent an email to @kjenova

Ward9250:
A license was added which differs only from the MIT in that there is no copyright. I also have an email from kjenova with express permission for us to use the fogsaa repo. :+1:

BenJWard:
This should be moved to BioAlignments.jl

PR #223, opened 2016-07-23
add gc_content

bicycle1885:
This adds the `gc_content` function to calculate GC content of a sequence.

### DIFF
--- a/src/seq/sequence.jl
+++ b/src/seq/sequence.jl
@@ -108,6 +108,37 @@ Base.findfirst(seq::Sequence, val) = findnext(seq, val, 1)
 Base.findlast(seq::Sequence, val) = findprev(seq, val, endof(seq))
 
 
+"""
+    gc_content(seq::Sequence)
+
+Calculate GC content of `seq`.
+"""
+function gc_content(seq::Sequence)
+    gc = 0
+
+    if eltype(seq) == DNANucleotide
+        for x in seq
+            if x == DNA_C || x == DNA_G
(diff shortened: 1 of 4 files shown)


src/seq/sequence.jl
```diff
@@ -108,6 +108,37 @@ Base.findfirst(seq::Sequence, val) = findnext(seq, val, 1)
...
+function gc_content(seq::Sequence)
+    gc = 0
+
+    if eltype(seq) == DNANucleotide
+        for x in seq
+            if x == DNA_C || x == DNA_G
+                gc += 1
+            end
+        end
+    elseif eltype(seq) == RNANucleotide
```
vchuravy: For me it seems this could be written without the `if elseif` statement, by  using Julias dispatch and definining

```
G(::Type{DNANucleotide}) = DNA_G
C(::Type{DNANucleotide}) = DNA_C
G(::Type{RNANucleotide}) = RNA_G
C(::Type{RNANucleotide}) = RNA_C

function gc_content(seq::Sequence)
   gc = 0
   for x in seq
      if x == T(eltype(seq)) || G(eltype(seq))
         gc += 1
      end
   end
   gc
end
```

bicycle1885: Sounds nice 👍

bicycle1885:
I will merge this after squashing commits if no objections.

Ward9250:
Looks fantastic @bicycle1885 :+1:

PR #224, opened 2016-07-24
add rand(::Kmer) to generate random k-mers

bicycle1885:
This adds a k-mer generator.


bicycle1885:
This is included in #225.

PR #225, opened 2016-07-24
Shuffle BioSequence and Kmer

bicycle1885:


Ward9250:
LGTM mate! :)

src/seq/bioseq.jl
```diff
@@ -237,6 +237,16 @@ alphabet{A}(::Type{BioSequence{A}}) = alphabet(A)
 Base.length(seq::BioSequence) = length(seq.part)
 Base.eltype{A}(::Type{BioSequence{A}}) = eltype(A)
 
+function Base.count(f::Function, seq::BioSequence)
+    n = 0
+    for x in seq
+        if f(x)
+            n += 1
+        end
```
diegozea: Is
@simd for x in seq
    n += Int(f(x))
end
faster?

diegozea: @bicycle1885

bicycle1885: It may be. But I think the Julia compiler is not so smart to parallelize the loop because it contains a function call, which will not be inlined as far as I know.

diegozea:
Even if `@simd` doesn't work, [avoiding the branch/if](http://stackoverflow.com/questions/11227809/why-is-it-faster-to-process-a-sorted-array-than-an-unsorted-array) could give more performance, couldn't?

bicycle1885:
In this benchmark, there is no noticeable difference. I think function call is much more expensive than branching.

 ```julia
function mycount1(f, xs)
    n = 0
    for x in xs
        if f(x)
            n += 1
        end
    end
    return n
end

function mycount2(f, xs)
    n = 0
    for x in xs
        n += ifelse(f(x), 1, 0)
    end
    return n
end

let
    xs = rand(0x00:0xff, 10_000_000)
    println("--- mycount1 ---")
    mycount1(x -> x == 0x00, xs)
    @time mycount1(x -> x == 0x00, xs)
    @time mycount1(x -> x == 0x00, xs)
    @time mycount1(x -> x == 0x00, xs)

    println("--- mycount2 ---")
    mycount2(x -> x == 0x00, xs)
    @time mycount2(x -> x == 0x00, xs)
    @time mycount2(x -> x == 0x00, xs)
    @time mycount2(x -> x == 0x00, xs)
end
```

On Julia 0.5:
```
~/v/ragel (julia-fix-syntax|…) $ julia-dev test.jl
--- mycount1 ---
  0.013750 seconds (1.76 k allocations: 79.983 KB)
  0.014053 seconds (1.63 k allocations: 72.374 KB)
  0.015327 seconds (1.63 k allocations: 72.374 KB)
--- mycount2 ---
  0.015788 seconds (1.66 k allocations: 73.440 KB)
  0.015321 seconds (1.66 k allocations: 73.409 KB)
  0.017677 seconds (1.66 k allocations: 73.409 KB)
```

diegozea:
Thanks for testing it :) The actual implementation is fine then.

PR #228, opened 2016-07-26
Evolutionary distance computation for BioSequence.

Ward9250:
Fulfills some of issue #266 

This PR will start with Raw mutations counts, P distances, JC69, and K80. For BioSequences.


src/phylo/distances.jl
```diff
@@ -0,0 +1,46 @@
...
+abstract MutationModel
+
+immutable Raw <: MutationModel end
+
+immutable JukesCantor69 <: MutationModel end
+
+macro checkambig(expr)
+    quote
+        if isambiguous(a[i]) || isambiguous(b[i])
+            break
```
diegozea: Why are you using `break` instead of `throw` an error?

src/phylo/distances.jl
```diff
@@ -0,0 +1,46 @@
...
+    return mismatches(a, b) / length(a)
+end
+
+@inline function distance_pairdel(a::BioSequence, b::BioSequence, model::Raw)
+    @assert length(a) == length(b)
+    d = 0
+    l = 0
+    for i in eachindex(a)
+        @checkambig begin
+            d += a[i] != b[i] ? 1 : 0
```
diegozea: You can use `ifelse` here.

src/phylo/distances.jl
Ward9250: It should be a continue not a break, that's a typo on my part.

src/phylo/distances.jl
```diff
@@ -170,7 +232,58 @@ end
 
 # K80 Distance computation.
 
+<<<<<<< HEAD
```
kdmurray91: Whoops! @Ward9250

Ward9250: I've seen it - it's gone on my uncommitted workings 😆

kdmurray91: Sounds awesome! Excuse the uninvited interjections, for some reason the diffs display in email alerts and though I'd let you know in case you missed it! Sorry for the noise.

Ward9250: This is not problematic for me at all, in fact we encourage such participation! :)

Ward9250:
Ok I think this is ready for review.

Ward9250:
I'd like to merge soon if no objections.

src/phylo/Phylo.jl
```diff
@@ -55,6 +58,17 @@ export
...
     add_branch!,
 
+    ## Evolutionary distance computation.
+    N_Mutations,
+    P_Distance,
+    JukesCantor69,
+    Kimura80,
+    distance,
+    Raw,        # alias for N_Mutations{DifferentMutation}
+    P,          # alias for P_Distance{DifferentMutation}
```
bicycle1885: `P` and `Raw` looks too short and vague.

src/phylo/distances.jl
```diff
@@ -0,0 +1,172 @@
...
+#
+# Part of the Bio.Phylo module.
+#
+# This file is a part of BioJulia. License is MIT: https://github.com/BioJulia/Bio.jl/blob/master/LICENSE.md
+
+# Types
+# -----
+
+# Evolutionary distances
+abstract EvoDist
```
bicycle1885: If you name it `EvolutionaryDistance` you can remove the comment above 😄 . I think it is no use in abbreviating "Distance" as "Dist"; it saves four types but that's all.

src/seq/nucleotide.jl
```diff
@@ -122,6 +122,16 @@ function isambiguous(nt::Nucleotide)
 end
 
 """
+    isppambiguous(nt::Nucleotide)
+
+Test if nucleotide could be either a Purine or a Pyrimidine.
+"""
+function isppambiguous(nt::Nucleotide)
```
bicycle1885: I don't think this is so useful.

src/phylo/distances.jl
```diff
@@ -0,0 +1,172 @@
...
+abstract UncorrectedDist <: EvoDist
+abstract CorrectedDist <: EvoDist
+abstract TsTv <: CorrectedDist
+
+immutable N_Mutations{T} <: UncorrectedDist end
+immutable P_Distance{T} <: UncorrectedDist end
```
bicycle1885: Is there any references for `N_Mutations` and `P_Distance`? Also, underscore in a type name looks a little bit unusual in Julia.

src/phylo/Phylo.jl
Ward9250: Ok it can be removed.

src/phylo/distances.jl
Ward9250: Sure that sounds good!

src/seq/nucleotide.jl
Ward9250: Ah this thing again, this is an artefact from before when this PR was distances & also the mutation counting. It got removed in the mutation counting PR as well, I'll remove this.

src/phylo/distances.jl
Ward9250: Just to make sure - references in the academic literature? Ok I can change the underscores.

Ward9250:
@bicycle1885 Based on your comments I've changed N_Mutations{T} to Count{T}, and P_Distance{T} to Proportion{T}, removing the underscore, and I think, giving them both more semantic names e.g. I think the meaning of e.g. Count{Transitions} is clear, as is Proportion{Transitions}, indeed Proportion is more descriptive for people who have never heard of the measure referred to as a p-distance (Phylogeneticists and population geneticists certainly would have, but other biologists may not have).

src/phylo/distances.jl
bicycle1885: Yes, textbook or scientific paper will do. But if p-distance is very common in phylogenetics, that may not be necessary.

src/var/distances.jl
```diff
@@ -0,0 +1,172 @@
...
+#
+# Part of the Bio.Var module.
+#
+# This file is a part of BioJulia. License is MIT: https://github.com/BioJulia/Bio.jl/blob/master/LICENSE.md
+
+# Types
+# -----
+
+# Evolutionary distances
+abstract EvolutionaryDistances
```
bicycle1885: Why is this plural?

src/var/distances.jl
```diff
@@ -0,0 +1,172 @@
...
+abstract UncorrectedDist <: EvolutionaryDistances
+abstract CorrectedDist <: EvolutionaryDistances
```
bicycle1885: "Dist" should be "Distance".

src/var/distances.jl
```diff
@@ -0,0 +1,171 @@
...
+crucial difference: two-kinds of mutation are considered
+called Transitions and Transversions.
+Transitions and transversions can occur with different probabilities in this
+model, however, both transition and transversion rates/probabilities are the
+same for all sites along the DNA sequence.
+"""
+immutable Kimura80 <: TsTv end
+
+typealias JC69 JukesCantor69
+typealias K80 Kimura80
```
bicycle1885: I think these type aliases are redundant. We should stick to use a single type name to avoid unnecessary confusion.

Ward9250: JC69 and K80 and other models in phylogenetics and evo.genetics have such few letter abbreviations that are commonly used, but they can be defined elsewhere in specific places if really needed, so I took them away.

Ward9250:
@bicycle1885 tests pass now :+1: if there's nothing else I'd like to merge and tick a few things off of #266.

src/var/Var.jl
```diff
@@ -12,17 +12,28 @@ using Bio.Seq
...
-    ### Checking mutation types of nucleotides.
+    count_mutations,
+
+    # Genetic and Evolutionary distances
+    Count,
+    Proportion,
+    JukesCantor69,
+    Kimura80,
+    JC69,       # alias for JukesCantor69
+    K80,        # alias for Kimura80
```
bicycle1885: I guess these type aliases are already removed (a peculiar fact is that Julia doesn't warn you for this kind of mistakes).

Ward9250: That is indeed odd that the build didn't flag that up! :O

Ward9250: Ok just did a git commit amend, that should be gone now.

bicycle1885:
Okay, please go ahead :shipit: .

PR #229, opened 2016-07-26
Export isambiguous from Seq module.

Ward9250:
Very simple PR.

Sometimes if a user is dealing with individual nucleotides, or iterating a seq nucleotide by nucleotide, it would be useful to be able to use the isambiguous method.

### DIFF
--- a/docs/src/man/seq.md
+++ b/docs/src/man/seq.md
@@ -241,6 +241,7 @@ false
 alphabet
 gap
 iscompatible
+isambiguous
 ```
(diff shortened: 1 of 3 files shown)


bicycle1885:
👍  But I think tests are needed before merging it.

Ward9250:
Tests added for DNANucleotide, RNANucleotide, and AminoAcid.

bicycle1885:
LGTM :shipit:

PR #233, opened 2016-07-27
Re search

bicycle1885:
This pull request supersedes #143, which is far behind from the current master branch and I couldn't push a new branch to it (I don't know why).


bicycle1885:
Okay, this is now ready for review and merge. Some improvements of performance and interface may follow but the current status is not so bad. But before merging, as usual, I'd like to squash all commits.

Ward9250:
:+1:

PR #234, opened 2016-07-27
add demultiplexer

bicycle1885:
This adds demultiplexing algorithms for multiplexed sequencing. These are still unsophisticated but will work for the purpose.

You can create a barcode set from k-mers and demultiplex sequences that have a potentially disrupted barcode as a prefix:
```julia
# 4bp DNA barcodes
code = HammingCode(DNAKmer{4}["ACTG", "CATC", "TAGC"])
for record in open("seqs.fasta", FASTA)
    barcode_id = demultiplex(code, record.seq)
    # ...
end
```

To be honest, I'm not familiar with demultiplexing techniques. @kdmurray91, could you give me some comments about these interfaces and algorithms?


kdmurray91:
@bicycle1885 I'll do a review of this on the weekend. A few things to watch are that barcodes can be different lengths (e.g. in RADseq & GBS), and that there can be 1000s of barcodes per lane, so having an `O(1)` algorithm with respect to the number of barcodes would be best. Oh, and I'd recommend not using Levenshtein distance demulitplexing with barcodes shorter than 8bp, and assert that the minimum pairwise levenshtein distance between barcodes is at least 1 greater than the allowable levenshtein distance.

bicycle1885:
Thank you, @kdmurray91 for valuable advice!

Are the barcodes with different lengths the result of sequence errors (indels) or by design (specific to RADseq and GBS)?

kdmurray91:
@bicycle1885 The differing lengths are due to the fragments being sequences starting with the restriction site. If all bases are the same across all clusters, Illumina sequencing doesn't work very well (especially if this happens at the start of the reads). Therefore the base composition of each cycle is balanced using longer, variable length barcode sequences (usually 4-8bp).

Another thing to mention is the need for combinatorial barcoding, where you have e.g. 12 R1 and 8 R2 barcodes that together give 96 samples per lane.

bicycle1885:
I guess I got it. So you use a trie tree in axe to demultiplex variable-length barcodes because it needs prefix searching.

I naively think combinatorial barcoding can be handled using a pair of demultiplexers; one for R1 barcodes and the other for R2.

I added a static double-array trie for prefix searching. It is not yet used but I will make demultiplexing algorithms base on it soon.

kdmurray91:
Another thing to consider: the trie will need to store 'N' chars, as these are an allowable "error" in NGS data. I'll do a full review later today hopefully.

bicycle1885:
@kdmurray91 Following your advice, I noticed major revise would be needed to this pull request. So, I think your full review is not needed right now. I'll update it today and tomorrow, then please review it after that.

kdmurray91:
@bicycle1885 Thanks for letting me know, and sorry for not remembering about `N`s earlier.

bicycle1885:
I've significantly changed the algorithm and interfaces of the demultiplexer and moved the code from `Bio.Align` to `Bio.Seq`.

Now it basically exports a type and a function: `Demultiplexer` and `demultiplex`. `Demultiplexer` takes a list of DNA barcodes with some optional arguments and `demultiplex` takes an instance of `Demultiplexer` and a DNA sequence to identify a barcode. The test cases and docstrings would be helpful to understand what I intend.

So, @kdmurray91, please check this pull request when you have time 😄 .

kdmurray91:
ACK, I'm having a rather chaotic week, so I might not be super speedy, but I'll look at the diff in parts.

src/seq/demultiplexer.jl
```diff
@@ -0,0 +1,336 @@
...
+
+# Arguments
+* `barcodes`: the sorted list of DNA barcodes; integer IDs are assigned in this order.
+* `n_max_errors=1`: the number of maximum correctable errors.
+* `distance=:hamming`: the distance metric (`:hamming` or `:levenshtein`).
+"""
+function Demultiplexer(barcodes::Vector{DNASequence};
+                       n_max_errors::Integer=1,
+                       distance::Symbol=:hamming)
+    if !issorted(barcodes)
```
kdmurray91: Can't we just sort them in place?

src/seq/demultiplexer.jl
```diff
@@ -0,0 +1,336 @@
...
+                error("barcode must be umambiguous")
+            end
+        end
+    end
+
+    # build trie trees for each error level (0-n_max_errors)
+    tries = BarcodeTrie[]
+    for m in 0:n_max_errors
+        # generate "erroneous" barcodes
+        barcodes′ = DNASequence[]
```
kdmurray91: personal pet peeve, but I'd like to `s/barcodes'/mutated_barcodes/` for clarity. Feel free to ignore this

src/seq/demultiplexer.jl
```diff
@@ -0,0 +1,336 @@
...
+        if i != 0
+            return i
+        end
+    end
+    if !linear_search_fallback
+        # not found
+        return 0
+    end
+
+    # find the best matching barcode using linear search
```
kdmurray91: Excuse the ignorant question, but under which circumstances do we want to perform this linear search?

src/seq/demultiplexer.jl
```diff
@@ -0,0 +1,336 @@
...
+
+"""
+    demultiplex(demultiplexer::Demultiplexer, seq::Sequence, linear_search_fallback::Bool=false)
+
+Return a barcode index that matches `seq` with least errors.
+
+The upper limit of the number of maximum errors is bounded by the `n_max_errors`
+parameter of `demultiplexer`. When `linear_search_fallback` is `true`, this
+function tries to find the best matching barcode using linear search.
+"""
```
kdmurray91: Perhaps this should return a tuple of `(i, mismatches)`, where mismatches is either `dist_min` or the index of the trie in which the match was found (minus 1).

Or even create a `BarcodeMatch` type that keeps track of the barcode, the number of mismatches, the length and any extra metadata (if any exists).

kdmurray91:
@bicycle1885 Looks awesome, thanks heaps for implementing this. I'm going to write a tiny CLI wrapper for this and test it against `axe`.

src/seq/demultiplexer.jl
bicycle1885: Integer IDs are assigned to barcodes in this order. If we sort it in place, the IDs are intermixed and the user cannot know which is which. But, yes, I can make it more user-friendly by keeping the original order of barcodes. I will do that later.

src/seq/demultiplexer.jl
bicycle1885: I thought users may want to somehow assign a barcode for each sequence even when the barcode is too erroneous to identify the original one. This linear search is off by default. I now think I need to check whether this linear search is better than random assignment.

src/seq/demultiplexer.jl
bicycle1885: I agree with you. The number of mismatches would be valuable information sometimes. I think returning a pair of barcode ID and the number of mismatches would achieve a good balance between functionality and simplicity.

bicycle1885:
Thank you @kdmurray91 . I will update this pull request ASAP (maybe tomorrow) to address your comments.

src/seq/demultiplexer.jl
kdmurray91: Ah, of course. Sounds good.

src/seq/demultiplexer.jl
kdmurray91: Ah, it probably is to a point, but clearly if you have `NNNN` in the read, this becomes random assignment (well arbitrary assignment at least). Provided that we return the number of mismatches, then I say it's fine. It may be worth capping the maximum distance at say 1/2 the barcode length or something, beyond which the linear search assignment won't reflect the most likely original barcode as they're too different.

src/seq/demultiplexer.jl
kdmurray91: :+1:

src/seq/demultiplexer.jl
bicycle1885: Thinking about it more carefully, I found there is no need to check the order of barcodes. So I removed it.

src/seq/demultiplexer.jl
bicycle1885: done.

src/seq/demultiplexer.jl
bicycle1885: Now `demultiplex` returns the distance along with the assigned barcode ID, so the user can set a cutoff threshold when needed. I also changed the algorithm to select the barcode to return: select the minimum distance barcodes and return one of them at random. I believe this can avoid skewed assignments when there are so many erroneous sequences.

bicycle1885:
@kdmurray91 I'd like to merge this one if you say OK.

kdmurray91:
go for it @bicycle1885! I have half-finished an implementation of Axe in julia with this PR, which I will use to test (both this and Axe itself) more thoroughly.

bicycle1885:
Thank you!

bicycle1885:
Ah, @kdmurray91, please use [merge squashing](https://help.github.com/articles/about-pull-request-merge-squashing/) when you merge a pull request next time, which squashes all temporary commits into a single commit and tidies up the commit log.

kdmurray91:
@bicycle1885 Ah crap, I though someone had told github to automatically do so.

bicycle1885:
I found a setting to force merge squashing, so I turned it on.

kdmurray91:
Cheers!

PR #235, opened 2016-07-28
ispurine and ispyrimidine for nucleotides

Ward9250:
Implements ispurine and ispyrimiadine for nucleotide types.

These methods return true if a nucleotide is surely a purine or is surely a pyrimidine. 
Where by "surely" we mean the base of the nucleotide is known with no ambiguity, or the ambiguity of the base does not change the certainty of whether the base is purine or pyrimadine.


Ward9250:
This is done and RTR. @bicycle1885 Do you think we could increase performance for getting pyrimidines and purines in BioSequences by doing some clever bit ops on whole chunks of sequence at a time rather than getting each nucleotide and testing it individually? Signature would look something like `isPurine(seq::BioSequence)` and would return bit or logical vectors indicating which nucleotides are purine.

src/seq/nucleotide.jl
```diff
@@ -103,6 +103,24 @@ function isGC(nt::DNANucleotide)
 end
 
 """
+    isPurine(nt::Nucleotide)
+
+Test if nucleotide is surely a Purine.
+"""
+function isPurine(nt::Nucleotide)
```
bicycle1885: camelCase is not the Julia naming convention. It should be `ispurine`.

src/seq/nucleotide.jl
```diff
@@ -103,6 +103,24 @@ function isGC(nt::DNANucleotide)
...
+    return (compatbits(nt) | 0b0101) == 0b0101
```
bicycle1885: This is wrong. When `nt = DNA_Gap` `compatbits(nt)` is `0b0000` and hence `(compatbits(nt) | 0b0101) = 0b0101`. And the result is `isPurine(DNA_Gap)` is `true`.

src/seq/nucleotide.jl
```diff
@@ -103,6 +103,24 @@ function isGC(nt::DNANucleotide)
...
+end
+
+"""
+    isPyrimidine(nt::Nucleotide)
+
+Test if nucleotide is surely a Pyrimidine.
+"""
+function isPyrimidine(nt::Nucleotide)
```
bicycle1885: `ispyrimidine`

src/seq/nucleotide.jl
Ward9250: Can this be done with bits operations another way, or do we have to resort to a `nt == DNA_A || nt == DNA_G || nt == DNA_R`, and create seperate method defs out for DNANucleotide and RNANucleotide?

bicycle1885: Also note that `DNA_R`, an ambiguous DNA nucleotide, is surely a purine because R = A or G.

bicycle1885: Also note that `DNA_R`, an ambiguous DNA nucleotide, is surely a purine because R = A or G.

Ward9250: Yes this is what I was trying to achieve, A, G and R, should return true, all others should be false. I believe currently A G and R do return true, and the others do return false - except gaps which also return true.

bicycle1885: I think whay you intended is: `(compatbits(nt) & 0b0101) != 0x00`.

src/seq/Seq.jl
```diff
@@ -46,6 +46,8 @@ export
     RNA_Gap,
     iscompatible,
     isambiguous,
+    isPurine,
+    isPyrimidine,
```
kdmurray91: ITYM `ispurine`? i.e. lower case P per the function definitions?

diegozea: I'm in favor if ispurine (lowercase P).

Ward9250:
Awesome thanks guys, acted on all review issues, now waiting to see if tests pass.

test/seq/runtests.jl
```diff
@@ -307,11 +307,29 @@ end
...
         end
         for x in alphabet(RNANucleotide)
-            @test isambiguous(x) == (x > RNA_U)
+            @test isambiguous(nt) == (nt > RNA_U)
+        end
+    end
+
+    @testset "isPurine" begin
+        for nt in alphabet(DNANucleotide)
+            @test isPurine(nt) == (nt == DNA_A || nt == DNA_G)
```
bicycle1885: You didn't update names in tests :(

Ward9250: Oh my this morning is not my morning!

src/seq/nucleotide.jl
Ward9250: That still does not work, it stops gaps returning true, but DNA_M and others return true when only DNA_A, DNA_G, and DNA_R should return true, so currently any ambiguity code that could be A, and any code that could be G returns true. We only want, A, G, or A|G to return true.

bicycle1885: Sorry, I was careless. The following snippet should work:
```jl
function ispurine(nt::Nucleotide)
    bits = compatbits(nt)
    return ((bits & 0b1010) == 0) && ((bits & 0b0101) != 0)
end
```

PR #236, opened 2016-07-29
follow the Libz change

bicycle1885:


bicycle1885:
CI failure on OS X is due to timeout after tests. I think it is not related to this pull request.

PR #237, opened 2016-07-29
Backports for 0.2

bicycle1885:
This backports a few commits to the release-0.2 branch. After merging, I will tag and release v0.2.1 if no objections.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -5,6 +5,6 @@ Compat
 IndexableBitVectors 0.1.1
 IntervalTrees
 Iterators
-Libz 0.1.1
+Libz 0.2
 LightGraphs
 LightXML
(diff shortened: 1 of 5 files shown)


Ward9250:
Is there an ETA on when julia becomes 0.5 release? I'm just wondering if the distance stuff I'm working on should go into master or on the 0.2 branch too.

bicycle1885:
I think no ETA but I expect 0.5 will be released next month or the next. You can check the remaining steps here: https://github.com/JuliaLang/julia/issues/17418. I don't think it is a good idea to add new features to the 0.2 branch because it is intended for backward-compatible bug fixes and usability issues. Bio.jl 0.3 would be the best timing to include what you're working now on. In my opinion, 0.3 with various new features should be released just after the official Julia 0.5 release and we should drop the support of Julia 0.4 because it is a heavy burden. [HTSFileFormats.jl](https://github.com/bicycle1885/HTSFileFormats.jl) is supposed to be included in 0.3, too.

PR #238, opened 2016-07-29
FASTA Index

### LINKED ISSUE #177: .fai (FASTA index) support
(memorandum)

Supporting the .fai file format would be useful to index large a FASTA file and load arbitrary sequences from it. The format is easy to parse (http://www.htslib.org/doc/faidx.html) but seeking a file with `Base.seek` requires some modifications of BufferedStreams.jl and the FASTA parser.

bicycle1885:
This will close #177.

### DIFF
--- a/src/seq/fai.jl
+++ b/src/seq/fai.jl
@@ -1,0 +1,70 @@
+# FAI
+# ===
+#
+# Index for random access to FASTA files.
+#
+# This file is a part of BioJulia.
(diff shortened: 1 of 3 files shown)


bicycle1885:
Now ready to merge 😃

Ward9250:
:+1:

Ward9250:
Maybe an entry in the manual, but happy if that goes in here or in another PR, depends what you'd like to do :)

bicycle1885:
I will write docs along with 2bit file format soon.

PR #240, opened 2016-07-30
Io docs

bicycle1885:
Rewrite of IO docs, still work-in-progress.


bicycle1885:
I added some details for each file format and addressed random access indexing there (https://github.com/BioJulia/Bio.jl/pull/238#issuecomment-236364704).

Ward9250:
:+1: LGTM

PR #243, opened 2016-07-31
Fix fastq quality encoding inference

### LINKED ISSUE #242: FASTQ parser infers different quality encodings for every record
When parsing a FASTQ file, for each record,`infer_quality_encoding()` picks the first encoding in `preferred_quality_encodings` that is compatible with the quality string range. So two records from the same FASTQ file can have different inferred quality encodings. This means that most FASTQ files will be parsed incorrectly.

Moreover, trying to fix the problem by specifying a quality encoding, as in

`open(filename, FASTQ, quality_encoding=Seq.SANGER_QUALITY_ENCODING)`

does not work, because `infer_quality_encoding()` ignores the `encodings` argument by overwriting it with `ALL_QUALITY_ENCODINGS`.

kemaleren:
Closes #242.

This branch fixes the problem by requiring that exactly one quality encoding is provided to the FASTQParser, so no inference is done to determine the quality encoding.

An alternative solution would infer the quality encoding for all sequences in the file at once, but then they could not be parsed record by record.

### DIFF
--- a/src/precompile.jl
+++ b/src/precompile.jl
@@ -6,11 +6,11 @@
 
 if VERSION < v"0.5-"
     precompile(Base.open, (ASCIIString, Type{Seq.FASTA},))
-    precompile(Base.open, (ASCIIString, Type{Seq.FASTQ},))
+    precompile(Base.open, (ASCIIString, Type{Seq.FASTQ}, Type{Seq.QualityEncoding},))
     precompile(Base.open, (ASCIIString, Type{Intervals.BED},))
 else
     precompile(Base.open, (String, Type{Seq.FASTA},))
-    precompile(Base.open, (String, Type{Seq.FASTQ},))
+    precompile(Base.open, (String, Type{Seq.FASTQ}, Type{Seq.QualityEncoding},))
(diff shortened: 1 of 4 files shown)


kemaleren:
Tests fail because I still need to update `tests/seq/runtests.jl`.

Ward9250:
This looks good to me @kemaleren, so long as tests get updated and pass I'm happy with this, we have a policy of getting two reads of a PR but I expect @bicycle1885 will have a look as well before too long. Thanks for fixing this! :+1:

kemaleren:
Thanks @Ward9250. One caveat: this PR will break any code relying on opening files like `open(filename, FASTQ)`. However, the caller really should be explicit about which encoding to use. I bet a lot of code out there is silently using the wrong encoding.

bicycle1885:
I agree. The user should specify the quality encoding explicitly.

Ward9250:
I agree, and any application or script created should be able to specify an encoding with a flag (and probably assume one by default), so I don't think the lack of inference is a big deal.

Ward9250:
Tests pass for PR - I think Appveyor fails due to faix and not these changes.

PR #246, opened 2016-08-05
make unexported docstrings comments

bicycle1885:


bicycle1885:
CI fails on Windows (`unlink: resource busy or locked (EBUSY)`). I'm not sure why but I believe this is not related to this change.

bicycle1885:
I will merge this soon. This is just a small stylistic improvement.

PR #251, opened 2016-08-08
add random sequence generators

bicycle1885:
This adds random sequence generators for DNA, RNA, and amino acids. At the moment, the only available generator is `StationaryGenerator`, but the interface is extensible so that we can add more elaborate generators like HMM or RNN.

### DIFF
--- a/src/seq/randseq.jl
+++ b/src/seq/randseq.jl
@@ -52,11 +52,13 @@ function randseq{T}(generator::StationaryGenerator{T}, len::Integer)
 end
 
 const StationaryDNAGenerator = StationaryGenerator(
-    [DNA_A, DNA_C, DNA_G, DNA_T], [0.25, 0.25, 0.25])
+    collect(dna"ACGT"), ones(3) / 4)
+
 const StationaryRNAGenerator = StationaryGenerator(
-    [RNA_A, RNA_C, RNA_G, RNA_U], [0.25, 0.25, 0.25])
+    collect(rna"ACGU"), ones(3) / 4)
+
 const StationaryAAGenerator = StationaryGenerator(
-    alphabet(AminoAcid)[1:20], ones(19) / 20)
+    collect(aa"ARNDCQEGHILKMFPSTWYV"), ones(19) / 20)
 
 """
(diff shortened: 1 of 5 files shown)


bicycle1885:
I will merge this tomorrow if no objections.

test/seq/runtests.jl
```diff
@@ -10,6 +10,7 @@ end
 import Compat
 using Bio.Seq,
     BufferedStreams,
+    StatsBase,
```
kdmurray91: Forgive me if I'm wrong, but I can't see where  `StatsBase` is still used? I assume this was included for the now-commented out `cdf(Binomial(...))` calls?

src/seq/randseq.jl
```diff
@@ -0,0 +1,88 @@
...
+function StationaryGenerator{T,U<:AbstractFloat}(
+        elems::AbstractVector{T}, probs::AbstractVector{U})
+    return StationaryGenerator{T}(elems, probs)
+end
+
+alphatype(::Type{DNANucleotide}) = DNAAlphabet{4}
+alphatype(::Type{RNANucleotide}) = RNAAlphabet{4}
+alphatype(::Type{AminoAcid})     = AminoAcidAlphabet
+
+function randseq{T}(generator::StationaryGenerator{T}, len::Integer)
```
kdmurray91: Is it worth exporting this function and some of the above types? Would that make it any easier for user code to create a subtype of the `SequenceGenerator` type and use `randseq` with such a type? Or could one use `Bio.Seq.randseq{T}(gen::MyNewStationaryGenerator{T}, ...)` to the same effect?

test/seq/runtests.jl
bicycle1885: `countmap` is imported from StatsBase.

src/seq/randseq.jl
bicycle1885: I'm not sure about the demand of random sequence generator, so I thought it would be better to start it as a minimum viable utility. `SequenceGenerator` may be worth exporting so that users can define their own generators more easily.

PR #253, opened 2016-08-10
Add labelled square matrix parser

kdmurray91:
As talked about on gitter, this PR adds a parser for the common LSM format used to store e.g. distance matrices. The parser is currently a little ad-hoc, and adapted from @bicycle1885's substitution matrix parsers.

I will also attach a commit that removes that parser, and uses the `readlsm` function included here.

This is ready to review

### DIFF
--- a/src/util/labelledmatrix.jl
+++ b/src/util/labelledmatrix.jl
@@ -2,11 +2,7 @@ export readlsm
 
 const _default_delims = [' ','\t']
 
-"""
-Reads a Labelled Square Matrix
-
-Returns a tuple of `(matrix, labels)`
-"""
+"Reads a Labelled Square Matrix, returning a tuple of `(matrix, labels)`"
 function readlsm{T}(::Type{T}, io::IO, delim=_default_delims, comment='#')
     matrix = Matrix{T}()
     labels = UTF8String[]
--- a/src/util/labelledmatrix.jl
+++ b/src/util/labelledmatrix.jl
@@ -1,5 +1,3 @@
-module LabelledSquareMatrices
-
 export readlsm
 
 """
@@ -56,5 +54,3 @@ end
 function readlsm(path::AbstractString, delim='\t')
     return readlsm(Float64, path, delim)
 end
-
-end # module LabelledSquareMatrices
--- a/src/util/labelledmatrix.jl
+++ b/src/util/labelledmatrix.jl
@@ -42,13 +42,19 @@ function readlsm{T}(::Type{T}, io::IO, delim='\t', comment='#')
 end
 
 function readlsm(io::IO, delim='\t')
-    readlsm(Float64, io, delim)
+    return readlsm(Float64, io, delim)
+end
+
+function readlsm{T}(::Type{T}, path::AbstractString, delim='\t')
+    lsm = nothing
+    open(path) do file
+        lsm = readlsm(T, file, delim)
+    end
+    return lsm
 end
 
 function readlsm(path::AbstractString, delim='\t')
-    open(path) do file
-        lsm = readlsm(Float64, file, delim)
-    end
+    return readlsm(Float64, path, delim)
 end
 
 end # module LabelledSquareMatrices
--- a/src/util/labelledmatrix.jl
+++ b/src/util/labelledmatrix.jl
@@ -1,17 +1,19 @@
 export readlsm
 
+const _default_delims = [' ','\t']
+
 """
 Reads a Labelled Square Matrix
 
 Returns a tuple of `(matrix, labels)`
 """
-function readlsm{T}(::Type{T}, io::IO, delim='\t', comment='#')
+function readlsm{T}(::Type{T}, io::IO, delim=_default_delims, comment='#')
     matrix = Matrix{T}()
     labels = UTF8String[]
     nitems = 0
     row = 0
     for line in eachline(io)
-        line = chomp(line)
+        line = rstrip(line)
         if startswith(line, comment) || isempty(line)
             continue  # skip comments and empty lines
         end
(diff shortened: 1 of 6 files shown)


bicycle1885:
Nice job! I will take a look at it later more carefully. Just a comment: I think it is a kind of overkill to create a new module just for a single function `readlsm`.

kdmurray91:
Cheers. I couldn’t work out where to stick the function, so I just made a new module. I agree that it's a little silly, but where should this function go?

bicycle1885:
Maybe in `Bio.Align`? I'd like to use it for parsing substitution matrices in `Bio.Align`.

kdmurray91:
@bicycle1885 How about a new `Bio.Stats` or such, for general bio-statistical functionality? It's probably moot, but this is useful for *many* more things than subst. matrices, so `Bio.Align` makes less sense to me. What do you think?

kdmurray91:
Test failures are due to the absence of the test files in BioFmtSpeciments master (see BioJulia/BioFmtSpecimens#13)

Ward9250:
:+1: @kdmurray91 This is also useful for phylogenetics, so is definitely a general utility for biology.

bicycle1885:
At the moment, I cannot say 👍  to create a `Bio.Stats` module because we cannot make it a full-fledged module. If `Bio.Align` is not a good place, why don't we create a module like `Bio.Utils` and push things under src/util/ in it? I don't like modules named utils so much but we still need it because we sometimes have to place miscellaneous functions like this.

kdmurray91:
@bicycle1885 I agree that `Bio.Utils` is probably the best of many silly places. I will make this change then push, and it should be ready to merge (along with the respective BioFmtSpecimens PR)

bicycle1885:
Fun bikeshedding: `Bio.Util` or `Bio.Utils`?

kdmurray91:
I just went with `Bio.Util` as that seemed to match other submodules (and the existing dir name).

Ward9250:
:+1:

kdmurray91:
Latest commit fixes an issue parsing matrices with multiple spaces as delimiters.

kdmurray91:
I can't work out how to re-run appveyor, but I re-ran travis CI after merging the BioFmtSpecimens PR and the tests now pass.

src/util/labelledmatrix.jl
```diff
@@ -0,0 +1,57 @@
+export readlsm
+
+const _default_delims = [' ','\t']
+
+"""
+Reads a Labelled Square Matrix
+
+Returns a tuple of `(matrix, labels)`
+"""
```
bicycle1885: Docstrings in Bio.jl should follow the [guideline](http://julia.readthedocs.io/en/latest/manual/documentation/#documentation) of Julia.

test/util/labelledmatrix.jl
```diff
@@ -0,0 +1,131 @@
...
+end
+
+import Bio.Util: readlsm
+
+integral_matrix = """\
+\tA\tB\tC
+A\t1\t2\t4
+B\t2\t1\t2
+C\t4\t2\t1
+"""
```
bicycle1885: Put these global variables inside a `@testset` macro. I think values used in a test should be placed near the test.

test/util/labelledmatrix.jl
```diff
@@ -0,0 +1,131 @@
...
+                 ]
+all_subst_matrices = ["BLOSUM45", "BLOSUM50", "BLOSUM62", "BLOSUM80",
+                      "BLOSUM90", "NUC.4.4", "PAM250", "PAM30", "PAM70"]
+
+@testset "LabelledSquareMatrices" begin
+    @testset "types" for T in [Int64, Float64]
+        m, l = readlsm(T, IOBuffer(integral_matrix))
+        expt_m = Array{T}(integral_matrix_expt)
+        @test m == expt_m
+        @test l == UTF8String["A", "B", "C"]
```
bicycle1885: `@test eltype(m) == T` is needed to check the elements type of a matrix.

src/util/labelledmatrix.jl
Ward9250: If in doubt, follow the example set in the seq module :)

kdmurray91:
Think I've addressed all comments. I hope I got the docstring right, there was some variation amongst the docstring style in the Seq module.

bicycle1885:
The standard style of docstrings I meant is:
> Always show the signature of a function at the top of the documentation, with a four-space indent so that it is printed as Julia code.

> This can be identical to the signature present in the Julia code (like mean(x::AbstractArray)), or a simplified form. Optional arguments should be represented with their default values (i.e. f(x, y=1)) when possible, following the actual Julia syntax. Optional arguments which do not have a default value should be put in brackets (i.e. f(x[, y]) and f(x[, y[, z]])). An alternative solution is to use several lines: one without optional arguments, the other(s) with them. This solution can also be used to document several related methods of a given function. When a function accepts many keyword arguments, only include a <keyword arguments> placeholder in the signature (i.e. f(x; <keyword arguments>)), and give the complete list under an # Arguments section (see point 4 below).

> Include a single one-line sentence describing what the function does or what the object represents after the simplified signature block. If needed, provide more details in a second paragraph, after a blank line.

> The one-line sentence should use the imperative form (“Do this”, “Return that”) instead of the third person (do not write “Returns the length...”) when documenting functions. It should end with a period. If the meaning of a function cannot be summarized easily, splitting it into separate composable parts could be beneficial (this should not be taken as an absolute requirement for every single case though).
http://julia.readthedocs.io/en/latest/manual/documentation/

So for exmaple, I would write the docstring of `readlsm` like this way:
```julia
"""
    readlsm(path, delim=[' ', '\t'])
    readlsm(type, path, delim=[' ', '\t'])

Read a labelled square matrix file from `path` and return a pair of the matrix and the labels.

<longer description here if needed>

<examples here if needed>
"""
function readlsm(...)
```

kdmurray91:
@bicycle1885 Fixed in the latest commit. Sorry for being a bit dim about this, I have many years of Python experience to un-learn!

bicycle1885:
This looks good to me :shipit:. I will merge this soon if no further comments.

kdmurray91:
Thanks heaps Kenta!

PR #254, opened 2016-08-10
Added U for ambiguous nucleotides in Seq docs

kescobo:
Nothing complicated, just a small change to docs.


bicycle1885:
Thank you a lot!

PR #255, opened 2016-08-12
New features for Bio.Structure

### LINKED ISSUE #151: Improvements to Bio.Structure
`Bio.Structure` is now merged and working. I'm opening this issue to act as a discussion point for future development of the module.

I plan to add a few new features which I think should go in the core module:
- Get sequence from the residues in a PDB file (which will link with the other BioJulia stuff).
- Calculate bond angles/dihedrals/Ramachandran plots.
- Calculate contact maps.

And also to do some stuff on performance:
- Benchmark the module against similar packages.
- Speed up particular parts, especially the stage of parsing that goes `Vector{Atom}` to `ProteinStructure`.

If anyone has anything else they would like to see, chime in here. For example:
- Parsing of the PDB header, which I can imagine being useful though I have not yet found essential.
- Calculation of protein mass and centre of mass - would need more stringent element parsing.
- Superposition, which I would like to implement long-term but which requires sequence alignments to be integrated in first.

jgreener64:
This is a PR to improve the `Bio.Structure` module. Since I wrote it I have learned more about Julia, and thought a bit more about the implementation. There are some changes to the type system but the user-facing functions are all the same.

Some changes are implemented now, but when complete this PR will add:
- Significantly faster file parsing and much less memory allocation.
- Calculation of bond angles/dihedrals/Ramachandran plots.
- Calculation of contact maps.
- Getting sequence from the residues in a PDB file.
- More complete docs.

Still work and testing needed. Closes #151.

### DIFF
--- a/docs/src/man/structure.md
+++ b/docs/src/man/structure.md
@@ -27,6 +27,7 @@ To parse a PDB file into a Structure-Model-Chain-Residue-Atom framework:
 
 ```julia
 julia> struc = read(filepath_1EN2, PDB)
+Bio.Structure.ProteinStructure
 Name                        -  1EN2.pdb
 Number of models            -  1
 Chain(s)                    -  A
@@ -50,7 +51,7 @@ The elements of `struc` can be accessed as follows:
 | `struc['A'][50]`            | Shortcut to above if it is not a hetero residue and the insertion code is blank | `AbstractResidue` |
 | `struc['A']["H_90"]`        | Model 1, chain A, hetero residue 90                                             | `AbstractResidue` |
 | `struc['A'][50]["CA"]`      | Model 1, chain A, residue 50, atom name CA                                      | `AbstractAtom`    |
-| `struc['A'][15]["CG"]['A']` | For disordered atoms, access a specific location                                | `DisorderedAtom`  |
+| `struc['A'][15]["CG"]['A']` | For disordered atoms, access a specific location                                | `Atom`            |
 
 Disordered atoms are stored in a `DisorderedAtom` container but calls fall back to the default atom, so disorder can be ignored if you are not interested in it.
 
@@ -120,7 +121,7 @@ end
 
 Models are ordered numerically; chains are ordered by character, except the empty chain is last; residues are ordered by residue number and insertion code with hetero residues after standard residues; atoms are ordered by atom serial.
 
-`collect` can be used to get lists of sub-elements. `collectatoms`, `collectresidues`, `collectchains` and `collectmodels` return lists of a particular type from a structural element or element list.
+`collect` can be used to get arrays of sub-elements. `collectatoms`, `collectresidues`, `collectchains` and `collectmodels` return arrays of a particular type from a structural element or element array.
 
 Selectors are functions passed as additional arguments to these functions. Only elements that return `true` when passed to the selector are retained.
 
@@ -166,7 +167,7 @@ julia> countresidues(struc, standardselector)
 85
 ```
 
-The sequence of a protein can be retrieved by passing a `Chain` or list of residues to `AminoAcidSequence`:
+The sequence of a protein can be retrieved by passing a `Chain` or array of residues to `AminoAcidSequence`:
 
 ```julia
 julia> AminoAcidSequence(struc['A'], standardselector)
@@ -215,7 +216,7 @@ julia> distance(struc['A'][10], struc['A'][20])
 10.782158874733762
 
 julia> rad2deg(bondangle(struc['A'][50]["N"], struc['A'][50]["CA"], struc['A'][50]["C"]))
-69.22234153916602
+110.77765846083398
 
 julia> rad2deg(dihedralangle(struc['A'][50]["N"], struc['A'][50]["CA"], struc['A'][50]["C"], struc['A'][51]["N"]))
 -177.38288114072924
@@ -241,23 +242,31 @@ plot(x=resnumber.(calphas),
      Geom.line)
 ```
 
-**B)** To find all C-alpha atoms within 5 Angstroms of residue 38:
+**B)** To print the PDB records for all C-alpha atoms within 5 Angstroms of residue 38:
 
 ```julia
 for at in calphas
     if distance(struc['A'][38], at) < 5.0 && resnumber(at) != 38
-        show(at)
+        println(join(pdbline(at)))
     end
 end
 ```
 
-**D)** To show the contact map of a structure, if you have Hinton installed:
+**D)** To view the contact map of a structure:
 
 ```julia
-using Hinton
 cbetas = collectatoms(struc, cbetaselector)
 contacts = contactmap(cbetas, 7.0)
-println(hintontxt(contacts))
+for i in 1:length(cbetas)
+    for j in 1:length(cbetas)
+        if contacts[i,j]
+            print("█")
+        else
+            print(" ")
+        end
+    end
+    println()
+end
 ```
 
 `cbetaselector` selects C-beta atoms, or C-alpha atoms for glycine residues. `contactmap` can also be given two structural elements as arguments, in which case a non-symmetrical 2D array is returned showing contacts between the elements.
--- a/docs/src/man/structure.md
+++ b/docs/src/man/structure.md
@@ -12,7 +12,7 @@ DocTestSetup = quote
 end
 ```
 
-The `Bio.Structure` module provides functionality to read [Protein Data Bank](http://www.rcsb.org/pdb/home/home.do) (PDB) files and manipulate macromolecular structures. It is designed to be used as a platform on which others can build to create tools for macromolecular structure analysis.
+The `Bio.Structure` module provides functionality to manipulate macromolecular structures, and in particular to read and write [Protein Data Bank](http://www.rcsb.org/pdb/home/home.do) (PDB) files. It is designed to be used for standard structural analysis tasks, as well as acting as a platform on which others can build to create more specific tools.
 
 
 ## Parsing PDB files
@@ -219,7 +219,7 @@ PDB format files can be written:
 writepdb("1EN2_out.pdb", struc)
 ```
 
-Any element type can be given as input. Selectors can also be given as additional arguments:
+Any element type can be given as input. Atom selectors can also be given as additional arguments:
 
 ```julia
 writepdb("1EN2_out.pdb", struc, backboneselector)
--- a/docs/src/man/structure.md
+++ b/docs/src/man/structure.md
@@ -132,7 +132,7 @@ Selectors are functions passed as additional arguments to `collectresidues` and 
 | `collectatoms(struc, calphaselector)`                   | Collect the C-alpha atoms of an element                           | `Array{AbstractAtom,1}`    |
 | `collectatoms(struc, calphaselector, disorderselector)` | Collect the disordered C-alpha atoms of an element                | `Array{AbstractAtom,1}`    |
 
-It is easy to define your own selector. The below will collect all atoms with x coordinate less than 0:
+It is easy to define your own atom or residue selectors. The below will collect all atoms with x coordinate less than 0:
 
 ```julia
 xselector(atom::AbstractAtom) = x(atom) < 0
@@ -143,22 +143,6 @@ Alternatively, you can use an anonymous function:
 
 ```julia
 collectatoms(struc, atom -> x(atom) < 0)
-```
-
-Selectors can also be passed to many other functions, including `read`:
-
-```julia
-julia> calphas = read(filepath_1EN2, PDB, calphaselector)
-Name                        -  1EN2.pdb
-Number of models            -  1
-Chain(s)                    -  A
-Number of residues          -  85
-Number of point mutations   -  5
-Number of other molecules   -  0
-Number of water molecules   -  0
-Number of atoms             -  85
-Number of hydrogens         -  0
-Number of disordered atoms  -  2
 ```
 
 `countmodels`, `countchains`, `countresidues` and `countatoms` can be used to count elements. For example:
--- a/docs/src/man/structure.md
+++ b/docs/src/man/structure.md
@@ -77,6 +77,7 @@ Properties can be retrieved as follows:
 | `residue`          | Residue an atom belongs to                                    | `Residue`                       |
 | `ishetero`         | `true` if the residue or atom is a hetero residue/atom        | `Bool`                          |
 | `isdisorderedatom` | `true` if the atom is disordered                              | `Bool`                          |
+| `pdbline`          | PDB ATOM/HETATM record for an atom                            | `String`                        |
 | `resname`          | Residue name of a residue or atom                             | `String`                        |
 | `resnumber`        | Residue number of a residue or atom                           | `Int`                           |
 | `inscode`          | Insertion code of a residue or atom                           | `Char`                          |
@@ -247,7 +248,7 @@ plot(x=resnumber.(calphas),
 ```julia
 for at in calphas
     if distance(struc['A'][38], at) < 5.0 && resnumber(at) != 38
-        println(join(pdbline(at)))
+        println(pdbline(at))
     end
 end
 ```
--- a/docs/src/man/structure.md
+++ b/docs/src/man/structure.md
@@ -70,7 +70,7 @@ Properties can be retrieved as follows:
 | `z`                | z coordinate of an atom                                       | `Float64`                       |
 | `coords`           | coordinates of an atom                                        | `Array{Float64,1}`              |
 | `occupancy`        | Occupancy of an atom (default is `1.0`)                       | `Float64`                       |
-| `tempfac`          | Temperature factor of an atom (default is `0.0`)              | `Float64`                       |
+| `tempfactor`       | Temperature factor of an atom (default is `0.0`)              | `Float64`                       |
 | `element`          | Element of an atom (default is `"  "`)                        | `String`                        |
 | `charge`           | Charge of an atom (default is `"  "`)                         | `String`                        |
 | `residue`          | Residue an atom belongs to                                    | `Residue`                       |
@@ -227,7 +227,7 @@ A few examples of `Bio.Structure` usage are given below.
 using Gadfly
 calphas = collectatoms(struc, calphaselector)
 plot(x=resnumber.(calphas),
-    y=tempfac.(calphas),
+    y=tempfactor.(calphas),
     Guide.xlabel("Residue number"),
     Guide.ylabel("Temperature factor"),
     Geom.line)
--- a/docs/src/man/structure.md
+++ b/docs/src/man/structure.md
@@ -41,17 +41,16 @@ Number of disordered atoms  -  27
 
 The elements of `struc` can be accessed as follows:
 
-| Command                     | Returns                                                                         | Return type      |
-| :-------------------------- | :------------------------------------------------------------------------------ | :--------------- |
-| `struc[1]`                  | Model 1                                                                         | `Model`          |
-| `struc[1]['A']`             | Model 1, chain A                                                                | `Chain`          |
-| `struc['A']`                | The lowest model (model 1), chain A                                             | `Chain`          |
-| `struc['A']["50"]`          | Model 1, chain A, residue 50                                                    | `Residue`        |
-| `struc['A'][50]`            | Shortcut to above if it is not a hetero residue and the insertion code is blank | `Residue`        |
-| `struc['A']["H_90"]`        | Model 1, chain A, hetero residue 90                                             | `Residue`        |
-| `struc['A'][50]["CA"]`      | Model 1, chain A, residue 50, atom name CA                                      | `Atom`           |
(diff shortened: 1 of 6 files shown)


bicycle1885:
New features look really awesome! I don't use PDB so often but my labmates may switch to Bio.jl from  BioPyXXXX. Just curious; is there ETA in your mind?

jgreener64:
Great! Thanks for the interest.

Hopefully the end of next week will have working code, then another week for refining/testing. Predicted RTR in two weeks I would say. Just a case of squeezing the work round all the other manic commitments of PhD :wink:.

Also I laughed at BioPyXXXX :ok_hand:.

jgreener64:
Okay, so this is ready to review!

Changes implemented as stated above. The changes to the type system are that each element is directly associated with its super-element (e.g. a `Residue` has a `chain` field), whereas before it was not (it just had a `chain_id` field). This provides more functionality but necessitates big changes, so it may be easier to read from scratch than looking at the diff.

However, during the switch to Julia v0.5 and all the associated repo faffing I appear to have done something weird to the commit history. I don't think commits from other people should appear here? Is there a way to remove this?

bicycle1885:
Cool! Thanks a lot!

I don't know what happened in the cluttered commit log, but a simple way to tidy up may be:
1. creating a new branch from the origin/master,
2. cherry-picking necessary commits you made to the branch and
3. force-pushing it to this pull request.

I recommend you to create a backup branch of this pull request before doing that so that you won't lose your works. Once you finish cleaning up the pull request, I will start to review it ASAP.

jgreener64:
Thanks @bicycle1885. Fixed with a rebase and force push.

Have a look and let me know what you think!

src/structure/model.jl
```diff
@@ -9,211 +9,288 @@ export
...
+
+function Residue(name::AbstractString,
+                number::Integer,
+                ins_code::Char,
+                het_res::Bool,
+                ch::Chain)
+    return Residue(name, number, ins_code, het_res, [], Dict(), ch)
+end
+
+Residue(ch::Chain) = Residue("ALA", 1, ' ', false, ch)
```
bicycle1885: I think this kind of default values (`"ALA"`, `1`) will lead to hard-to-find bugs in the future. Is that possible to force making a structure from outer (`ProteinSructure`) to inner (`Atom`) elements? Passing a wrapper element to constructors will remove the necessity of these default values.

src/structure/model.jl
```diff
@@ -222,378 +299,574 @@ end
...
+
+function Base.setindex!(struc::ProteinStructure, ch::Chain, ch_id::Char)
+    defaultmodel(struc)[ch_id] = ch
     return struc
 end
 
+# Check if an atom name exists in a residue as a whitespace-padded version
+function findatombyname(res::Residue, atom_name::AbstractString)
+    # Look for atom name directly
+    if atom_name in keys(res.atoms)
```
bicycle1885: `haskey(res.atoms, atom_name)`

src/structure/model.jl
```diff
@@ -222,378 +299,574 @@ end
...
+        return res.atoms[atom_name]
+    # Pad out name to 4 characters to read PDB atom names with whitespace
+    elseif length(atom_name) == 3
```
bicycle1885: I'm not sure but I thought that linear search may be faster than padding names because this padding will create so many garbage objects while searching.

src/structure/model.jl
```diff
@@ -222,378 +299,574 @@ end
...
 
 """
-Determines if an `AbstractAtom` represents a non-hetero atom, i.e. came from an
-ATOM record.
+Get the atom name of an `AbstractAtom`.
+`spaces` determines whether surrounding whitespace is retained (default
+`false`).
 """
-ishetatom(atom::Atom) = atom.het_atom
+function atomname(at::Atom; spaces::Bool=false)
```
bicycle1885: `strip` or `stripspaces` would be easier to understand.

src/structure/model.jl
```diff
@@ -222,378 +299,574 @@ end
...
+    if spaces
+        return at.name
+    else
+        return strip(at.name)
```
bicycle1885: If you choose `strip` as the keyword argument name, you can call this as `Base.strip(at.name)`.

src/structure/model.jl
```diff
@@ -222,378 +299,574 @@ end
...
+The default is the highest occupancy, or lowest character alternative location
+ID for ties.
 """
-atomid(atom::Atom) = (resid(atom; full=true), resname(atom), atomname(atom))
+defaultaltlocid(dis_at::DisorderedAtom) = dis_at.default
+
+
+# Constructor acts as a setter for the default alt loc ID
+function DisorderedAtom(dis_at::DisorderedAtom, default::Char)
+    @assert default in altlocids(dis_at) "The new default alternative location ID must be present in the atom"
```
bicycle1885: `@assert` should not be used for checking function arguments.  `throw(ArgumentError(message))` would be better in general.

src/structure/model.jl
```diff
@@ -663,463 +956,658 @@ Base.eltype(::Type{Residue}) = AbstractAtom
...
+        end
+    end
+    return AminoAcidSequence(seq)
+end
 
 
 # Descriptive showing of elements
 
 function Base.show(io::IO, struc::ProteinStructure)
     if countmodels(struc) > 0
```
bicycle1885: Showing the type name first would be nice (e.g. `println(io, summary(struct))`).

src/structure/model.jl
```diff
@@ -663,463 +956,658 @@ Base.eltype(::Type{Residue}) = AbstractAtom
...
-        println(io, "Number of hydrogens         -  ", countatoms(model, stdatomselector, hydrogenselector))
-        println(io, "Number of disordered atoms  -  ", countatoms(model, stdatomselector, disorderselector))
+        println(io, "Chain(s)                    -  ", join(chainids(mod)))
+        println(io, "Number of residues          -  ", countresidues(mod, stdresselector))
+        println(io, "Number of point mutations   -  ", countresidues(mod, stdresselector, disorderselector))
+        println(io, "Number of other molecules   -  ", countresidues(mod, hetresselector) - countresidues(mod, hetresselector, waterselector))
+        println(io, "Number of water molecules   -  ", countresidues(mod, hetresselector, waterselector))
+        println(io, "Number of atoms             -  ", countatoms(mod, stdatomselector))
+        println(io, "Number of hydrogens         -  ", countatoms(mod, stdatomselector, hydrogenselector))
+        println(io, "Number of disordered atoms  -  ", countatoms(mod, stdatomselector, disorderselector))
```
bicycle1885: Printing a newline at the last line will print an extra newline in REPL. I think `print(io, "Number of...."))` would be better.

src/structure/model.jl
```diff
@@ -663,463 +956,658 @@ Base.eltype(::Type{Residue}) = AbstractAtom
...
-        println(io, "Number of disordered atoms  -  ", countatoms(disorderedres(disordered_res, res_name), disorderselector))
+        println(io, "Number of atoms             -  ", countatoms(disorderedres(dis_res, res_name)))
+        println(io, "Number of hydrogens         -  ", countatoms(disorderedres(dis_res, res_name), hydrogenselector))
+        println(io, "Number of disordered atoms  -  ", countatoms(disorderedres(dis_res, res_name), disorderselector))
     end
 end
 
-Base.show(io::IO, atom::Atom) = println(io, pdbline(atom)...)
-Base.showcompact(io::IO, atom::Atom) = print(io, pdbline(atom)...)
+Base.show(io::IO, at::Atom) = println(io, pdbline(at)...)
```
bicycle1885: `show` should print "an informative text representation of a value" (from the help of `Base.show`) rather than a PDB-formatted text. Informative text would be something like [this](https://github.com/BioJulia/Bio.jl/blob/master/src/align/hts/sam/record.jl#L48-L62).

src/structure/model.jl
```diff
@@ -663,463 +956,658 @@ Base.eltype(::Type{Residue}) = AbstractAtom
...
 
-function Base.show(io::IO, disordered_atom::DisorderedAtom)
-    for atom in disordered_atom
-        show(io, atom)
+Base.showcompact(io::IO, at::Atom) = print(io, pdbline(at)...)
```
bicycle1885: Now you should not extend `Base.showcompact`:
> To offer a compact representation different from its standard one, a custom type should test `get(io, :compact, false)` in its normal show method.

bicycle1885:
I've quickly skimmed the diffs and added some comments. Looks really awesome!
I'll need a few more days to finish review since the change is so big.

jgreener64:
Thanks, will implement these suggestions in the mean time.

src/structure/model.jl
jgreener64: Removed this in the end.

src/structure/model.jl
jgreener64: Cool, changed.

src/structure/model.jl
jgreener64: Ah yes, changed throughout.

src/structure/model.jl
jgreener64: Added.

src/structure/model.jl
jgreener64: Cool, changed throughout.

src/structure/model.jl
jgreener64: Changed to `strip`.

src/structure/model.jl
jgreener64: I always forget this haha, changed throughout.

src/structure/model.jl
jgreener64: Ah okay. How would I go about implementing that in this case? Thanks.

jgreener64:
Have added changes to reflect the above @bicycle1885. When you have time for a further look let me know of other suggestions, thanks!

src/structure/spatial.jl
```diff
@@ -1,66 +1,298 @@
...
+
+
+"""
+Calculate the `Vector`s of phi and psi angles of a `StructuralElementOrList`.
+The vectors have `nothing` for residues where an angle cannot be calculated,
+e.g. due to missing atoms or lack of an adjacent residue.
+Additional arguments are residue selector functions - only residues that return
+`true` from the functions are retained.
+"""
+function ramachandranangles(el::StructuralElementOrList,
```
bicycle1885: Including `nothing` in an array will result in very bad performance in Julia. I think using `NaN`, which is a value of `Float64`, or using `NullableArray` of NullableArrays.jl is a better way to avoid the problem.

src/structure/spatial.jl
```diff
@@ -1,66 +1,298 @@
...
+        end
+    end
+    push!(psi_angles, nothing) # Last res has no next res
+    return phi_angles, psi_angles
+end
+
+
+"""
+Calculate the contact map for a `StructuralElementOrList`, or between two
+`StructuralElementOrList`s. This is an `Array` with `true` where the
```
bicycle1885: Actually, it will not be an `Array` object if you use `falses`. It is `BitArray` (or `BitMatrix`).

src/structure/spatial.jl
```diff
@@ -1,66 +1,298 @@
...
+        end
     end
-    return sqrt(minimum(sq_dists))
+    return contacts
 end
 
-distance(atom_one::AbstractAtom, atom_two::AbstractAtom) = sqrt((x(atom_one) - x(atom_two)) ^ 2 + (y(atom_one) - y(atom_two)) ^ 2 + (z(atom_one) - z(atom_two)) ^ 2)
+function contactmap(el::StructuralElementOrList, contact_dist::Real)
+    sq_contact_dist = contact_dist ^ 2
+    contacts = eye(Bool, length(el))
```
bicycle1885: I think we should stick to either `Array{Bool,2}` or `BitArray{2}`.

src/structure/spatial.jl
```diff
@@ -1,66 +1,298 @@
...
+"""
+Calculate the phi angle in radians for an `AbstractResidue`. Arguments are the
+residue (atoms "N", "CA" and "C" required) and the previous residue (atom "C"
+required).
+"""
+function phiangle(res::AbstractResidue, res_prev::AbstractResidue)
+    "C"  in atomnames(res_prev) ? nothing : throw(ArgumentError("Atom with atom name \"C\" not found in previous residue"))
+    "N"  in atomnames(res)      ? nothing : throw(ArgumentError("Atom with atom name \"N\" not found in residue"))
+    "CA" in atomnames(res)      ? nothing : throw(ArgumentError("Atom with atom name \"CA\" not found in residue"))
+    "C"  in atomnames(res)      ? nothing : throw(ArgumentError("Atom with atom name \"C\" not found in residue"))
```
bicycle1885: A sequence of if-elseif clauses looks less magical to me.

jgreener64:
Thanks, will address these.

src/structure/spatial.jl
jgreener64: Yeah true.

src/structure/spatial.jl
```diff
@@ -1,66 +1,298 @@
...
 """
 function rmsd(coords_one::Array{Float64}, coords_two::Array{Float64})
-    @assert size(coords_one) == size(coords_two) "Sizes of coordinate arrays are different - cannot calculate RMSD"
-    difference = coords_one - coords_two
-    return sqrt(sum(difference .* difference) / size(coords_one, 2))
+    if size(coords_one) != size(coords_two)
+        throw(ArgumentError("Sizes of coordinate arrays are different - cannot calculate RMSD"))
+    end
+    diff = coords_one - coords_two
+    return sqrt(sum(diff .* diff) / size(coords_one, 2))
```
jgreener64: Ah yes.

src/structure/model.jl
```diff
@@ -222,378 +267,575 @@ end
...
 """
-Determines if an `AbstractAtom` represents a hetero atom, i.e. came from a
-HETATM record, or if an `AbstractResidue` represents a hetero molecule, i.e.
-consists of HETATM records.
+Determines if an `AbstractAtom` represents a non-hetero atom, e.g. came from an
+ATOM record in a Protein Data Bank (PDB) file.
 """
-ishetero(atom::AbstractAtom) = ishetatom(atom)
+ishetatom(at::Atom) = ishetres(residue(at))
+ishetatom(dis_at::DisorderedAtom) = ishetatom(defaultatom(dis_at))
```
jgreener64: `ishetero` exists already, and it is sufficient I agree, so I will remove `ishetatom` and `ishetres`.

src/structure/model.jl
```diff
@@ -222,378 +267,575 @@ end
...
+
 
 """
 Determines if an `AbstractAtom` is a `DisorderedAtom`, i.e. if there are
 multiple locations present for an atom.
 """
-isdisorderedatom(::Atom) = false
+isdisorderedatom(at::AbstractAtom) = isa(at, DisorderedAtom)
```
jgreener64: Unlike `ishetero` I think these need to stay separate, because disorder on the atom and residue levels are separate. For example a disordered residue can contain atoms in a particular residue (point mutation) that are not disordered, and disordered atoms do not need to be part of a disordered residue.

src/structure/model.jl
```diff
@@ -663,463 +925,715 @@ Base.eltype(::Type{Residue}) = AbstractAtom
...
+collectchains(ch::Chain) = [ch]
+
+collectchains(el::Union{AbstractResidue, AbstractAtom}) = [chain(el)]
+
+function collectchains(mods::Vector{Model})
+    ch_list = Chain[]
+    for mod in mods
+        append!(ch_list, collectchains(mod))
+    end
+    return sort(ch_list)
```
jgreener64: Cool, changed throughout.

docs/src/man/structure.md
```diff
@@ -200,23 +225,51 @@ A few examples of `Bio.Structure` usage are given below.
...
 using Gadfly
-atoms = collectatoms(struc, calphaselector)
-res_numbers = map(resnumber, atoms)
-temp_facs = map(tempfac, atoms)
-plot(x=res_numbers, y=temp_facs, Geom.line)
+calphas = collectatoms(struc, calphaselector)
+res_numbers = map(resnumber, calphas)
+temp_facs = map(tempfac, calphas)
+plot(x=res_numbers,
+    y=temp_facs,
```
jgreener64: Didn't even know that syntax existed! Changed.

docs/src/man/structure.md
```diff
@@ -200,23 +225,51 @@ A few examples of `Bio.Structure` usage are given below.
...
+
+`cbetaselector` selects C-beta atoms, or C-alpha atoms for glycine residues. `contactmap` can also be given two structural elements as arguments, in which case a non-symmetrical 2D array is returned showing contacts between the elements.
+
+**E)** To show the Ramachandran phi/psi angle plot of a structure, if you have Gadfly installed:
+
+```julia
+using Gadfly
+phi_angles, psi_angles = ramachandranangles(struc, stdresselector)
+plot(x=map(rad2deg, phi_angles),
+    y=map(rad2deg, psi_angles),
```
jgreener64: Changed.

src/structure/model.jl
```diff
@@ -222,378 +267,575 @@ end
...
+
+
+"Get the occupancy of an `AbstractAtom`. Defaults to `1.0`."
+occupancy(at::Atom) = at.occupancy
+occupancy(dis_at::DisorderedAtom) = occupancy(defaultatom(dis_at))
+
+
+"Get the temperature factor of an `AbstractAtom`. Defaults to `0.0`."
+tempfac(at::Atom) = at.temp_fac
+tempfac(dis_at::DisorderedAtom) = tempfac(defaultatom(dis_at))
```
jgreener64: Yeah makes sense, have changed.

src/structure/spatial.jl
jgreener64: Changed.

src/structure/spatial.jl
jgreener64: Good spot - have made this `BitArray{2}` like previously.

src/structure/spatial.jl
jgreener64: Yeah makes sense, have used `NaN`.

docs/src/man/structure.md
```diff
@@ -185,13 +185,38 @@ PDB format files can be written:
...
+julia> rad2deg(bondangle(struc['A'][50]["N"], struc['A'][50]["CA"], struc['A'][50]["C"]))
+69.22234153916602
+
+julia> rad2deg(dihedralangle(struc['A'][50]["N"], struc['A'][50]["CA"], struc['A'][50]["C"], struc['A'][51]["N"]))
+-177.38288114072924
+
+julia> rad2deg(psiangle(struc['A'][50], struc['A'][51]))
+-177.38288114072924
+```
+
```
jgreener64: Yes will do.

jgreener64:
Suggestions implemented, thanks @bicycle1885. Let me know if there's anything else, otherwise feel free to merge!

jgreener64:
@bicycle1885 when you get a chance let me know if this is ready to merge, thanks.

bicycle1885:
Sorry, I didn't merge because the TravisCI failed on Julia 0.5. But it was not reproducible on my laptop, so I thought I needed to kick it again. Now I'm running it again. If it passes, I will merge soon.

jgreener64:
Yeah, the build seems to be failing on the spatial tests for v0.5 (not nightlies). This is weird. Possibly something to do with lack of inlining during test runs? Or something to do with simd. Will have a play.

bicycle1885:
I guess SIMD warnings are not directly related to the failure. They are just false alarms when running the Julia program with some option. The most probable cause would be in opening/closing data files. I remember this kind of hanging happened due to opening too many files but I cannot recall where I saw it.

jgreener64:
Yeah it could be that. Triggering build with printed checkpoints to track the error.

jgreener64:
Failure seems to occur between checkpoint 11 and 12, i.e. on the very last set of tests ("# Test contact map"). It also seems to occur intermittently on the Julia v0.5 builds.

One option would be to take these tests out.

bicycle1885:
What happens if we comment out the first/second half of the code between checkpoint 11 and 12?

jgreener64:
Have added more detailed checkpoints to explore this.

jgreener64:
Problem appears to be with the line `cmap = contactmap(struc_1AKE['A'], struc_1AKE['B'], 5.0)` - have commented this out and will see what happens.

jgreener64:
Have replaced the test in question with an easier version that checks the same thing.

jgreener64:
Okay, I think it's all fixed now!

bicycle1885:
Awesome! Now all tests pass. I'll merge this today if no further comments.

bicycle1885:
Merged. Thank you very much, @jgreener64!

jgreener64:
Great, thanks a lot for all the help @bicycle1885!

Ward9250:
Nice one @jgreener64, thanks very much :+1:

PR #256, opened 2016-08-12
Make tests that create files happen in temp dir

kdmurray91:
This PR adds a function `intempdir` that lets a block of code (as an anonymous function) happen in a temp directory that gets created and deleted automagically. I use this function to stop the `Interval` tests from leaking BED files into the test sources directory.

### DIFF
--- a/src/intervals/bed.jl
+++ b/src/intervals/bed.jl
@@ -115,6 +115,10 @@ function Base.eof(parser::BEDParser)
     return eof(parser.state.stream)
 end
 
+function Base.close(parser::BEDParser)
+    return close(parser.state.stream)
+end
+
 function Base.open(input::BufferedInputStream, ::Type{BED})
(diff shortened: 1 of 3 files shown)


kdmurray91:
This raises the question: Should we do some kind of CI check (at least in travis) that asserts there are no files left over in the source directories after the tests run? something like `test -z "$(git clean -ndx)"`?

kdmurray91:
This fails on windows for some unintelligible reason. Does someone with windows experience want to talk a look?

kdmurray91:
Looks like closing all files fixes the windows error and all tests pass. The changes to the BED sources simply define `close()` on the parser, and the refactored BED tests ensure that all files are closed (and that tests happen in temp dirs).

Can I merge this @Ward9250 @bicycle1885?

Ward9250:
:+1: I think so mate!

PR #257, opened 2016-08-14
remove clipboard(seq)

bicycle1885:
This code is no longer needed.


Ward9250:
Been trying to catch up on the week I had no net - why is clipboard no
longer needed? I think I missed that development.

bicycle1885:
`clipboard(x)` calls `print(x)` internally and `print(::Sequence)` now prints a sequence in ASCII. So, we can use the default `clipboard` method for sequence types without these specialized methods.

PR #258, opened 2016-08-15
Mutation counting

Ward9250:
Implementation of counting of different types of mutations.

A nessecery precursor to the evolutionary distance PR, but sizable enough that I've separated into its own PR.

At first (in this PR), only two kinds of mutation counting will be implemented: counting different nucleotides, and counting the number of transitions and number of transversions.

The count_mutations method uses a strict ambiguity test, which discounts a site completely from the computation if either nucleotide is uncertain or a gap i.e. ambiguity codes.

I tried experimenting (hence all the commits!) with making use of all symbols including ambiguities as sometimes it is possible to tell if two ambiguous sites, are different or are say a transition/transversion despite their ambiguity. However because of the complexity of doing this, plus the bias it could introduce in downstream analysis if applied uncritically (for example you can always detect more mutated sites with ambiguity codes, than you can conserved sites), it is not worth doing - if a sequence were that ambiguous, maybe you need better sequence data in the first place ;).


Ward9250:
Ready for review. If tests pass I'd like to squash and merge asap if there are no objections.

Ward9250:
Tests now pass, I'd like to merge in the morning (UK) if no objections. There's plenty of room for improvement e.g. bitwise computations for 2 bit sequences if they are known. But for now this works well for the distance computation PR.

src/var/Var.jl
```diff
@@ -0,0 +1,30 @@
...
+#
+# This file is a part of BioJulia. License is MIT: https://github.com/BioJulia/Bio.jl/blob/master/LICENSE.md
+
+
+
+module Var
+using Bio.Seq
+
+export
+
```
kdmurray91: pedantry: do we need the extra newlines here?

src/var/mutation_counting.jl
```diff
@@ -0,0 +1,152 @@
...
+
+Count the number of differences between two DNA sequences.
+Returns a tuple of the number of sites that are different, and the number of sites
+considered (sites with any ambiguity characters are not considered).
+"""
+function count_mutations{A<:alphabets}(seq1::BioSequence{A},
+    seq2::BioSequence{A},
+    t::Type{Different})
+
+    @assert length(seq1) == length(seq2)
```
kdmurray91: Should this use `error()` to return a more informative message to the user?

src/var/mutation_counting.jl
```diff
@@ -0,0 +1,152 @@
...
+A, T, G, or C, then this function returns true.
+"""
+@inline function is_ambiguous_strict{T<:Nucleotide}(a::T, b::T)
+    return isambiguous(a) || isambiguous(b)
+end
+
+typealias alphabets Union{DNAAlphabet{2}, DNAAlphabet{4}, RNAAlphabet{2}, RNAAlphabet{4}}
+
+"""
+    count_mutations(seq1::BioSequence{A}, seq2::BioSequence{A}, kind::Type{Different})
```
kdmurray91: is the `{A}` here required? you don't have it in the signatures for the other mutation types.

src/var/mutation_counting.jl
```diff
@@ -0,0 +1,152 @@
...
+    @assert length(seq1) == length(seq2)
+    nDifferences = 0
+    nSites = length(seq1)
+
+    for (a, b) in zip(seq1, seq2)
+        if is_ambiguous_strict(a, b)
+            nSites -= 1
+            continue
+        end
+        if a != b
```
kdmurray91: Could one re-factor this to use an if statement on the type of mutation required, to reduce code duplication?

kdmurray91:
Looks great @Ward9250. I've left a couple of comments, but I don't consider them serious enough to block a merge if you need this in master now-ish.

docs/src/man/var.md
```diff
@@ -0,0 +1,36 @@
...
+| :------------- | :-------------------------------------- |
+| `Different`    | Nucleotides that are different          |
+| `Transition`   | Transition mutations                    |
+| `Transversion` | Transversion mutations                  |
+
+`Different` simply compares two nucleotides to see if they can be deemed
+the same or different. Ambiguous cases are detected and excluded from the
+computation.
+
+`TsTv` compares two nucleotides to see if they can be determined
```
bicycle1885: `TsTv` is unclear to me.

docs/src/man/var.md
```diff
@@ -0,0 +1,36 @@
...
+to be a transition mutation, a transversion mutation, or no mutation at all.
+As with `DifferentBases`, cases where a nucleotide is ambiguous are excluded
```
bicycle1885: `DifferentBases`: do you mean `Different`?

src/seq/nucleotide.jl
```diff
@@ -123,6 +123,16 @@ function ispyrimidine(nt::Nucleotide)
 end
 
 """
+    isppambiguous(nt::Nucleotide)
+
+Test if nucleotide could be either a Purine or a Pyrimidine.
+"""
+function isppambiguous(nt::Nucleotide)
```
bicycle1885: Is this really worth defining & exporting from Bio.Seq? There is no abstraction here.

src/var/mutation_counting.jl
```diff
@@ -0,0 +1,152 @@
...
+
+
+# Counting mutations
+# ------------------
+
+# Mutation types
+
+immutable Different end
+immutable Transition end
+immutable Transversion end
```
bicycle1885: These names are too generic. Names like `TransitionMutation` would be better I think.

src/var/mutation_counting.jl
```diff
@@ -0,0 +1,152 @@
...
+    is_ambiguous_strict{T<:Nucleotide}(a::T, b::T)
+
+The strictest test of ambiguity, if either nucleotide is not
+A, T, G, or C, then this function returns true.
+"""
+@inline function is_ambiguous_strict{T<:Nucleotide}(a::T, b::T)
+    return isambiguous(a) || isambiguous(b)
+end
+
+typealias alphabets Union{DNAAlphabet{2}, DNAAlphabet{4}, RNAAlphabet{2}, RNAAlphabet{4}}
```
bicycle1885: All alphabet types are a subtype of `Bio.Seq.Alphabet`. Why do we need this alias?

src/var/mutation_counting.jl
```diff
@@ -0,0 +1,152 @@
...
+Returns a tuple of the number of sites that are different, and the number of sites
+considered (sites with any ambiguity characters are not considered).
+"""
+function count_mutations{A<:alphabets}(seq1::BioSequence{A},
+    seq2::BioSequence{A},
+    t::Type{Different})
+
+    @assert length(seq1) == length(seq2)
+    nDifferences = 0
+    nSites = length(seq1)
```
bicycle1885: `nDifferences` and `nSites` look like CamelCase, which is not a standard way to name variables in Julia, though these are local variables and not a big deal.

src/var/mutation_counting.jl
```diff
@@ -0,0 +1,152 @@
...
+end
+
+typealias alphabets Union{DNAAlphabet{2}, DNAAlphabet{4}, RNAAlphabet{2}, RNAAlphabet{4}}
+
+"""
+    count_mutations(seq1::BioSequence{A}, seq2::BioSequence{A}, kind::Type{Different})
+
+Count the number of differences between two DNA sequences.
+Returns a tuple of the number of sites that are different, and the number of sites
+considered (sites with any ambiguity characters are not considered).
```
bicycle1885: Insert a blank line after the first sentence.

> Include a single one-line sentence describing what the function does or what the object represents after the simplified signature block. If needed, provide more details in a second paragraph, after a blank line.
<http://julia.readthedocs.io/en/latest/manual/documentation/>

bicycle1885:
About the design choice: What about counting all mutations and packing them in a single object like this:
```julia
immutable MutationCounts
    transition::Int
    transversion::Int
    deletion::Int
    insertion::Int
    difference::Int
    length::Int
end

function count_mutations(seq1, seq2)
    ...
    return MutationCounts(...)
end
```

src/var/mutation_counting.jl
Ward9250: Changed names - next commit.

src/seq/nucleotide.jl
Ward9250: I think this is something accidentally left in from when I was trying to be clever and use all the possible ambiguity codes before I decided/realised that doing so was a stupid idea (expanded in first comment). I'll remove this in the next commit.

docs/src/man/var.md
Ward9250: I do, corrected in next commit.

docs/src/man/var.md
Ward9250: Removed in next commit.

src/var/mutation_counting.jl
Ward9250: I had this in there so as to exclude the amino acid, character and void alphabets. This PR only handles counting mutations in DNA and RNA sequences.

src/var/mutation_counting.jl
Ward9250: Changed to non camel in next commit.

src/var/mutation_counting.jl
Ward9250: Added in next commit.

src/var/Var.jl
Ward9250: Extra lines under header removed in next commit.

src/var/mutation_counting.jl
Ward9250: I see what you mean. I hesitate to add a run-time decision on the mutation type though. Would a generated function work for this purpose?

src/var/mutation_counting.jl
bicycle1885: OK, in that case, more explicit name like `NucleotideAlphabet` will make sense.

Ward9250:
Thanks for comments, I've responded to each one and changes will be pushed from my desktop in next commit, but for now it's 3:31am so I'm signing off for a bit!

@bicycle1885 Re the design choice. I would indeed like a method that returns a summary of all the mutation counts. However for distance computations only info on one or two mutation types is required e.g. JC69 only considers `Different` mutations, whilst K80 considers `Transition` and `Transversion` mutations (hence my provision of a `count_mutations` methods accepting both of those types to compute both together in one loop). However, if the overhead of computing say `Transitions` and `Transversion` counts, even though you only need `Different` counts is negligible then it might be a better idea.

Actually I think one method that computes them all is a good idea anyway, but the above was my reasoning behind individual methods as well.

src/var/mutation_counting.jl
Ward9250: Awesome, I've added the name change and will be in next commit!

Ward9250:
Ok, just completed a rebase to squash, when check turn green I'll merge.

Ward9250:
There is an error on OSX, but I believe it is to do with HDF5, and not related to the code submitted in this PR.

PR #261, opened 2016-08-16
remove deprecated names

bicycle1885:
We should have removed these functions before the release of Bio.jl 0.3.


bicycle1885:
This change would be not controversial. I'm going to merge this soon.

PR #262, opened 2016-08-16
rename ispalindrome to ispalindromic

bicycle1885:
I'm not completely sure but `ispalindromic` would be more common than `ispalindrome`. More generally, I'd like to think that `XXX` of `isXXX` should be an adjective word if possible.


Ward9250:
:+1:

PR #264, opened 2016-08-19
add Distributions.jl to REQUIRE

bicycle1885:
Distributions.jl is required in src/phylo/dating.jl.


Ward9250:
Good catch @bicycle1885 thanks for that!

PR #270, opened 2016-08-20
Furthur readme edit

Ward9250:
Furthur to @bicycle1885's changes, I forgot to ask for Bio.Phylo to be added!
Also propose adding header above badges.


bicycle1885:
Nice, thank you! It's my bad 😢 .

Ward9250:
Nah you were 100% right, the new style looks great!

Issue #272, opened 2016-08-22
Base.parse for Bio.jl types


Ward9250:
So I'm writing a script which uses arg parser, and I've seen that you can enforce types of command line arguments, but also you can do this for custom types: http://argparsejl.readthedocs.io/en/latest/argparse.html#parsing-to-custom-types
This would be cool for making scripts that accept biological sequences, or phylogenetic trees or, well anything through a text command line arg. You can ArgParse.parse_item, but the default parse_item using Base.parse so I thought it would be cool to implement those methods for our datatypes. Sequences should be easy to do, as we already have convert functions that produce biological sequences from strings. So this could be as simple as `Base.parse(DNASequence, string) = convert(DNASequence, string)`??? I'm going to open a PR.

bicycle1885:
Can we close now? xref: #273

Ward9250:
I think we can, I was going to keep it open to remind me to do similar for Phylo so as Newick can be parsed by ArgParse too, but I'll just remember.

PR #273, opened 2016-08-22
Added Base.parse for biological sequences

Ward9250:
Implements #272 for Biological Sequences.


bicycle1885:
This definition is more general:
```
Base.parse{S<:Sequence}(::Type{S}, str::AbstractString) = convert(S, str)
```

PR #275, opened 2016-08-22
improve the performance of sequence type prediction

bicycle1885:
This improves the performance of parsing a FASTA file when the sequence is a long DNA or RNA sequence like chromosomal sequences.

Benchmark results of parsing chromosome 1 of human are:

Before (master):
```
  6.017943 seconds (34.78 M allocations: 801.361 MB, 1.46% gc time)
```

After:
```
  3.909558 seconds (34.78 M allocations: 801.361 MB, 2.55% gc time)
```


Ward9250:
:+1:

PR #277, opened 2016-08-23
improve the performance of encoding ASCII string

bicycle1885:
This improves the performance of encoding an ASCII string into a DNA/RNA sequence, which also improves the performance of parsing a FASTA file.

Before (#275):
```
 3.909558 seconds (34.78 M allocations: 801.361 MB, 2.55% gc time)
```

After:
```
  3.013747 seconds (34.78 M allocations: 801.361 MB, 3.01% gc time)
```


Ward9250:
:+1:

PR #279, opened 2016-08-24
Create a majority vote sequence from a vector of sequences

Ward9250:


Ward9250:
@bicycle1885 Can I get your opinion on something?

In order to get this PR to work, we need to get the most abundant nucleotide from a Composition{T} variable.  

maximum to work for `Composition{T}`, or some other method to get the most abundant nucleotide.

First I thought we could just use the maximum and minimum functions from the julia standard library - since Composition{T} inherits from Associative.

However, doing `maximum()` on a Composition or a Dict does not work. This is because this function is part of `reduce.jl` in julia, where `maxmimum(dict) = mapreduce(identity, scalarmax, dict)  = mapfoldl(identity, scalarmax, dict) = mapfoldl_impl(identity, scalarmax, v0, dict, i)`

For Dict and Composition, this essentially iterates, which yields a Pair every iteration and does `scalarmax` on the current and previous Pair, which returns which of the two is higher. To do this `scalarmax` uses `isless`, and `isless(::Pair, ::Pair)` as defined in `operators.jl` in julia is not suitable, and means when you iterate over a dict, Pairs are compared with `isless` using the keys, and not the values.

So I have two options. First we could redefine isless for Pair{Nucletide, Int}, this would make it work with maximum, but I'm hesitant to do this as it would make the behaviour of maximum for Composition & Dict{Nucleotide, Int}, different from all other Associative's like Dicts. Which might cause headaches for people who don't realise.

Second option is we simply introduce a new method for Composition only called something like mostfrequent()/mostabundant()/freqmax().

I think the second option is more sensible, but wanted to get your input.

bicycle1885:
I agree with you. I think overriding the order of the `Pair` type is generally a bad idea and `mostfrequent` or something for `Composition` would make sense to me.

One think I'm worrying about is `mostfrequent` has multiple choices of the return value when there is a tie. What should `mosfrequent(composition(dna"AACC"))` return? Returning a smaller symbol may lead to a bias in some analysis.

Ward9250:
That could be problematic when making consensus sequences also. One solution is the function returns a vector of each nucleotide which has a count == max(counts). Most of the time this output vector is of length 1, but sometimes higher, and contains all nucleotides when they are all equally abundant.

bdeonovic:
@bicycle1885 in case of ties for nucleotide frequency could we return ambiguous nucleotide code? Or perhaps have that be a toggleable option?

bicycle1885:
Unfortunately that won't work because `composition` counts ambiguous nucleotides, too.

Ward9250:
Would one way it could work be that ambigs increment their constituent nucleotides?

Say you had a Composition with A => 3 and T => 3 and C=>2, say there was a W => 1.

W get's -= 1 and A and T get += 1.

I believe since each ambiguous character increments it's underlying states equally, it doesn't bias away from either one of the underlying states. But with other amigs, one of the underlying states can still win out as more frequent, if there is a tie, then the ambig nucleotide for that tie can be returned.

bicycle1885:
Consensus sequence does not always "majority vote" of sequences but it may be a more complicated description like regular expression or probability model. So I think, at the moment, we should not implement the `consensus` method. Instead, we should write ad-hoc consensus construction algorithms that meet the demands. Once we find a good design, then we can add such a `consensus` function like the `pairalign` function of `Bio.Align`.

Ward9250:
You're right consensus might mean other things. So should we instead call this PR I'm working on majorityvote instead? Instead of making a consensus method make a majorityvote method. Then it's usable, but it's very clear that it is a specific algorithm, and if it gets absorbed by some consensus functionality later then that's all good?

Ward9250:
@bicycle1885 I've added 3 majority vote algorithms which do the same thing but are written in different ways to see which are more performant. Each one tallies the gap votes, a votes, t votes, c votes and g votes, and ambiguous ones cast a vote for each constituent equally as I previously described. If there is one winner at the end it is returned. If there are ties, each is merged to make an ambiguous nucleotide which is returned.
There's not much in it time wise:
```julia
julia> @benchmark _mvote_compat1(c)
BenchmarkTools.Trial: 
  samples:          10000
  evals/sample:     10
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  1008.00 bytes
  allocs estimate:  15
  minimum time:     1.40 μs (0.00% GC)
  median time:      1.46 μs (0.00% GC)
  mean time:        1.65 μs (7.56% GC)
  maximum time:     389.30 μs (98.33% GC)

julia> @benchmark _mvote_compat2(c)
BenchmarkTools.Trial: 
  samples:          10000
  evals/sample:     10
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  928.00 bytes
  allocs estimate:  14
  minimum time:     1.25 μs (0.00% GC)
  median time:      1.31 μs (0.00% GC)
  mean time:        1.49 μs (7.95% GC)
  maximum time:     346.61 μs (98.71% GC)

julia> @benchmark _mvote_compat3(c)
BenchmarkTools.Trial: 
  samples:          10000
  evals/sample:     10
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  928.00 bytes
  allocs estimate:  14
  minimum time:     1.25 μs (0.00% GC)
  median time:      1.31 μs (0.00% GC)
  mean time:        1.49 μs (7.82% GC)
  maximum time:     303.10 μs (98.46% GC)
```

Ward9250:
I'm going to go with the third method.

Ward9250:
Hey @bicycle1885 feeling pretty pleased with myself: I managed to create an improved `majorityvote` function at the bottom of the src/seq/bioseq.jl file. It uses the `seqmatrix` functionality I merged earlier, managed to loop efficiently by choosing the right matrix orientation for looping, and made code straight-line so I can see julia decided the inner loop can be SIMD and put appropriate llvm instructions in the `@code_llvm` output :)

Currently it only works for DNASequences, I want it to work for RNA sequences too, but wondered if you thought, looking at the function, whether parametric of @generated was the way to go (one issue with parametric is currently DNA_A, DNA_T and so on, are used in the body of the function).

Ward9250:
Ok, I've got the faster `majorityvote` working. Currently it is a generated function, but I think it can be made a general function if I implement a `seqmatrix` method that returns a matrix of UInt8's instead and so the voting operations on those instead of specific nucleic acid types. It might even add a tiny performance increase.

bicycle1885:
I'll take a look it ASAP.

kdmurray91:
Looks good to me!

src/seq/alphabet.jl
```diff
@@ -42,6 +42,8 @@ Void alphabet (internal use only).
 """
 immutable VoidAlphabet <: Alphabet end
 
+typealias NucleotideAlphabets Union{DNAAlphabet,RNAAlphabet}
```
bicycle1885: I think type names should be singular.

src/seq/bioseq.jl
```diff
@@ -1100,3 +1098,130 @@ function seqmatrix{A<:Alphabet}(vseq::AbstractVector{BioSequence{A}}, major::Sym
         throw(ArgumentError("major must be :site or :seq"))
     end
 end
+
+"""
+    seqmatrix{A<:Alphabet,T}(vseq::Vector{BioSequence{A}}, major::Symbol, conv::Type{T})
```
bicycle1885: The type argument should come first (c.f. `convert`, `parse`, `zeros`, etc.). Also, `conv` will not be used and can be safely removed.

src/seq/bioseq.jl
```diff
@@ -1100,3 +1098,130 @@ function seqmatrix{A<:Alphabet}(vseq::AbstractVector{BioSequence{A}}, major::Sym
...
+
+Construct a matrix of T from a vector of `BioSequence`s.
```
bicycle1885: Backquote around `T`.

src/seq/bioseq.jl
```diff
@@ -1100,3 +1098,130 @@ function seqmatrix{A<:Alphabet}(vseq::AbstractVector{BioSequence{A}}, major::Sym
...
+
+If parameter major is set to `:site`, the matrix is created such that one
```
bicycle1885: Backquote around `major` since it is an argument of the function.

src/seq/bioseq.jl
```diff
@@ -1100,3 +1098,130 @@ function seqmatrix{A<:Alphabet}(vseq::AbstractVector{BioSequence{A}}, major::Sym
...
+
+julia> seqmatrix(seqs, :seq, UInt8)
+3×4 Array{UInt8,2}:
+ 0x01  0x08  0x02  0x04
+ 0x01  0x08  0x02  0x04
+ 0x01  0x08  0x02  0x04
+```
+"""
+function seqmatrix{A<:Alphabet,T}(vseq::AbstractVector{BioSequence{A}}, major::Symbol, conv::Type{T})
+    nsites = length(vseq[1])
```
bicycle1885: This will throw a bounds error when `vseq` is empty. This case should be checked.

test/seq/runtests.jl
```diff
@@ -3200,4 +3217,15 @@ end
...
+@testset "majorityvote" begin
+    dnas = [dna"CTCGATCGATCC", dna"CTCGAAAAATCA", dna"ATCGAAAAATCG", dna"ATCGGGGGATCG"]
+    dna2 = [dna"CTCGATCGATCC", dna"CTCGAAA", dna"ATCGAAAAATCG", dna"ATCGGGGGATCG"]
+    rnas = [rna"CUCGAUCGAUCC", rna"CUCGAAAAAUCA", rna"AUCGAAAAAUCG", rna"AUCGGGGGAUCG"]
+    rna2 = [rna"CUCGAUCGAUCC", rna"CUCGAAAAAUCA", rna"AUCGAAAAAUCG", rna"AUCGUCG"]
+    @test majorityvote(dnas) == dna"MTCGAAARATCG"
+    @test majorityvote(rnas) == rna"MUCGAAARAUCG"
+    @test_throws ArgumentError majorityvote(dna2)
+    @test_throws ArgumentError majorityvote(rna2)
+end
```
bicycle1885: I think you need to test a few pathological cases (e.g. `DNASequence[]`, `[dna"", dna""]`)

src/seq/bioseq.jl
```diff
@@ -1100,3 +1098,130 @@ function seqmatrix{A<:Alphabet}(vseq::AbstractVector{BioSequence{A}}, major::Sym
...
+function majorityvote{A<:NucleotideAlphabets}(seqs::AbstractVector{BioSequence{A}})
+    mat = seqmatrix(seqs, :site, UInt8)
+    nsites = size(mat, 2)
+    nseqs = size(mat, 1)
+    result = BioSequence{A}(nsites)
+    @inbounds for site in 1:nsites
+        votes = zeros(Int, 16)
+        for seq in 1:nseqs
+            nuc = mat[seq, site]
+            votes[1] += nuc == 0x00
```
bicycle1885: How should gap symbols be handled? If the gap symbol does no vote, I believe this line is not necessary.

Ward9250: By intent and design, gap symbols vote for a gap.

Ward9250:
@bicycle1885 Implemented all of your suggestions :+1:

src/seq/bioseq.jl
```diff
@@ -1102,18 +1103,18 @@ end
 """
     seqmatrix{A<:Alphabet,T}(vseq::Vector{BioSequence{A}}, major::Symbol, conv::Type{T})
```
bicycle1885: The change of the arguments' order is not reflected here.

Ward9250: Good catch - fixed.

bicycle1885:
Why did coveralls run so many times?

Ward9250:
I'm not sure, but it could be because I did --amends to commits and force pushed them to the PR rather than make many commits.

Ward9250:
Am I good to merge on tests pass?

bicycle1885:
Yes, please.

Ward9250:
Awesome, thanks!

PR #220, opened 2016-07-16
Julia v0.5

bicycle1885:
This is a pull request to make a step forward to the next Julia v0.5 and ultimately drop the support of the current Julia v0.4.

At the moment, to pass all tests, you will need to
* <del>build the latest master branch of Julia</del> Julia 0.5 RC2, and
* <del>clone the latest master branch of <del>Libz.jl</del> (use v0.2) and LightXML.jl</del> install the lastest Libz.jl and LightXML.jl

PR #280, opened 2016-08-25
fix Julia version requirement

bicycle1885:
I forgot to fix the requirement of Julia version in #220 .


Ward9250:
:+1:

bicycle1885:
Travis CI doesn't start to run but this change would be harmless.

PR #281, opened 2016-08-25
fix Bio.Util docs

bicycle1885:


kdmurray91:
Ah, whoops. Looks like I forgot to fix this up before we merged the PR. :+1:

bicycle1885:
Never mind. I overlooked it while reviewing, too.

bicycle1885:
Travis CI doesn't start a job on OS X but it seems to be a technical problem of Travis's side. I'm going to merge this because it already passes on Linux and Windows.

Issue #282, opened 2016-08-25
BioArrays - Containing multiple BioSequences 


Ward9250:
Currently in Bio.jl, if you want to do something with multiple sequences you have to either iterate over a vector of sequences and do each op on each individual sequence. Or worse if you are like me and often want to do operations site by site, where the same site in every sequence is examined in turn, then you have to employ double loops or smarter iterator use.

I'd like to propose a new type or a modification of BioSequence, that allows storing of multiple biological sequences as a matrix. i.e Think of BioSequence is an alias for BioArray{A,1}, where BioArray{A,2} allows you to store multiple sequences in a contiguous block of memory in site-major, or sequence major order (maybe a type parameter can denote site-major and sequence major order). 

@bicycle1885 @kdmurray91 What are your thoughts? Personally I think this would be a huge benefit to the kind of algorithms and operations that are common in my kind of field, and popular genotype file formats like PED and BED (Binary PED, not to be confused with the other BED format) would require some kind of type like this anyway in the future, it would be amazing if we could massage BioSequence to do this so as it just works for sequences, BED and anything else that ever wants a set of symbols in either site major or sequence major order.

On the face of it I can imagine it working, it would require the data field of BioSequence to be an Array{Uint64,N} instead of Vector{UInt64} and then some modifications to the getting and setting related code. Obviously there will be other things to resolve.

bicycle1885:
Using the usual matrix would be sufficient in most cases. Are there any benefits defining such type?

Ward9250:
You wouldn't get the compression and fast bit operations with a normal matrix. The benefits lie in having multiple sequences stored in the same chunk of contiguous memory, you could take advantage of BioSequence bitwise operations, not only across one sequence, but also across sites. 

One concrete example where this would assist people interested in population genetics and genetic variance, would be the genetic distance and mismatch counting algorithms. Currently the `distance(::Vector{BioSequence})` method executes `distance(::BioSequence, ::BioSequence)` repeatedly for every pair of sequences, and this itself contains loops. So you're looping about quite a lot and different sequences are located in different areas of memory. If there were a 2d type that was site-major in its order, the counting of mutations and computation of distance can be done in one method a-la the dna.dist C code in ape (https://github.com/cran/ape/blob/master/src/dist_dna.c).

bicycle1885:
[...]

PR #283, opened 2016-08-25
Added sitemajor, seqmajor, docs, and tests

Ward9250:
Following from discussion on #282, I've implemented methods for the creation of sequence-major and site-major matrices. Very useful for computing on multiple sequences sometimes.

### DIFF
--- a/docs/src/man/seq.md
+++ b/docs/src/man/seq.md
@@ -463,6 +463,8 @@ complement
 reverse_complement
 mismatches
 composition
+seqmajor
+sitemajor
 ```
 
 
--- a/docs/src/man/seq.md
+++ b/docs/src/man/seq.md
@@ -463,8 +463,7 @@ complement
 reverse_complement
 mismatches
 composition
-seqmajor
-sitemajor
+seqmatrix
 ```
 
 
--- a/src/seq/bioseq.jl
+++ b/src/seq/bioseq.jl
@@ -1033,54 +1033,19 @@ end
 # -------------------
 
 """
-    sitemajor{A<:Alphabet}(vseq::Vector{BioSequence{A}})
+    seqmatrix{A<:Alphabet}(vseq::Vector{BioSequence{A}}, major = :seq|:site)
 
 Construct a matrix of nucleotides or amino acids from a BioSequence.
-The matrix is created such that one nucleotide from each sequence is placed in
-each column i.e. the matrix is laid out in site-major order.
 
+If parameter major is set to `:site`, the matrix is created such that one
+nucleotide from each sequence is placed in each column i.e. the matrix is laid
+out in site-major order.
 This means that iteration over one position of many sequences is efficient,
 as julia arrays are laid out in column major order.
 
-# Examples
-```julia
-julia> seqs = [dna"AAAAAAAAAA", dna"TTTTTTTTTT", dna"CCCCCCCCCC", dna"GGGGGGGGGG"]
-4-element Array{Bio.Seq.BioSequence{Bio.Seq.DNAAlphabet{4}},1}:
- 10nt DNA Sequence:
-AAAAAAAAAA
- 10nt DNA Sequence:
-TTTTTTTTTT
- 10nt DNA Sequence:
-CCCCCCCCCC
- 10nt DNA Sequence:
-GGGGGGGGGG
-
-julia> sitemajor(seqs)
-4x10 Array{Bio.Seq.DNANucleotide,2}:
- DNA_A  DNA_A  DNA_A  DNA_A  DNA_A  DNA_A  DNA_A  DNA_A  DNA_A  DNA_A
- DNA_T  DNA_T  DNA_T  DNA_T  DNA_T  DNA_T  DNA_T  DNA_T  DNA_T  DNA_T
- DNA_C  DNA_C  DNA_C  DNA_C  DNA_C  DNA_C  DNA_C  DNA_C  DNA_C  DNA_C
- DNA_G  DNA_G  DNA_G  DNA_G  DNA_G  DNA_G  DNA_G  DNA_G  DNA_G  DNA_G
-
-```
-"""
-function sitemajor{A<:Alphabet}(vseq::Vector{BioSequence{A}})
-    nsites = minimum([length(seq) for seq in vseq])
-    nseqs = length(vseq)
-    mat = Matrix{eltype(A)}(nseqs, nsites)
-    for seq in 1:nseqs, site in 1:nsites
-        mat[seq, site] = vseq[seq][site]
-    end
-    return mat
-end
-
-"""
-    seqmajor{A<:Alphabet}(vseq::Vector{BioSequence{A}})
-
-Construct a matrix of nucleotides or amino acids from a BioSequence.
-The matrix is created such that each sequence is placed in
-each column i.e. the matrix is laid out in sequence-major order.
-
+If the parameter major is set to `:seq`, the matrix is created such that each
+sequence is placed in one column i.e. the matrix is laid out in sequence-major
+order.
 This means that iteration over each sequence is efficient,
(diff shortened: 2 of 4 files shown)


bicycle1885:
Just a suggestion: why not switch the axis according to an option? `seqmatrix(seqs, major=:site|:seq)` looks better to me.

Ward9250:
I like that idea! Lemme just do another commit

src/seq/bioseq.jl
```diff
@@ -1027,3 +1027,80 @@ end
...
+
+If the parameter major is set to `:seq`, the matrix is created such that each
+sequence is placed in one column i.e. the matrix is laid out in sequence-major
+order.
+This means that iteration over each sequence is efficient,
+as julia arrays are laid out in column major order.
+
+# Examples
+```julia
+julia> seqs = [dna"AAAAAAAAAA", dna"TTTTTTTTTT", dna"CCCCCCCCCC", dna"GGGGGGGGGG"]
```
bicycle1885: The example sequences look too long and will fill up my terminal when I see the help. For example, two sequences of length 3 would be enough to catch.

src/seq/bioseq.jl
```diff
@@ -1027,3 +1027,80 @@ end
 function enc64{A}(::BioSequence{A}, x)
     return UInt64(encode(A, convert(eltype(A), x)))
 end
+
+
+# Sequences to Matrix
+# -------------------
+
+"""
+    seqmatrix{A<:Alphabet}(vseq::Vector{BioSequence{A}}, major = :seq|:site)
```
bicycle1885: Inserting spaces around `=` is not so common in Julia. See `split` help for example.

src/seq/bioseq.jl
```diff
@@ -1027,3 +1027,80 @@ end
...
+
+Construct a matrix of nucleotides or amino acids from a BioSequence.
```
bicycle1885: Backquote around `BioSequence`. Also, it should be plural.

src/seq/bioseq.jl
```diff
@@ -1027,3 +1027,80 @@ end
...
+ 10nt DNA Sequence:
+AAAAAAAAAA
+ 10nt DNA Sequence:
+TTTTTTTTTT
+ 10nt DNA Sequence:
+CCCCCCCCCC
+ 10nt DNA Sequence:
+GGGGGGGGGG
+
+julia> seqmatrix(seqs, major = :site)
```
bicycle1885: `seqmatrix` doesn't accept any keyword argument.

src/seq/bioseq.jl
Ward9250: Ok fixed to length = 3 (a few seqs are slightly longer than length 3, but that's intentional to test the truncating behaviour of the function - similar to how `mismatches` truncates).

Ward9250:
All suggestions implemented, ready to merge.

docs/src/man/seq.md
```diff
@@ -463,6 +463,8 @@ complement
 reverse_complement
 mismatches
 composition
+seqmajor
+sitemajor
```
bicycle1885: You left old function names here.

Ward9250: Oh yeah! I'll fix that before merge.

src/seq/bioseq.jl
```diff
@@ -1027,3 +1027,73 @@ end
...
+ DNA_G  DNA_G  DNA_G
+
+ julia> seqmatrix(seqs, :seq)
+ 3x4 Array{Bio.Seq.DNANucleotide,2}:
+  DNA_A  DNA_T  DNA_C  DNA_G
+  DNA_A  DNA_T  DNA_C  DNA_G
+  DNA_A  DNA_T  DNA_C  DNA_G
+```
+"""
+function seqmatrix{A<:Alphabet}(vseq::Vector{BioSequence{A}}, major=:site)
```
bicycle1885: I think you should mention the optional argument `major` in the docstring, or completely remove this default value.

bicycle1885: Also, loosening the type constraint of `vseq` as `AbstractVector{BioSequence{A}}` may be good.

Ward9250: I thought I mention it above:
"If parameter major is set to `:site`, the matrix is created such that one
 +nucleotide from each sequence is placed in each column i.e. the matrix is laid
 +out in site-major order.
 +This means that iteration over one position of many sequences is efficient,
 +as julia arrays are laid out in column major order.
 +
 +If the parameter major is set to `:seq`, the matrix is created such that each
 +sequence is placed in one column i.e. the matrix is laid out in sequence-major
 +order.
 +This means that iteration over each sequence is efficient,
 +as julia arrays are laid out in column major order."

Oops, I was eager beaver to merge, I can edit any doctrings in some direct commits, if it needs editing.

src/seq/bioseq.jl
```diff
@@ -1027,3 +1027,73 @@ end
...
+    if major == :site
+        nsites = minimum([length(seq) for seq in vseq])
```
bicycle1885: Should this function work for sequences of different length? I think it should check that every sequence has the same length and otherwise should throw an exception.

test/seq/runtests.jl
```diff
@@ -782,6 +782,54 @@ end
...
     end
 
+    @testset "Conversion to Matrices" begin
+        dna = [dna"AAA", dna"TTTAAA", dna"CCC", dna"GGG"]
+        rna = [rna"AAA", rna"UUU", rna"CCCUUU", rna"GGG"]
+        prot = [aa"AMG", aa"AMG", aa"AMG", aa"AMG"]
+        sitemajdna = [
+            DNA_A  DNA_A  DNA_A;
+            DNA_T  DNA_T  DNA_T;
+            DNA_C  DNA_C  DNA_C;
```
bicycle1885: Tips: you don't need to append a semicolon for each line in a matrix.

src/seq/bioseq.jl
Ward9250: Sorry @bicycle1885 I was eager to merge so I could do work stuff. I'll make any other edits in another pr.

That is one option, I did consider this, but I went with the truncation solely because that's what I believe `mismatches` does and I wanted to have the behaviour consistent.

bicycle1885: At least, this behavior should be mentioned in the docstring.

test/seq/runtests.jl
Ward9250: That's pretty cool! I thought that was necessary.

src/seq/bioseq.jl
bicycle1885: I mean the default argument of `major`: will `seqmatrix(seqs)` be site-major or seq-major? It is not mentioned in the docstring.

src/seq/bioseq.jl
bicycle1885: Other functions in Julia will check the length of arguments:
```
julia> hcat([1,2,3], [4,5,6,7])
ERROR: DimensionMismatch("vectors must have same lengths")
 in hcat(::Array{Int64,1}, ::Array{Int64,1}) at ./array.jl:735

```

I believe the `mismatches` function is inconsistent.

Ward9250: I'll make it more strict and require sequences of equal length - I think the more predictable behaviour the better, otherwise if someone doesen't remember to check in their calling method or script, depending on what they're doing they might get burned.

test/seq/runtests.jl
bicycle1885: A newline and a semicolon are interchangeable. So semicolons are redundant when you break rows with newlines.

src/seq/bioseq.jl
bicycle1885: 👍

PR #284, opened 2016-08-26
Extra refinement to `seqmatrix`

Ward9250:
@bicycle1885 - Implemented your post merge suggestions :+1: 
* Added sequence length checking behaviour to `seqmatrix`.
* Added tests of throw cases.
* Loosening the type constraint of `vseq` to `AbstractVector{BioSequence{A}}`.
* `major` is now just a parameter, not an optional one.

### DIFF
--- a/src/seq/bioseq.jl
+++ b/src/seq/bioseq.jl
@@ -1033,7 +1033,7 @@ end
 # -------------------
 
 """
-    seqmatrix{A<:Alphabet}(vseq::Vector{BioSequence{A}}, major=:seq|:site)
+    seqmatrix{A<:Alphabet}(vseq::Vector{BioSequence{A}}, major::Symbol)
 
 Construct a matrix of nucleotides or amino acids from a vector of `BioSequence`s.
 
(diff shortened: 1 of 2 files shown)


src/seq/bioseq.jl
```diff
@@ -1076,20 +1076,23 @@ julia> seqmatrix(seqs, :site)
   DNA_A  DNA_T  DNA_C  DNA_G
 ```
 """
-function seqmatrix{A<:Alphabet}(vseq::Vector{BioSequence{A}}, major=:site)
+function seqmatrix{A<:Alphabet}(vseq::AbstractVector{BioSequence{A}}, major::Symbol)
+    l = length(vseq[1])
+    nseqs = length(vseq)
+    @inbounds for i in 2:nseqs
+        length(vseq[i]) != l | throw(ArgumentError("Sequences in vseq must be of same length"))
```
bicycle1885: `|` is not a short-circuit operator. You should use `||` instead.

Ward9250: :O Oh yes that's a typing mistake, I meant to write ||

Ward9250:
@bicycle1885 Just removed the redundant semi-colons. I think that all of your remaining suggestions from the previous PR implemented now :+1:

PR #286, opened 2016-08-26
add bit-op docs for nucleotides

bicycle1885:


docs/src/man/seq.md
```diff
@@ -91,6 +91,45 @@ true
 
 ```
 
+Every nucleotide is encoded using lower 4 bits of a byte. An unambiguous
+nucleotide has only one *true* bit and the other bits are *false*s. The table
```
Ward9250: "using lower 4 bits" <- "using the lower 4 bits".

Also should it be "has only one set bit" & "the other bits are unset"? I'm not sure which is more intuitive to less computer savvy biologists,

docs/src/man/seq.md
```diff
@@ -91,6 +91,45 @@ true
...
+below summarizes all unambiguous nucleotides and their corresponding bits. An
+ambiguous nucleotide is the bitwise OR of unambiguous nucleotides that the
+ambiguous nucleotide can take. For example, `DNA_R`--`DNA_A` or `DNA_G`--is
```
Ward9250: Suggestion: "`DNA_R` (meaning the nucleotide is either `DNA_A` *or* `DNA_G`)" ?

Ward9250:
Two minor comments for what is written but mostly just suggestions - looks good!

docs/src/man/seq.md
bicycle1885: Thanks again!

That sounds more intuitive considering bit flags.

docs/src/man/seq.md
bicycle1885: 👍

PR #287, opened 2016-08-28
Improvements to mutation counting.

Ward9250:
Improvements on mutation counting for vectors of sequences.

- [x] Implement improved pairwise distance methods for vectors of BioSequence.
    - [x] Count.
    - [x] Proportion.
    - [x] JukesCantor69.
    - [x] Kimura80.
    - [x] Tests.
    - [x] Documentation.
- [x] Implement improved pairwise mutation counting methods for vectors of BioSequence.
    - [x] DifferentMutation.
    - [x] TransitionMutation.
    - [x] TransversionMutation.
    - [x] TransitionMutation & TransversionMutation.
    - [x] Tests.
    - [x] Documentation.


Ward9250:
@bicycle1885 I'm noticing many identical coveralls comments again. I've opened an issue at coveralls about this to ask if it's the service or if it's our travis configuration that's reponsible.

Ward9250:
@bicycle1885 @kdmurray91 Ready to have people take a look at this now. 

One thing I was wondering, maybe it could be a separate PR, you'll notice some methods make assumptions about the orientation of any matrices produced by `seqmatrix`. This is clearly indicated in their doc-strings. But I was wondering if this could be dealt with by implementing a very lightweight type wrapping Matrix, which is parametric over an `Orientation` type of `SiteMajor` or `SequenceMajor` - eliminating the need for runtime checks of orientation, and eliminating the potential problems of incorrect use of these methods that have assumptions of matrix arrangement.

src/seq/nucleotide.jl
```diff
@@ -97,27 +97,27 @@ end
 
 Test if nucleotide is surely a purine.
 """
-function ispurine(nt::Nucleotide)
+@inline function ispurine(nt::Nucleotide)
     bits = reinterpret(UInt8, nt)
-    return bits != 0 && (bits & 0b1010) == 0
+    return (bits != 0) & ((bits & 0b1010) == 0)
```
bicycle1885: Just out of curious: are there any benchmark results that support this change?

Ward9250: I haven't benchmarked, but I inspected the llvm instructions generated, and without it there's no instructions to use the simd capabilities of the processor.

docs/src/man/var.md
```diff
@@ -9,33 +9,19 @@ end
...
+The types of mutations that can currently be counted are `DifferentMutation`,s
+`TransitionMutation`s, and `TransversionMutation`s.
 
-| Type                   | Meaning                                 |
-| :--------------------- | :-------------------------------------- |
-| `DifferentMutation`    | Nucleotides that are different          |
-| `TransitionMutation`   | Transition mutations                    |
-| `TransversionMutation` | Transversion mutations                  |
-
-`DifferentMutation` simply compares two nucleotides to see if they can be deemed
```
kdmurray91: Bikeshedding: how about "AnyMutation" rather than "DifferentMutation"?

src/seq/nucleotide.jl
kdmurray91: If easy, :+1: to a benchmark, but that sounds right @Ward9250

src/var/distances.jl
```diff
@@ -98,71 +108,170 @@ end
...
+    return count_mutations(T, seqs)
+end
+
+"""
+    distance{T<:MutationType,N<:Nucleotide}(::Type{Count{T}}, seqs::Matrix{N})
+
+Compute the number of mutations of type `T` between a set of sequences in a
+pairwise manner.
 
 This method of distance returns a tuple of the number of mutations of type T
```
kdmurray91: Pedantry: text wrapping here needs a `gq` in vim :)

src/var/mutation_counting.jl
```diff
@@ -15,10 +15,43 @@
...
 immutable TransitionMutation <: MutationType end
-immutable TransversionMutation <: MutationType end
 
+"""
+`TransversionMutation` describes a situation with two aligned nucleotides,
+where a purine has mutated into a pyrimadine or vice versa.
+
+Possible transversion mutations are:
+A -> T
+T -> A
```
kdmurray91: Pedantry: Could be shortened to `A <-> T` and so on?

src/var/mutation_counting.jl
```diff
@@ -27,128 +60,168 @@ The strictest test of ambiguity, if either nucleotide is not
 A, T, G, or C, then this function returns true.
 """
 @inline function is_ambiguous_strict{T<:Nucleotide}(a::T, b::T)
-    return isambiguous(a) || isambiguous(b)
+    return isambiguous(a) | isambiguous(b)
```
kdmurray91: I assume this is intentional, in which case perhaps document why bitwise rather than logical?

Ward9250: It is intentional. The reason is in julia the logical operators using double symbols e.g. "||" have short-circuit behaviour, which when used in loops can prevent the julia compiler from emitting simd instructions for that loop. To ensure straight-line code that can be vectorized using simd, the `|` version of the operator is used which ensures both sides of the argument are evaluated first.

src/var/mutation_counting.jl
Ward9250: Sure!

docs/src/man/var.md
Ward9250: OOOh I like that for an alternate name!

Ward9250: Changed DifferentMutation to `AnyMutation`

src/seq/nucleotide.jl
Ward9250: @bicycle1885 @kdmurray91 

I've just run some benchmarks on my work machine, and I've found actually there's no difference in speed with the change of logical operators, and that both produce simd llvm instructions.

My current explanation for this is - as it states in the julia manual, the compiler can sometimes turn short `&&`, `||`, and `?:` expressions into straight-line code. Which is probably what is happening here. However on my home laptop, simd llvm instructions were not emitted, perhaps the compiler on my home machine didn't turn them into straight line code, and so I didn't get use of simd in the loops.

So ultimately I guess it depends on what we decide. Currently is see no detriment to performance with this change on one laptop, and yet I saw a gain to performance on my home laptop when I wrote this a few days ago. So if this change guarantees simdable code whereas the short circuit operators only sometimes get turned into simdable code, I'm all for this change on the balance of things. But I'm going to repeat the benchmarks I've done on my home machine again, to make absolutely sure.

bicycle1885: If no performance gain is detectable, I feel like keeping it as is to minimize the difference from the master. Changing it may send a wrong signal to readers that this change is based on evidence. We can replace them with non-short-circuit operators anytime once benchmark supports that.

Ward9250: I've just checked on my home laptop and I'm seeing simd instructions and equal performance with the short-circuit operations now. The only other explanation is I've recently updated a load of software this week, including R and julia, and perhaps I started this work on one version and moved to another and that explains the difference. Either way, we get simd without changing those operations now! I'll revert.

Ward9250:
Following previous discussion and checking my home laptop, I've reverted the functions in src/seq/nucleotide.jl back to their short-circuit form and am ready to merge tomorrow as tests pass and no further objections arise.

kdmurray91:
@Ward9250 What was the reason for reverting to short circut ops? Were the streamlined ops slower in a benchmark. (sorry if you answered this above, and I lost that in the torrent of coveralls spam).

There was what looked like an intermittent failure in travis for OSX, so I re-started the build.

Ward9250:
The reason to change to non-short-circuit ops was that on my home laptop it was required to get vectorised llvm instructions. But then after the change was queried I did benchmarks on my work machine and saw no difference between short circuit ops and non-short-circuit ops, and inspecting the llvm instructions saw that the compiler managed to generate straight-line code from the short circuit ops, which is something the documentation says the compiler *can* do, but not always. So I went home and tried again and found on my home machine it was also managed to make straight-line code with the short circuit ops and so the change could be reverted, as it was no longer nessecery. I recently changed versions of quite a few things on my machine including R and julia, and that's my current working explanation as to why it didn't vectorize before.

Issue #288, opened 2016-08-30
IO interface cleanup


kdmurray91:
In several threads across issues and gitter recently @Ward9250, @bicycle1885 and I have discussed improvements that could be made to the design and consistency of the IO interfaces of parsers and writers. A rough summary of these ideas follows.
- [x] Change open argument order such that types come first (@vchuravy)
- [ ] Supporting `open(::Type{FMT}, io::IO)` 
- [x] Defining a `stream{T<:AbstractParser}(parser::T)` function to retrieve the underlying IO object (or BufferedIOStream, or BGZFStream etc) from a parser or writer. EDIT: Done in #292 
- [x] Using the `stream(parser)` function to automatically define standard IOStream operations like `eof()`, `flush()`, `close()`, `write()`, etc. in a unified way. EDIT: Done in #292 
- [x] Parser -> Reader: Done in #290

bicycle1885:
Sorry for wandering off the topic a little bit, but I'd like to know how you feel about renaming `XXXParser`s to `XXXReader`s. Since we have *writer*s to serialize data into a formatted file, I think having corresponding *reader*s will be more symmetric.

vchuravy:
How about using `open(::Type, filename)` as an interface?

kdmurray91:
@vchuravy Why? I'm quite new to Julia, is that a more common pattern? I guess that is how one does e.g. `zeros(Int8, (2, 3))`.

Ward9250:
:+1: for Readers.

bicycle1885:
> Allowing open() do file; ...; end; style "context managers" (excuse the python) by defining open(fn::Function, args..., ::Type{FMT})

I think the last item is not needed because `Base.open` already implements a `open(f::Function, args...)` method.

bicycle1885:
One more proposal I have in my mind for a long time is removing the format types (e.g. `FASTA`) and using reader/writer types directly or indirectly by wrapping it with `open`. Taking @vchuravy's suggestion into account, creating a reader object looks like:
```julia
# direct call
reader = FASTAReader("hg38.fa")
# indirect call
reader = open(FASTAReader, "hg38.fa")
```

Ward9250:
I think that would be ok.

vchuravy:
@kdmurray91 @bicycle1885 

The style `f(::Type, args...)` is as far as I know a quite common pattern and the closet cousin would be `parse(::Type, String)` in Base.

kdmurray91:
[...]

PR #292, opened 2016-08-30
add the Bio.IO.stream function

bicycle1885:
This implements the `Bio.IO.stream` function, which fixes a part of #288.

@kdmurray91, can you check if this solves the issue you mentioend?

### DIFF
--- a/src/IO.jl
+++ b/src/IO.jl
@@ -8,11 +8,6 @@
 
 module IO
 
-export
-    FileFormat,
-    AbstractReader,
-    AbstractWriter
-
 """
 Abstract file format type.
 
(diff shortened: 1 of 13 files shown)


Ward9250:
:+1:

Ward9250:
Do you find any benefit to @inline-ing the stream function? So as accessing the field just get's inlined into calling code?

bicycle1885:
I guess no benefit because `stream` will not be called very often.

kdmurray91:
@bicycle1885 Beautiful, I had an attempt to do this in the [parser refactor](https://github.com/BioJulia/Bio.jl/tree/parser-refactor) branch, and unsurprisingly yours works far better.

bicycle1885:
@kdmurray91 Thank you :) I should have said you before starting my work.

I'm going to merge this if CI says okay.

PR #265, opened 2016-08-19
BAM and SAM formats

bicycle1885:
This supersedes #206 and #101 and implements parsers for BAM and SAM file formats on top of #220 . The implementation moved from https://github.com/bicycle1885/HTSFileFormats.jl is almost finished, but I still need to add some tests and solve a problem about the SAM header parser.

PR #293, opened 2016-08-30
implement open(::IO, BAM)

bicycle1885:
This implements a missing piece of #265, and also is a part of #288.


Ward9250:
:+1:

kdmurray91:
Would be great if there were a test for this.

bicycle1885:
Added.

bicycle1885:
I think it's okay to merge.

PR #294, opened 2016-08-31
add alignment(::SAM/BAMRecord)

bicycle1885:
This implements an accessor `alignment` for SAM/BAM records. The diff may look huge but it is just because I relocated some code into separated files.


Ward9250:
LGTM

PR #295, opened 2016-08-31
Initial implementation of SequenceMatrix

Ward9250:
Following my comment on #287 here's a draft of the lightweight type mentioned, to contain the nucleotide matrices in a site-major or sequence-major orientation and allow method dispatch on the orientation.

### DIFF
--- a/src/var/seq_matrix.jl
+++ b/src/var/seq_matrix.jl
@@ -1,0 +1,46 @@
+abstract Orientation
+
+immutable SequenceMajor <:Orientation end
+immutable SiteMajor <: Orientation end
+
+immutable SequenceMatrix{T,O<:Orientation}
+    data::Matrix{T}
+end
+
+function SequenceMatrix{T,A<:Alphabet}(::Type{T}, ::Type{SequenceMajor}, vseq::Vector{BioSequence{A}})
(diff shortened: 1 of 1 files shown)


Ward9250:
Improvements to Var module are taking a different direction in #334 so I'm going to close this, and if it's needed again, then re-open.

bicycle1885:
Please delete the branch if you're sure you no longer need it. Also, ca you make new branches on your own fork repository when creating pull requests? That will make the list of branches on BioJulia/Bio.jl much cleaner.

Ward9250:
Ok then, I'll add a rule to the contribution guidelines:

From now, new branches created by anybody have to be created on that person's own fork, with the exception of release branches or large milestone branches.

I think that sounds reasonable.

Ward9250:
Ok I've added that guidance to https://github.com/BioJulia/Contributing so that should be live online when the mkdocs run completes. Some existing branches deleted, and I'll work my way through the ones left.

bicycle1885:
Thanks a lot!

Ward9250:
There's probably a way to synch my fork with BioJulia, and afterwards just move the branches across to the fork and then remove them from BioJulia.

PR #297, opened 2016-08-31
rename position accessors of alignment records

bicycle1885:
This pull request changes the name of accessors to get the leftmost mapping position. `position(record)` looks a little bit vague to me so I think `leftposition(record)` is a better name. Also, this adds `rightposition` accessor to get the rightmost mapping position, which would be very useful when handling HTS alignment data.


Ward9250:
Aways in favour of reduction of ambiguity! 👍

PR #298, opened 2016-09-01
unify methods definitions of seqname, sequence and metadata

bicycle1885:
This cleans up method definitions of `seqname`, `sequence` and `metadata`. These methods are now defined in `Bio.Seq` and other modules extend them. I will update docs soon.


bicycle1885:
I will merge this soon if no objections.

kdmurray91:
LGTM

PR #299, opened 2016-09-01
add left/rightposition accessors to Interval

bicycle1885:
This adds `leftposition` and `rightposition` accessor functions to the `Interval` type. Currently, we extend the `Base.first` and `Base.last` functions to access left/rightmost position of an interval. However, `Base.first` and `Base.last` already have well-defined semantics in the standard library and I think we should not abuse them for the interval types. Along with #297, we can always use the `leftposition` and `rightposition` functions for interval-like objects.

### DIFF
--- a/src/intervals/interval.jl
+++ b/src/intervals/interval.jl
@@ -42,6 +42,24 @@ end
 seqname(i::Interval) = i.seqname
 strand(i::Interval) = i.strand
 
+"""
+    leftposition(i::Interval)
+
+Return the leftmost position of `i`.
+"""
+function leftposition(i::Interval)
+    return i.first
+end
+
+"""
+    rightposition(i::Interval)
+
+Return the rightmost position of `i`.
+"""
+function rightposition(i::Interval)
+    return i.last
+end
+
 IntervalTrees.first(i::Interval) = i.first
 IntervalTrees.last(i::Interval) = i.last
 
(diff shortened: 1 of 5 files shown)


src/intervals/interval.jl
```diff
@@ -100,7 +118,16 @@ function isoverlapping{S, T}(a::Interval{S}, b::Interval{T})
...
-    print(io, i.seqname, ":", i.first, "-", i.last, "    ", i.strand, "    ", i.metadata)
+    if get(io, :compact, false)
+        print(io, i.seqname, ":", i.first, "-", i.last, "  ", i.strand, "  ", i.metadata)
+    else
+        println(io, summary(i), ':')
+        println(io, "  sequence name: ", i.seqname)
+        println(io, "  leftmost position: ", i.first)
+        println(io, "  rightmost position: ", i.last)
+        println(io, "  strand: ", i.strand)
+          print(io, "  metadata: ", i.metadata)
```
kdmurray91: Appears to be some indenting issue here

kdmurray91:
:+1:

Looks good to me, except that one indenting fix @bicycle1885

src/intervals/interval.jl
bicycle1885: This is intentional; the variable `io` and the string literal are beautifully aligned to the lines above 😁

kdmurray91:
Ah, that makes more sense!

Issue #289, opened 2016-08-30
Common makefile for all ragel parsers; ragel in CI.


kdmurray91:
As mentioned in a comment in #265, it would be useful to unify all the `Makefile`s that generate julia code from ragel sources into a single makefile under `./src`. While we're at it, it would be great if we could run this makefile in Travis CI (at least when master is updated) to ensure that julia sources are not changed by re-running ragel, i.e. that we don't implement things directly in the generated source code.

I will do this has I have time, unless someone beats me to it (in which case please un-assign me from the issue).

Cheers,
K

kdmurray91:
Closed in #300

PR #300, opened 2016-09-01
move parsers into separated directories

bicycle1885:
This relocates parser files into directories (for example, FASTA related files are placed in the `src/seq/fasta` directory) and replaces all `Makefile`s with a single `src/Makefile` (#289). After this change, I'm going to tackle #288.


kdmurray91:
LGTM!

PR #303, opened 2016-09-02
Rewrite the FASTQ reader

bicycle1885:
This giant rewrite will improve the FASTQ reader and clean up quality encoding code.

* `QualityEnoding` is now a structure type, not bit flag.
* `FASTQReader` and `FASTQWriter` takes an argument `quality_encoding` to determine the encoding.
    * The default value of `quality_ecoding` is `:sanger`.
    * Encoding inference is completely removed.
    * `FASTQReader` checks whether all base qualities are within the range of `quality_encoding`.
* `FASTQReader` supports `fill_ambiguous` argument to fill ambiguous nucleotides with a given nucleotide (say `DNAA`), which will meet @timbitz's request.
* Update docs.


src/seq/fastq/reader.jl
```diff
@@ -34,10 +44,28 @@ function Base.eltype{S}(::Type{FASTQReader{S}})
...
+    elseif name == :solexa
+        return SOLEXA_QUAL_ENCODING
+    elseif name == :illumina13
+        return ILLUMINA13_QUAL_ENCODING
+    elseif name == :illumina15
+        return ILLUMINA15_QUAL_ENCODING
+    elseif name == :illumina18
+        return ILLUMINA18_QUAL_ENCODING
+    end
+    error("quality encoding ':$(name)' is not supported")
```
kdmurray91: Would a const Dict be better? Or are there performance issues with that?

src/seq/fastq/quality.jl
```diff
@@ -6,221 +6,69 @@
...
+            throw(ArgumentError("the ranges of ASCII and base quality don't match"))
+        end
+        return new((first(ascii), last(ascii)), (first(qual), last(qual)))
+    end
 end
 
-function Base.:&(a::QualityEncoding, b::QualityEncoding)
-    return convert(QualityEncoding, convert(UInt16, a) & convert(UInt16, b))
+# Return the offset of converting a base quality to an ASCII code (quality + offset).
+function ascii_offset(encoding::QualityEncoding)
```
kdmurray91: `@inline`?

docs/src/man/reading.md
```diff
@@ -164,16 +168,17 @@ tcagTTAAGATGGGAT
 
 To read a file containing such records, one could use:
 ```julia
-reader = open(FASTQReader, "reads.fastq", Seq.SANGER_QUAL_ENCODING)
-seqrec = eltype(reader)()
-while !eof(reader)
-    read!(reader, seqrec)
-    # ... process sequence
+reader = open(FASTQReader, "reads.fastq")
```
kdmurray91: Would prefer to keep the more performant in-place parser style here. Also, please make clear in the docs that the default qual encoding is `:sanger`

timbitz:
Thanks @bicycle1885!

src/seq/fastq/reader.jl
bicycle1885: A const `Dict` is doable, but I prefer functions because it can handle invalid values and is truly immutable (`Dict` is mutable even if it is defined as a constant).

src/seq/fastq/quality.jl
bicycle1885: I think this function will not be a bottleneck because it can be called outside of a loop.

docs/src/man/reading.md
bicycle1885: Okay, but safety first is a Julian way so I think the copy iteration should be placed first. I added an in-place reading syntax with a comment after that.

bicycle1885:
I will merge this if no further comments.

PR #304, opened 2016-09-02
`flagmutations` for seqmatrixes

Ward9250:
Adds an `is_mutation` method for seqmatrices. 

Produces boolean matrixes indicating which positions are mutated and which are not for every pairwise combination of sequences. 

This is useful for e.g. if you're doing sliding window distance computations or mutation counting computations to find hotspots across sequences, regions of high/low divergence, and so on, as all you need to do is slide across these boolean matrices and so `sum` operations.

### DIFF
--- a/docs/src/man/var.md
+++ b/docs/src/man/var.md
@@ -7,7 +7,7 @@ DocTestSetup = quote
 end
 ```
 
-## Counting mutations
+## Identifying and counting mutations
 
 You can count the numbers of different types of mutations in a pairwise manner
 for a set of nucleotide sequences.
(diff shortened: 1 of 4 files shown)


bicycle1885:
I don't like the name; `isxxx` should return a boolean value not a matrix. Is that impossible to make it return `true`/`false` and call `map(ismutation, seqs)`?

Ward9250:
I think the map thing might be hard, because dosen't map just apply the function to each cell of an array one by one. Whereas this repeatedly scans across two different rows of a sequence matrix and calls the corresponding is_mutation function on two nucleotides. It can always be called something else.

Ward9250:
Tests are passing now so I'd like to merge soon and start on the actual methods I wanted to create that use this internally.

PR #307, opened 2016-09-03
add more docstrings for Bio.Seq

bicycle1885:


bicycle1885:
I've merged since this is just an improvement of docs.

PR #311, opened 2016-09-06
make qualities(::BAMRecord) return a vector of Int8

bicycle1885:
This makes `qualities(::BAMRecord)` return `Vector{Int8}` rather `Vector{UInt8}`. This will improve interoperations between FASTQ and BAM. Also, I believe 127 (= `typemax(Int8)`) is large enough to cover the range of Phred scores.


kdmurray91:
:+1:. Looks goo to me.

bicycle1885:
Thanks 😃

PR #318, opened 2016-09-07
fix ismapped

bicycle1885:
This fixes the implementation of `ismapped` according to the specs of the SAM/BAM file format:
> Bit 0x4 is the only reliable place to tell whether the read is unmapped.


bicycle1885:
Let me merge this since this is an easy fix.

PR #325, opened 2016-09-08
make BAMReader type-stable

bicycle1885:
This simple patch significantly (~15%) improves the performance of reading BAM records.

```julia
using Bio.Align

let
    reader = open(BAMReader, "GSE25840_GSM424320_GM06985_gencode_spliced.bam")
    record = BAMRecord()
    @time while !eof(reader)
        read!(reader, record)
    end
    close(reader)
end

#=
~/.j/v/Bio (master|…) $ julia5 test.jl
 15.774832 seconds (63.40 M allocations: 977.437 MB, 1.33% gc time)
~/.j/v/Bio (master|✚1…) $ julia5 test.jl
 13.658773 seconds (22.08 M allocations: 346.885 MB, 1.59% gc time)
=#
```

### DIFF
--- a/src/align/hts/bam/reader.jl
+++ b/src/align/hts/bam/reader.jl
@@ -32,7 +32,7 @@ function BAMReader(input::IO; index=nothing)
     return reader
 end
 
-function Base.eltype(::Type{BAMReader})
+function Base.eltype{T}(::Type{BAMReader{T}})
     return BAMRecord
 end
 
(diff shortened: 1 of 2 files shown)


kdmurray91:
Out of interest, why such an increase in performance?

bicycle1885:
I guess it is because Julia can avoid searching for the matching `read` method at runtime here: <https://github.com/BioJulia/Bio.jl/blob/4b8afbcd798d6dcb323afaf77ac3eb0309046238/src/align/hts/bam/reader.jl#L118>. Without this type parameter, Julia cannot determine the type of `reader.stream` and hence it needs to find the matching method from a method table based on dynamic types.

PR #334, opened 2016-09-09
Performant mutation counting.

Ward9250:
A PR to improve the current mutation counting functionality of `Bio.Var`.

Currently such methods require transformation of BioSequences into Nucleotide matrices and looping over the matrices and comparing nucleotides. This is analogous to the strategy used in the 8-bit encoding based C code used by the APE package for R. We can do better than this.

This PR will implement bit parallel mutation counting methods that will work on the 2 bit and 4 bit encoded nucleotide sequences in `Bio.Seq`. This will give great speed up as with DNAAlphabet{4}, 16 nucleotides can be considered at once, plus whatever speed up simd will give us looping over each 64bit integer of a biosequence data field. With 2 bit encoded sequences, presumably the speed up will be faster, since A). More nucleotides are considered at once, and B). Masking and checks for ambiguous or gap sites are not nessecery.


ljyanesm:
Hi Ben,
Did we write a benchmark for comparison with the previous method? I'd be nice to include it here.

Ward9250:
We didn't yet, I wanted to wrap up this into some higher level functions and get the tests done first, to make sure the results are actually correct :P .

Ward9250:
Remaining tasks for this work:

* [ ] Move naieve site counting methods into the var module and test them.
* [ ] Test the nibble masking and counting methods.
* [ ] Create optimised counting methods from the tested binary methods and test them.

Ward9250:
#436 has made this PR moot.

PR #339, opened 2016-09-13
Direct people to biojulia.github.io/Contributing/latest

Ward9250:
Most of our guidelines and rules ought to be consistently applied to all of the BioJulia packages. 

This moves to centralise the guidelines and will ensure consistency across the packages.


bicycle1885:
Looks good 👍

PR #344, opened 2016-10-03
Fix a couple BioSequence bugs

dcjones:
I ran into a couple bugs with sequences. Test cases to reproduce and fixes here.

### DIFF
--- a/src/seq/bioseq.jl
+++ b/src/seq/bioseq.jl
@@ -571,15 +571,21 @@ function orphan!{A}(seq::BioSequence{A}, size::Integer=length(seq), force::Bool=
     j, r = bitindex(seq, 1)
     data = Vector{UInt64}(seq_data_len(A, size))
 
-    if !isempty(seq)
+    if !isempty(seq) && !isempty(data)
         x = seq.data[j] >> r
-        l = min(endof(data), index(bitindex(seq, endof(seq))) - j + 1)
+        m = index(bitindex(seq, endof(seq))) - j + 1
+        l = min(endof(data), m)
         @inbounds @simd for i in 1:l-1
             y = seq.data[j+i]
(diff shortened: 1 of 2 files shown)


dcjones:
Ohh don't merge this yet. The second commit breaks a bunch of other tests.

dcjones:
Alright, got it figured out. These are fairly critical fixes so I'll merge later today if no one stops me.

Ward9250:
What was the problem?

dcjones:
Basically it's extremely easy for sub-sequences to get corrupted due to bugs in basic functions.
```julia

julia> seq = dna"ACGTACGTACGT";

julia> subseq = seq[3:6]
4nt DNA Sequence:
GTAC

julia> copy(subseq) # should be the same as subseq
4nt DNA Sequence:
ACGT
```

```julia

julia> seq = repeat(dna"ACGT", 8);

julia> subseq = seq[16:17]
2nt DNA Sequence:
TA

julia> Seq.orphan!(subseq) # should not change the sequence
2nt DNA Sequence:
T-
```

src/seq/bioseq.jl
```diff
@@ -554,7 +554,7 @@ end
 function Base.copy{A}(seq::BioSequence{A})
     # NOTE: no need to set `seq.shared = true` here
     # since `newseq` will be `orphan!`ed soon.
-    newseq = BioSequence{A}(seq.data, 1:endof(seq), true)
+    newseq = BioSequence{A}(seq, 1:endof(seq))
```
bicycle1885: Can you delete the two lines of comments above?

dcjones: Yeah, done.

bicycle1885:
Thank you!

PR #346, opened 2016-10-04
require Julia 0.5 or later

bicycle1885:
Because Julia 0.5 is already released.


kdmurray91:
Should this also be backported?

bicycle1885:
Yes, I think it should be (xref #345).

Issue #348, opened 2016-10-12
inconsistent dna"" - DNASequence() behavior


cjprybol:
This is an example of simulating RNA transcripts (in DNA format) from a fake transcriptome and then randomly making sequencing errors. There are two for loops, the first utilizes `dna""` to create a new, empty dna sequence and in the second for loop I use the `DNASequence()` function.

tracing through the steps with `@show`,  `dna""` doesn't seem to release the object in memory from the prior loop. This causes the `new_seq` variable to be continuously extended during the for loop iterations. `DNASequence()` has the behavior I expected, and assigns an empty sequence to the variable `new_seq` during each for loop.

The key lines are
```
@show new_seq = dna""
```
vs.
```
@show new_seq = DNASequence()
```

Is this intended behavior, or potentially a bug in the parsing of the `dna""` constructor?

**code**
```
using Bio.Seq
using StatsBase

srand(1);
transcriptome = [randdnaseq(rand(8:12)) for i=1:3];
number_of_reads = 5;
answers = sample(transcriptome, number_of_reads);

error = .10;
weights = WeightVec([1-error, error * 1/3, error * 1/3, error * 1/3]);

sequenced_reads_dna = []
for seq in answers
	@show new_seq = dna""
	for nuc in seq
		@show choice = sample(1:4, weights)
		if choice == 1
			@show push!(new_seq, nuc)
		elseif choice == 2
			# deletion, do nothing
			@show new_seq
		elseif choice == 3
			# mutation
			@show new_seq *= randdnaseq(1)
		elseif choice == 4
			# insertion
			@show new_seq *= push!(randdnaseq(1), nuc)
		end
	end
	push!(sequenced_reads_dna, new_seq)
end

sequenced_reads_function = []
for seq in answers
	@show new_seq = DNASequence()
	for nuc in seq
		@show choice = sample(1:4, weights)
		if choice == 1
			@show push!(new_seq, nuc)
		elseif choice == 2
			# deletion, do nothing
			@show new_seq
		elseif choice == 3
			# mutation
			@show new_seq *= randdnaseq(1)
		elseif choice == 4
			# insertion
			@show new_seq *= push!(randdnaseq(1), nuc)
		end
	end
	push!(sequenced_reads_function, new_seq)
end

answers
sequenced_reads_dna
sequenced_reads_function
```

**REPL**
```
               _
   _       _ _(_)_     |  A fresh approach to technical computing
[...]

PR #349, opened 2016-10-12
Fix of Non-standard string literals

Ward9250:
Implements an alteration of non-standard string literals behaviour as highlighted in #348.

### DIFF
--- a/docs/src/man/seq.md
+++ b/docs/src/man/seq.md
@@ -298,14 +298,13 @@ julia> char"αβγδϵ"
 ```
 
 However it should be noted that by default these sequence literals
-allocate the BioSequence object before runtime.
+allocate the `BioSequence` object before the code containing the sequence
+literal is run.
 This means there may be occasions where your program does not behave as you
 first expect, even though it is the intended behaviour.
 For example consider the following code:
 
 ```jlcon
-julia> using Bio.Seq
-
 julia> function foo()
            s = dna"CTT"
            push!(s, DNA_A)
@@ -321,15 +320,15 @@ In other words you might expect the following output:
 
 ```jlcon
 julia> foo()
-1nt DNA Sequence:
+4nt DNA Sequence:
 CTTA
 
 julia> foo()
-2nt DNA Sequence:
+4nt DNA Sequence:
 CTTA
 
 julia> foo()
-3nt DNA Sequence:
+4nt DNA Sequence:
 CTTA
 ```
 
@@ -337,34 +336,35 @@ However, this is not what happens, instead the following happens:
 
 ```jlcon
 julia> foo()
-1nt DNA Sequence:
+4nt DNA Sequence:
 CTTA
 
 julia> foo()
-2nt DNA Sequence:
+5nt DNA Sequence:
 CTTAA
 
 julia> foo()
-3nt DNA Sequence:
+6nt DNA Sequence:
 CTTAAA
 ```
 
-The reason is because the sequence literal is allocated before runtime, `s` in
-`foo` is always a reference to that one sequence that was allocated before the
-runtime of your program.
-So one sequence is created, and then pushed to every time `foo` is called. Thus
-that one allocated sequence grows with every call of `foo`.
+The reason for this is because the sequence literal is allocated only once
+before the first time the function `foo` is called and run. Therefore, `s` in
+`foo` is always a reference to that one sequence that was allocated.
+So one sequence is created before `foo` is called, and then it is pushed to
+every time `foo` is called. Thus, that one allocated sequence grows with every
+call of `foo`.
 
 If you wanted `foo` to create a new sequence each time it is called,
-the you can add a flag to the end of the sequence literal: a flag of 's'
-means the sequence will be allocated before runtime as is the default behaviour.
-However providing 'd' as the flag will make sure that the sequence is
-allocated at runtime. So to change `foo` so as it creates a new sequence
+then you can add a flag to the end of the sequence literal to dictate behaviour:
+A flag of 's' means 'static': the sequence will be allocated before code is run,
+as is the default behaviour described above.
+However providing 'd' flag changes the behaviour: 'd' means 'dynamic':
+the sequence will be allocated at whilst the code is running, and not before.
+So to change `foo` so as it creates a new sequence
 each time it is called, simply add the 'd' flag to the sequence literal:
 
 ```jlcon
-julia> using Bio.Seq
-
 julia> function foo()
            s = dna"CTT"d     # 'd' flag appended to the string literal.
            push!(s, DNA_A)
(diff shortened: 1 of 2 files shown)


src/seq/bioseq.jl
```diff
@@ -64,19 +64,27 @@ typealias CharSequence      BioSequence{CharAlphabet}
 remove_newlines(s) = replace(s, r"\r|\n", "")
 
 macro dna_str(seq, flags...)
-    return DNASequence(remove_newlines(seq))
+    quote
+        return DNASequence($(remove_newlines(seq)))
```
bicycle1885: `return` is unnecessary here.

Ward9250: Cool, just removed them :+1:.

Ward9250:
Tests are parsing, it's just doing that thing it sometimes does where builds get to the end and hang :/

bicycle1885:
Let's wait for an agreement on API discussed in #348.

Ward9250:
:+1: Pre-empted the change to incorporate both behaviours into the macros with flags, If we agree on it I just need to modify the tests a tiny bit to use the new flags 😄

src/seq/bioseq.jl
```diff
@@ -63,20 +63,48 @@ typealias CharSequence      BioSequence{CharAlphabet}
 
 remove_newlines(s) = replace(s, r"\r|\n", "")
 
-macro dna_str(seq, flags...)
-    return DNASequence(remove_newlines(seq))
+macro dna_str(seq, flag)
```
bicycle1885: I think we should overload the `dna_str` macro with `dna_str(seq)` (no `flag` argument) as well to recover the original behavior.

bicycle1885: Also, updating docs is helpful :).

Ward9250: Sounds good, 😄  I'll add those in an hour or so - moving to a new office and desk :+1:

docs/src/man/seq.md
```diff
@@ -297,7 +297,99 @@ julia> char"αβγδϵ"
...
+
+You might expect that every time you call `foo`, that a DNA sequence `CTTA` would
+be returned. You might expect that this is because every time `foo` is called,
+a new DNA sequence variable `CTT` is created, and and `A` nucleotide is pushed
+to it, and the result, `CTTA` is returned.
+In other words you might expect the following output:
+
+```jlcon
+julia> foo()
+1nt DNA Sequence:
```
bicycle1885: 4nt?

docs/src/man/seq.md
```diff
@@ -297,7 +297,99 @@ julia> char"αβγδϵ"
...
+julia> foo()
+3nt DNA Sequence:
+CTTA
+```
+
+However, this is not what happens, instead the following happens:
+
+```jlcon
+julia> foo()
+1nt DNA Sequence:
```
bicycle1885: 4nt?

docs/src/man/seq.md
```diff
@@ -297,7 +297,99 @@ julia> char"αβγδϵ"
...
+julia> foo()
+2nt DNA Sequence:
+CTTAA
+
+julia> foo()
+3nt DNA Sequence:
+CTTAAA
+```
+
+The reason is because the sequence literal is allocated before runtime, `s` in
```
bicycle1885: We need to be careful when using "runtime". Actually, everything in Julia happens at runtime since Julia is a dynamic programming language. So, precisely speaking the allocation also happens at runtime but before calling the `foo` function. I think something like "the sequence literal is allocated only once before the first-time call of the function" would be a better phrasing.

docs/src/man/seq.md
```diff
@@ -297,7 +297,99 @@ julia> char"αβγδϵ"
...
+`foo` is always a reference to that one sequence that was allocated before the
+runtime of your program.
+So one sequence is created, and then pushed to every time `foo` is called. Thus
+that one allocated sequence grows with every call of `foo`.
+
+If you wanted `foo` to create a new sequence each time it is called,
+the you can add a flag to the end of the sequence literal: a flag of 's'
```
bicycle1885: s/the you/you/

docs/src/man/seq.md
```diff
@@ -297,7 +297,99 @@ julia> char"αβγδϵ"
...
+means the sequence will be allocated before runtime as is the default behaviour.
+However providing 'd' as the flag will make sure that the sequence is
+allocated at runtime. So to change `foo` so as it creates a new sequence
+each time it is called, simply add the 'd' flag to the sequence literal:
```
bicycle1885: I think we should, as a mnemonic, mention 's' and 'd' represents *static* and *dynamic*, respectively.

docs/src/man/seq.md
```diff
@@ -297,7 +297,99 @@ julia> char"αβγδϵ"
...
+       end
+foo (generic function with 1 method)
+```
+
+Now every time `foo` is called, a new sequence `CTT` is created, and an `A`
+nucleotide is pushed to it:
+
+```jlcon
+julia> foo()
+1nt DNA Sequence:
```
bicycle1885: 4nt?

docs/src/man/seq.md
```diff
@@ -297,7 +297,99 @@ julia> char"αβγδϵ"
...
+
+If you wanted `foo` to create a new sequence each time it is called,
+the you can add a flag to the end of the sequence literal: a flag of 's'
+means the sequence will be allocated before runtime as is the default behaviour.
+However providing 'd' as the flag will make sure that the sequence is
+allocated at runtime. So to change `foo` so as it creates a new sequence
+each time it is called, simply add the 'd' flag to the sequence literal:
+
+```jlcon
+julia> using Bio.Seq
```
bicycle1885: remove `using Bio.Seq`.

docs/src/man/seq.md
```diff
@@ -297,7 +297,99 @@ julia> char"αβγδϵ"
...
 
-Sequence can also be constructed from strings or arrays of nucleotide or amino
+However it should be noted that by default these sequence literals
+allocate the BioSequence object before runtime.
+This means there may be occasions where your program does not behave as you
+first expect, even though it is the intended behaviour.
+For example consider the following code:
+
+```jlcon
+julia> using Bio.Seq
```
Ward9250: Fixed :+1:

docs/src/man/seq.md
Ward9250: Fixed :+1:

docs/src/man/seq.md
Ward9250: Fixed :+1:

docs/src/man/seq.md
Ward9250: Fixed :+1:

docs/src/man/seq.md
Ward9250: Absolutely, I've changed the wording to avoid the 'runtime' confusion! :+1:

docs/src/man/seq.md
Ward9250: Added :+1:

docs/src/man/seq.md
Ward9250: Fixed :+1:

docs/src/man/seq.md
Ward9250: Fixed :+1:

Issue #149, opened 2016-03-29
Entrez utilities


abhijithch:
Would this be the right package to have the Entrez utilities features? I would like to ask the repository owners if there are any plans for developing these features, and if I could contribute? Could someone please tag this with `Question` label.

Ward9250:
Hi @abhijithch! I think this is absolutely the right package for the Entrez utilities. We have given some thought to adding a module to the package for database queries (see #52 ), and we're always happy for others to make contributions to BioJulia repos: we recently started getting automated docs and web presence online, with [contributing help](biojulia.github.io/Contributing/latest).

bicycle1885:
Yes, I'm going to implement utilities for accessing DBs like Entrez (#52) but now I'm focusing on algorithms and data structures before that. Feature requests and contributions are welcome, of course.

abhijithch:
Thanks @Ward9250 and @bicycle1885, I will post a rough script of my needs. It will be great to have your feedback before I can fine tune it.

abhijithch:
```
using Requests
import Requests: get, options
using LightXML

function fetch_pubids(db, terms)
    src="http://eutils.ncbi.nlm.nih.gov/entrez/eutils/esearch.fcgi"
    resp=get(src; query = Dict("db"=>db, "term"=>terms))

    doc=LightXML.parse_string(bytestring(resp.data))
    root_elem=LightXML.root(doc)
    idList_elem=LightXML.get_elements_by_tagname(root_elem, "IdList")
    ids_xml=LightXML.get_elements_by_tagname(idList_elem[1], "Id")

    ids = Int[]
    for i=1:length(ids_xml)
        push!(ids, parse(Int, LightXML.content(ids_xml[i])))
    end
    return ids
end
```

The above very basic script returns the PUB IDs of the articles in which `terms` occur.  This currently works only for one term, but I can handle multiple terms by appending something like `cancer AND diabetes`. Couple of other things to be taken care of are, having options to select the type of fields to search the terms from, like `Abstracts`, `pdat`, `All fields` etc. 

So an ideal pseudo function for fetching the PUBMED ids for specific terms would be,
```
function search(db, terms, return_type; kwargs...)
    src = "http://eutils.ncbi.nlm.nih.gov/entrez/eutils/esearch.fcgi"
    query = Dict("db"=>db, "term"=>terms, kwargs)
    resp = get(src; query)
    return filter_resp(resp, return_type)
end
```
[...]

PR #350, opened 2016-10-25
add E-Utils

bicycle1885:
This is a Julian API to Entrez Programming Utilities (aka. E-utilities) discussed in #149. Still in alpha quality, but examples written in docs will work.
- [x] tests
- [x] more docs
- [x] utils to convert a response to a Julia object (done by EzXML.jl)


docs/mkdocs.yml
```diff
@@ -46,6 +46,7 @@ pages:
   - Intervals: man/intervals.md
   - Biological Variation: man/var.md
   - Structure: man/structure.md
+  - Services: man/services.md
```
kdmurray91: Perhaps update submodule list in `Readme.md` too?

bicycle1885: Thank you. Updated.

bicycle1885:
Some programs won't work as I expect, maybe due to the transition from GI numbers to accessions on the server side. I'll investigate more.

bicycle1885:
Now this is ready to review. These are fairly simple wrappers for E-Utilities. `efetch` may be the most useful function because it fetches sequences with annotations, but the default return type is [ASN.1](https://en.wikipedia.org/wiki/Abstract_Syntax_Notation_One), which is difficult to deal with in Julia. So, I recommend using XML by passing `retmode="xml"` parameter to the function.

docs/src/man/services.md
```diff
@@ -0,0 +1,147 @@
...
+More"](https://www.ncbi.nlm.nih.gov/books/NBK25499/).
+
+When a request succeeds the response object has a `data` field containing
+formatted data, which can be saved to a file as demonstrated above. However,
+users are often interested in a part of the response data and may want to
+extract some fields in it. In such a case,
+[EzXML.jl](https://github.com/bicycle1885/EzXML.jl) is helpful because it offers
+lots of tools to handle XML documents. The first thing you need to do is
+converting the response data into an XML document by `parsexml`:
+```jlcon
```
kdmurray91: Perhaps these should be `jldoctest`?

bicycle1885: The example code includes nondeterministic results because the databases of NCBI are continuously updated. So, it would be almost impossible to keep the tests green.

Ward9250: In that case if everyone's happy, this can merge?

PR #355, opened 2016-11-20
Blast xml fix

kescobo:
Per request from @bicycle1885 - I altered the BLAST results parser to use EzXML rather than LightXML. This works, though I'm not sure that using XPath queries all over the place is the most efficient approach. Happy to incorporate any feedback if there's a better way.

I added EzXML to REQUIRE, but did not remove LightXML, since I think there's something else that still relies on it.

### DIFF
--- a/src/tools/blast/blastcommandline.jl
+++ b/src/tools/blast/blastcommandline.jl
@@ -16,11 +16,11 @@ immutable BLASTResult
 end
 
 """
-    readblastXML(blastrun::String)
+    readblastXML(blastrun::AbstractString)
 
 Parse XML output of a blast run. Input is an XML string eg:
 ```julia
-results = readstring(open("blast_results.xml")) # need to use `readstring` instead of `readall` for v0.5
+results = readstring(open("blast_results.xml"))
 readblastXML(results)
 ```
 
--- a/src/tools/blast/blastcommandline.jl
+++ b/src/tools/blast/blastcommandline.jl
@@ -29,10 +29,10 @@ Returns Vector{BLASTResult} with the sequence of the hit, the Alignment with que
 function readblastXML(blastrun::String; seqtype="nucl")
     dc = parsexml(blastrun)
     rt = root(dc)
-    results = Vector{BLASTResult}([])
-    for iteration in find(rt, "//BlastOutput_iterations/Iteration")
-        queryname = content(findfirst(iteration, ".//Iteration_query-def"))
-        for hit in find(iteration, ".//Iteration_hits")
+    results = BLASTResult[]
+    for iteration in find(rt, "/BlastOutput/BlastOutput_iterations/Iteration")
+        queryname = content(findfirst(iteration, "Iteration_query-def"))
+        for hit in find(iteration, "Iteration_hits")
             if countelements(hit) > 0
                 hitname = content(findfirst(hit, ".//Hit_def"))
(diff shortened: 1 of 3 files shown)


src/tools/blast/blastcommandline.jl
```diff
@@ -26,32 +26,29 @@ readblastXML(results)
...
-        queryname = content(find_element(iteration, "Iteration_query-def"))
-        hits = get_elements_by_tagname(iteration, "Iteration_hits")
-        for hit in collect(child_elements(hits[1]))
-            hitname = content(find_element(hit, "Hit_def"))
-            hsps = get_elements_by_tagname(hit, "Hit_hsps")
-            for hsp in collect(child_elements(hsps[1]))
+function readblastXML(blastrun::String; seqtype="nucl")
+    dc = parsexml(blastrun)
+    rt = root(dc)
+    results = Vector{BLASTResult}([])
```
bicycle1885: `results = BLASTResult[]` would be a conventional syntax.

src/tools/blast/blastcommandline.jl
```diff
@@ -26,32 +26,29 @@ readblastXML(results)
...
+    for iteration in find(rt, "//BlastOutput_iterations/Iteration")
```
bicycle1885: `find(rt, "/BlastOutput/BlastOutput_iterations/Iteration")` is more explicit and reduces the search space.

src/tools/blast/blastcommandline.jl
```diff
@@ -26,32 +26,29 @@ readblastXML(results)
...
+        queryname = content(findfirst(iteration, ".//Iteration_query-def"))
```
bicycle1885: You can write it as `findfirst(iteration, "Iteration_query-def")`.

bicycle1885:
BTW, I couldn't find code that depends on LightXML except BLAST tools. Does dropping LightXML from dependencies cause TravisCI failure or something?

kescobo:
@bicycle1885 Thanks for the comments. I think I addressed them all in the latest commit. 

re: search space for things I'm hitting with `.//blah` type searches - the results are explicitly structured enough that I could probably eliminate all of them. Is it better if I'm explicit as possible? I'm probably just being lazy typing...

re: LightXML - Looking back, I think I was thinking of something that you were developing that you said you started using LightXML for, but it makes sense you switched to EzXML already. I removed it from REQUIRE.

src/tools/blast/blastcommandline.jl
```diff
@@ -16,7 +16,7 @@ immutable BLASTResult
 end
 
 """
-    readblastXML(blastrun::AbstractString)
+    readblastXML(blastrun::String)
```
bicycle1885: Could you restore the original argument type here (and `readblastXML`)?

Ward9250:
I was also experimenting with LightXML in the past, so that may well be why it is in there too.

src/tools/blast/blastcommandline.jl
kescobo: Oops - sorry, thought I'd caught them all. Fixed!

Ward9250:
@kescobo Nah I don't think so right now, and when it comes to needing XML again, I'll just use Ez :)

bicycle1885:
Looks good 👍  I'll merge soon.

PR #361, opened 2016-12-15
Added more to divergence time estimation

Ward9250:
Adds some methods which mean the divergence time estimation module in Phylo can do a little more than it did previously. Specifically, you can give an array of mutation counts / genetic distances to a method of `coaltime` and get an array of time estimates returned.


Ward9250:
I'd like to merge today if there are no objections.

Issue #369, opened 2017-01-04
needs update for string.data changes


stevengj:
After JuliaLang/julia#19449, soon to be merged for Julia 0.6, you will no longer be able to access the raw bytes of a string via `string.data`; instead, do `Vector{UInt8}(string)`.

This seems to affect at least: https://github.com/BioJulia/Bio.jl/blob/23f209ded52dd4b4f434895a99c721a0779b9ac8/src/StringFields.jl#L33-L36

and it looks like some other places in the code too.

(Note also that the `1:length(data)` in that function seems wrong for another reason: `data` is a `SubString`, and hence `length` returns the length in Unicode codepoints, whereas you maybe want `1:sizeof(data)` for the length in bytes.)

Ward9250:
Thanks for raising this @stevengj, I'll have a look at this today and make a PR proposing the changes!

Ward9250:
This should now be fixed.

PR #370, opened 2017-01-04
Fix issue #369

Ward9250:
These changes fix the concerns raised in issue #369, in preparation for julia 0.6: namely that you cannot access the raw bytes of a string using `string.data` anymore.

### DIFF
--- a/src/StringFields.jl
+++ b/src/StringFields.jl
@@ -31,8 +31,9 @@ function StringField(data::Vector{UInt8})
 end
 
 function StringField(data::SubString)
-    part = data.string.data[1+data.offset:data.offset+nextind(data, data.endof)-1]
-    return StringField(part, 1:length(data))
+    stringvec = Vector{UInt8}(data)
+    idx = (1 + data.offset):(data.offset + nextind(data, data.endof) - 1)
+    return StringField(stringvec[idx], 1:endof(data))
 end
 
 # From base unicode/utf8.jl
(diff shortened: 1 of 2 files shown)


Ward9250:
Ok, I've done a text search through the source code in my Atom editor, and in addition to the StringFields constructor for SubStrings, the only other methods I could see that used `string.data` (others please check too to make sure we don't miss anything) were in gff3.jl, where a StringField is constructed in the `getindex` and `haskey` methods for the `GFFAttribute` type. Where the argument `key` is provided as a `String`. So I've added a `StringField` constructor method that takes a String, and edited the methods in gff3.jl accordingly.

@stevengj I hope this is sufficient prep for the String change in julia 0.6, but do let me know if anything else needs to be done.

Ward9250:
Tests are passing, I'd like to merge by the end of the week if no furthur comments.

src/StringFields.jl
```diff
@@ -31,8 +31,13 @@ function StringField(data::Vector{UInt8})
 end
 
 function StringField(data::SubString)
-    part = data.string.data[1+data.offset:data.offset+nextind(data, data.endof)-1]
-    return StringField(part, 1:length(data))
+    stringvec = Vector{UInt8}(data.string)
+    idx = 1+data.offset:data.offset+nextind(data, data.endof)-1
+    return StringField(stringvec[idx], 1:endof(data))
```
musm: out of curiosity why the need to change from `length` to `endof`?

stevengj: `length` is the number of characters in a string, which is totally different (for non-ASCII data) from the last index in the string.

PR #372, opened 2017-01-12
Update documentation for Bio.Var

kdmurray91:
Mainly regarding the rename of DifferentMutation to AnyMutation.

@Ward9250 Could you double check this?

Cheers,
K

PR #374, opened 2017-01-13
Fix tiny typo introduced with #372

kdmurray91:
Will merge this in a few hours if no objections.


Ward9250:
:+1: Thanks @kdmurray91

PR #376, opened 2017-01-15
replace Ragel's FASTA parser with Automa's one

bicycle1885:
This replaces Ragel's FASTA parser with our [Automa.jl](https://github.com/BioJulia/Automa.jl)'s counterpart. Currently, this depends the latest version of Automa.jl, which is not yet merged into METADATA.jl.

The patterns of FASTA are supposed to be equivalent to those of fasta.rl, and the generated code will work completely in the same way as the current one. All tests pass on my local machine. In terms of performance, the new code generator seems to be comparable to Ragel.

A main drawback is compiling a machine is slow; it takes ~5 seconds to compile a FASTA machine. However, this happens only when precompiling Bio.jl and the cached code will be used next time, so I hope users won't notice it once installed.


Ward9250:
Ah I see the only reason we're getting red flags is indeed Automa not yet present on official metadata. This is looking very cool, I look forward to trying it out with Newick and VCF.

bicycle1885:
Thanks :)
I already started working on VCF. So, you can focus on other file formats.

Ward9250:
That will be awesome, if you have a branch or anything I can look at that will be awesome, I've been looking and experimenting the past few months with ways to improve the performance of "matrix of snps" type formats - your vcf's, your plink's and so on.

bicycle1885:
@Ward9250 This is a VCF parser I'm working on: https://gist.github.com/bicycle1885/c57e0a6f1a356fbc8436ba21c7ce7962. The design and interface may change soon, so it should not be relied on. I'm going to create a pull request of it hopefully within a few days.

Ward9250:
@bicycle1885 Oh Wow awesome, from this I'm getting the feeling Automa is really quite flexible then: in the way you're essentially splicing the generated code into a function, but much of everything else in the function - setting up variables, and what actions to perform, are really up to you?

kescobo:
>A main drawback is compiling a machine is slow; it takes ~5 seconds to compile a FASTA machine. 

Do you anticipate this compile time would be significantly greater for a more complex parser? If so, I'm wondering if there might be another way to save the parsers (something like the way ragel generated a .jl file) - maybe using something like JLD.jl?

Ward9250:
> Do you anticipate this compile time would be significantly greater for a more complex parser? If so, I'm wondering if there might be another way to save the parsers (something like the way ragel generated a .jl file) - maybe using something like JLD.jl?

Perhaps macro-resolved AST of the parser could be printed to file as code?

bicycle1885:
@kescobo @Ward9250 Yeah, it is realizable. That's why I added the `clean=true` option to the `Automa.generate_exec_code` function, which generates easier-to-read Julia code for human. But the "~5 seconds" include compiling Automa.jl itself, and when we use it to define another parser the package compilation is already finished and the cost of the parser compilation will be diminished. In FASTA case, the second compilation takes ~1 second, which is much smaller than the first compilation.

Writing generated code into a file in advance is not what I want to do because it increases complexity of development. I do not want to say "oh, you forgot to run the FASTA code generator" to new developers. If the compilation time is not tolerable, I will improve the performance of compiling in Automa.jl.

Ward9250:
I think the fact modules can be precompiled makes compilation time a small issue, and I much prefer the simplicity of using Automa.jl - something I can read and understand since it's julia, and using it to generate code that is included using @eval and usual julia metaprogramming. Much simpler vs using Ragel.

Ward9250:
@bicycle1885 Is this ready to go in now? Don't wanna hit the merge button if there's any loose ends I don't know about!

kescobo:
@bicycle1885 - oh I see... I think that amount of time is entirely reasonable, and agree with @Ward9250 that simplicity is key here. I suppose we'll see what happens when a lot of parsers get added in (I'm hoping to dive into genbank very soon... finally), but don't think doing something far more complicated to save a couple seconds on pre compilation is worth it.

bicycle1885:
I'll soon check the parser a little bit further using some realistic applications, and then I'll merge it.

bicycle1885:
This is ready to be merged but Travis CI fails on macOS, which is due to the maximum time limit of a job. It took >45 minutes on macOS while <25 minutes on Linux. We need to find a way to reduce the time but it is not directly related to this change.

Ward9250:
There are over 16,000 tests for Bio.Seq, I don't know if that's where the majority of the time is spent, but if there's a way of reducing the number of tests there without sacrificing coverage - less reps or something, then that might help?

bicycle1885:
I have no idea about tests. Perhaps installing BLAST and/or something requires so long time.

PR #378, opened 2017-01-23
WIP: VCF&BCF Reader

bicycle1885:
This implements a [VCF](https://en.wikipedia.org/wiki/Variant_Call_Format) v4.3 reader. Still work-in-progress, the design may change, but I'd like to share it with you so that you can see the progress of the implementation.

The VCF file format is extremely complicated, and hence the reader here is almost ignorant of the details of its semantics. For example, the reader handles reference/alternate alleles as strings.

- [x] BCFWriter
- [x] tests
- [x] clean up

---

EDIT: removed comments generated by Covaralls.


bicycle1885:
Notes on the current design:
* The parser of VCF is defined using Automa.jl.
* Four machines are generated, namely:
    1. VCF header section (`vcfheader_machine`)
    2. VCF body section (`vcfbody_machine`)
    3. VCF metainfo line (`vcfmetainfo_machine`)
    4. VCF record line (`vcfrecord_machine`)
* `VCFRecord` and `VCFMetaInfo` do deferred parsing (i.e. fields are extracted on demand).
    * This has a big advantage in performance because it does far fewer jobs while reading.
    * The type holds data bytes with indexes as follows:
```julia
type VCFRecord
    # data and indexes
    data::Vector{UInt8}
    chrom::UnitRange{Int}
    pos::UnitRange{Int}
    id::Vector{UnitRange{Int}}
    ...
end
```

bicycle1885:
This is almost finished. Please try the VCF and BCF parsers if you're interested in them. cc. @Ward9250

bicycle1885:
The diff is getting too big and I think it would be better to merge this first and then add follow-up patches to improve the quality once tests pass. I'm going to merge this tomorrow today if no comments.

Ward9250:
Sweet, travis failiure is probably it timing out on OSX again.

PR #379, opened 2017-01-31
declare Bio.distance to resolve the method name conflict

bicycle1885:
This resolves a method name conflict that happens between sub-modules of Bio.jl. The `distance` function is defined in three sub-modules: `Bio.Align`, `Bio.Structure` and `Bio.Var`. This is problematic when we call `using` two of them:
```
julia> using Bio.Align, Bio.Var
INFO: Recompiling stale cache file /Users/kenta/.julia/lib/v0.5/Bio.ji for module Bio.
INFO: compiling FASTA

julia> distance
WARNING: both Var and Align export "distance"; uses of it in module Main must be qualified
ERROR: UndefVarError: distance not defined

```

This patch declares the `distance` function in `Bio` and extends it in each submodule. If this approach works well, I'd like to use it more pervasively in Bio.jl.

CC: @Ward9250 and @jgreener64.

### DIFF
--- a/src/Bio.jl
+++ b/src/Bio.jl
@@ -2,6 +2,7 @@ __precompile__()
 
 module Bio
 
+include("declare.jl")
 include("IO.jl")
 include("Ragel.jl")
 include("StringFields.jl")
--- a/src/align/Align.jl
+++ b/src/align/Align.jl
@@ -112,6 +112,7 @@ export
     SAM_FLAG_SUPPLEMENTARY
 
 import Bio
+import Bio: distance
 using Bio.Seq
 using Bio.Intervals
 import Bio.StringFields: StringField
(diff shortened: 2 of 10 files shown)


jgreener64:
Well spotted, looks good to me 👍

Ward9250:
This also looks good to me. What were you thinking of when you say you want to use it more pervasively @bicycle1885?

bicycle1885:
@Ward9250 I meant we always declare functions in declare.jl when we find methods shared among submodules. For example, `seqname` is now defined in `Bio.Seq` and overloaded by `Bio.Intervals` and `Bio.Align`.  But the reason why it is defined in `Bio.Seq` is just because `Bio.Seq` is the first place where it is defined. So, I'd like to declare these functions just under the `Bio` module and overload them in submodules when needed.

Ward9250:
Ooooh yes that makes sense, I misread and thought you meant that you wanted to use 'distance' more pervasively in Bio.jl! I like this idea, in fact I will add it to contributing guidelines!

PR #383, opened 2017-01-31
Mutation Counting & Distance API redux

Ward9250:
Doing the work increasing the performance of mutation counting in DNA, and RNA sequences, I realised I really did not like the API/interface for mutation counting and Distance computation. This sorts the horrid stuff out and makes the API consistent, as well as adding more site/mutation types that can be counted.


jgreener64:
Minor: by 'naieve' do you mean 'naive'?

Ward9250:
I do indeed!
I'll fix that shortly.

Ward9250:
I'm going to try and do a big squash and merge this very soon, as it works as I intend it to, polishing and adjustment of the distance computations will then follow in a set of smaller PRs.

Ward9250:
Tests pass, except on mac but it seems to be an issue with building HDF5, unrelated to this PR. I'll merge at 4:00pm if there are not further comments.

PR #386, opened 2017-02-04
Tidy import

bicycle1885:
The larger our package is, the harder it is to identify names imported from other packages. The reason is mainly because the `using` syntax hides the origin of names and we (at least I) do not have good memory enough to remember all unqualified names.

This pull request will fix the problem by removing `using` used under the `src` directory. The rules are:
1. Do not use `using` to import names from other modules.
2. Use `import` to use other modules.
3. Import names from other modules only when the names are used repeatedly.
4. Use `imoprtall Bio` to import and extend methods shared among submodules of `Bio`.

Currently, I fixed `Bio.Seq`, `Bio.Align`, and `Bio.Intervals` only. I'm going to tidy up other modules if you agree with me.

### DIFF
--- a/src/Ragel.jl
+++ b/src/Ragel.jl
@@ -10,8 +10,8 @@ module Ragel
 
 export tryread!
 
-using BufferedStreams
 import Bio.IO: FileFormat, AbstractReader, stream
+import BufferedStreams: BufferedStreams, BufferedInputStream
 
 # A type keeping track of a ragel-based parser's state.
 type State{T<:BufferedInputStream}
--- a/src/align/Align.jl
+++ b/src/align/Align.jl
@@ -96,7 +96,7 @@ export
     seqlength,
     qualities,
     header,
-    isoverlapping,  # re-export from Bio.Intervals
+    isoverlapping,
     # SAM flags
     SAM_FLAG_PAIRED,
     SAM_FLAG_PROPER_PAIR,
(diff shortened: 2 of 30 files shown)


Ward9250:
So for bullet point 2 - what is the case for using import for using Bio modules in other Bio modules, if you're only using the methods and not extending them?

bicycle1885:
Just for simplicity of rules. We'll need to add one more exceptional rule if we allow `using` for Bio modules.

Ward9250:
Cool, one more question, but I'm on board with this change, I see some of the changes include reexporting of a reimported name, do we have to do this for all the methods extended that are imported from declaring.jl in Bio?

bicycle1885:
Yes, if you'd like to export them. I think listing all exported names at the top of a module definition is a good manner because it is more explicit than using some magical macros like Reexport.jl.

bicycle1885:
I've finished removing `using`s under src/ except src/phylo and src/var because these will conflict with #383. cc: @Ward9250

Ward9250:
Thanks @bicycle1885, I can change using a to imports as part of #383 for
those two modules if you like.

bicycle1885:
Okay, so I'm going to leave them as they are and merge this pull request. Please update them in #383.

Ward9250:
Sounds good to me! I'm hoping to have #383 ready to merge in an hour or so when the tests pass, if you could take a look at let me know if you're are happy with the import statements I will add that would be great.

bicycle1885:
Why not :). Let me know when #383 is ready for review.

PR #388, opened 2017-02-07
replace Pkg.dir calls with joinpath(dirname(...))

bicycle1885:
`Pkg.dir` won't point to the correct path when the package is installed in a non-standard location.


Ward9250:
I wonder if Pkg.dir will work correctly when the Pkg overhaul happens?

bicycle1885:
I have no idea, but the situation may be the opposite. People may start installing packages in non-standard locations and packages that call `Pkg.dir` won't work.

PR #389, opened 2017-02-07
encourage users to use Discourse for technical questions

bicycle1885:
I'd like to encourage users to use Discourse for technical questions. This is because Gitter has no threads and it is often difficult to distinguish interesting replies from others. Also, Discourse will catch more eyes from the Julia community, so I guess it may strengthen our presence in the community.


bdeonovic:
[link](https://discourse.julialang.org/) is probably good idea

bicycle1885:
Yep, thank you :)

PR #390, opened 2017-02-08
WIP: masked sequence

bicycle1885:
This is what I mentioned in https://gitter.im/BioJulia/Bio.jl?at=589a5affde504908229d2b84. We can mask specific regions of a sequence by wrapping a sequence by `MaskedSequence`. For example, if you'd like to mask 'N' regions, you call `mask(x -> x == DNA_N, seq)` and create a masked sequence.

- [ ] Support approximate search and regular expression search.
- [ ] Tests.
- [ ] Docs.

### DIFF
--- a/src/seq/Seq.jl
+++ b/src/seq/Seq.jl
@@ -154,7 +154,10 @@ export
     demultiplex,
     seqmatrix,
     majorityvote,
-    tryread!
+    tryread!,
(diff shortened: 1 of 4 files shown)


bicycle1885:
This will be transferred to another package (maybe BioSequences.jl?) after #425.

PR #392, opened 2017-02-11
Rename of DNANucleotide, RNANucleotide, Nucleotide

Ward9250:
Implements renames as per #391.

PR #393, opened 2017-02-13
deprecate DNA/RNANucleotide

bicycle1885:
This is a follow-up changes of #392:
* Add deprecations for old names.
* Add NEWS.md as per https://github.com/JuliaLang/julia/blob/master/NEWS.md.


Ward9250:
Looks cool, haven't used compat myself, so @deprecate_binding allows people to use the old names but warns them?

bicycle1885:
Yes.

```
julia> convert(DNA, 'A')
DNA_A

julia> convert(DNANucleotide, 'A')
WARNING: Seq.DNANucleotide is deprecated, use Bio.Seq.DNA instead.
  likely near no file:0
DNA_A

```

PR #404, opened 2017-03-06
add more methods for VCFHeader like SAMHeader

bicycle1885:
Also, some docs and conversions from `AbstractString` and `Vector{UInt8}` to `VCFMetainfo` are added.


bicycle1885:
This wouldn't be controversial because this is a consistency enhancement

PR #410, opened 2017-03-08
Reorganize BioSequence code

Ward9250:
This splits up the rather large julia file which contains all the code for BioSequences, into separate files, to make tackling and understanding the source code easier for people learning how BioSequences work.


src/seq/Seq.jl
```diff
@@ -188,12 +188,12 @@ function gap end
...
-include("sequence.jl")
 include("nucleicacid.jl")
 include("aminoacid.jl")
 include("alphabet.jl")
 include("bitindex.jl")
-include("bioseq.jl")
+include("sequence.jl")
+include("bioseq/bioseq.jl")
 include("hash.jl")
 include("randseq.jl")
```
bicycle1885: hash.jl and randseq.jl are also tightly related to `BioSequence`, so it may be better to move them into the bioseq directory.

src/seq/bioseq/bioseq.jl
```diff
@@ -0,0 +1,123 @@
...
+Base.summary(seq::CharSequence) = string(length(seq), "char ", "Char Sequence")
+
+include("sequences/indexing.jl")
+include("sequences/constructors.jl")
+include("sequences/iteration.jl")
+include("sequences/copying.jl")
+include("sequences/conversion.jl")
+include("sequences/stringliterals.jl")
+include("sequences/transformations.jl")
+include("sequences/operators.jl")
```
bicycle1885: "sequences/" is not needed.

src/seq/Seq.jl
Ward9250: Good idea, I'll move them!

bicycle1885:
Feel free to merge if it's ready.

PR #414, opened 2017-03-13
Fix to documentation

Ward9250:
This is a simple change to an error which is preventing the update of our online documentation, following re-organization of the `Seq` module.

I'll leave this up for an hour but since it's simple I want to merge it ASAP so as our online manual gets updated.


bicycle1885:
lgtm

PR #415, opened 2017-03-13
Genomic Distances using MASH

kescobo:
This is a PR for an idea [I brought up months ago in gitter](https://gitter.im/BioJulia/Bio.jl?at=584c96b1c29531ac5d4917e2). I'd like to add a small module to calculate the [MASH distance](https://genomebiology.biomedcentral.com/articles/10.1186/s13059-016-0997-x) between two genomes. 

Briefly, the theory is that you generate a hash for all kmers of size `k`, but only store a "bottom sketch" of a particular genome of size `s` - that is the `s` smallest hashes. If you store the hashes in sorted order, you only need to check if each subsequent hash is larger than a single value, so this is quite fast. The genome distance for two genomes is then essentially the [Jaccard index](https://en.wikipedia.org/wiki/Jaccard_index) of the minhashes, with some additional modification to account for the size of the kmers used. 

I have [an implementation](https://github.com/kescobo/MashTun/blob/master/src/MashTun.jl) that I've validated and is quite fast - from two FASTA files of ~5 Mbp each to MASH distance takes ~2 sec on my 3 year old macbook pro. Most of this time is spent generating the sketch, but once a sketch is made, the comparison between two genomes takes less than a millisecond (so adding additional genomes scales very well).

I'm ready to start adding documentation and tests, but thought I should open a PR and get some guidance on where this fits in Bio.jl, and if I should add anything else to the module.

### Minhash
- [x] Implement MinHash in Seq
- [x] MinHash Tests
- [x] Documentation

### MASH Distance
- [x] Implement MASH distance in Var
- [x] MASH Tests
- [x] Documentation

## Still to-do
- [x]  fix CI errors

### DIFF
--- a/src/seq/minhash.jl
+++ b/src/seq/minhash.jl
@@ -1,0 +1,102 @@
+# MinHash
+# ========
+#
+# Functions to generate MinHash sketches of biological sequences
+#
+# This file is a part of BioJulia.
+# License is MIT: https://github.com/BioJulia/Bio.jl/blob/master/LICENSE.md
+
+type MinHashSketch
+    sketch::Vector{UInt64}
+    kmersize::Int
+
+    function MinHashSketch(sketch::Vector, kmersize::Int)
+        length(sketch) > 0 || error("Sketch cannot be empty")
+        kmersize > 0 || error("Kmersize must be greater than 0")
+        new(sketch, kmersize)
+    end
+end
+
+Base.getindex(s::MASHSketch, part) = getindex(s.sketch, part)
+
+Base.length(s::MinHashSketch) = length(s.sketch)
+
+Base.size(s::MinHashSketch) = (length(s), s.kmersize)
+
+Base.start(s::MinHashSketch) = start(s.sketch)
+
+Base.next(s::MinHashSketch, state) = next(s.sketch, state)
+
+Base.done(s::MinHashSketch, state) = done(s.sketch, state)
+
+function Base.:(==)(a::MASHSketch, b::MASHSketch)
+    a.kmersize == b.kmersize && a.sketch == b.sketch
+end
+
+
+function revcomphash(kmer::Kmer)
+    k = minimum((kmer, reverse_complement(kmer)))
+    return hash(k)
+end
+
+
+function kmerminhash(seq::BioSequence, kmerset, kmerhashes::Vector{UInt64}, k::Int, s::Int)
+    typeof(kmerset) <: Set || typeof(kmerset) <: SortedSet || error("Kmerset must be a `Set` or `SortedSet`")
+    length(kmerhashes) <= s || error("Kmerhashes cannot be larger than the set size")
+
+    for kmer in each(DNAKmer{k}, seq)
+        if length(kmerhashes) == 0
+            if length(kmerset) < s
+                push!(kmerset, revcomphash(kmer[2]))
+            elseif length(kmerset) == s
+                kmerset = SortedSet(kmerset)
+                for i in kmerset
+                    push!(kmerhashes, pop!(kmerset))
+                end
+            end
+        end
+
+        if length(kmerhashes) == s
+            h = revcomphash(kmer[2])
+            if h < kmerhashes[end]
+                i = searchsortedlast(kmerhashes, h)
+                if i == 0 && i != kmerhashes[1]
+                    pop!(kmerhashes)
+                    unshift!(kmerhashes, h)
+                elseif h != kmerhashes[i]
+                    pop!(kmerhashes)
+                    insert!(kmerhashes, i+1, h)
+                end
+            end
+        end
+
+    end
+    return (kmerset, kmerhashes)
+end
+
+
+function minhash(seq::BioSequence, k::Int, s::Int)
+    kmerset = Set{UInt64}()
+    kmerhashes = Vector{UInt64}()
+    kmerset, kmerhashes = kmerminhash(seq, kmerset, kmerhashes, k, s)
+    return MinHashSketch(kmerhashes, k)
+end
+
+function minhash{T<:BioSequence}(seqs::Vector{T}, k::Int, s::Int)
+    kmerset = Set{UInt64}()
+    kmerhashes = Vector{UInt64}()
+
+    for seq in seqs
+        kmerset, kmerhashes = kmerminhash(seq, kmerset, kmerhashes, k, s)
+    end
+    return MinHashSketch(kmerhashes, k)
+end
+
+function minhash{T<:BioSequence}(seqs::FASTAReader{T}, k::Int, s::Int)
+    kmerset = Set{UInt64}()
+    kmerhashes = Vector{UInt64}()
+    for seq in seqs
+        kmerset, kmerhashes = kmerminhash(seq.seq, kmerset, kmerhashes, k, s)
+    end
+    return MinHashSketch(kmerhashes, k)
+end
--- a/src/seq/minhash.jl
+++ b/src/seq/minhash.jl
@@ -37,13 +37,6 @@ function Base.:(==)(a::MinHashSketch, b::MinHashSketch)
     return a.kmersize == b.kmersize && a.sketch == b.sketch
 end
 
-# A seqence and its reverse complement should be the same, so take the smallest
-# hash of a seq or its reverse complement.
-function revcomphash(kmer::Kmer)
-    k = min(kmer, reverse_complement(kmer))
-    return hash(k)
-end
-
 
 function kmerminhash!{k}(::Type{DNAKmer{k}}, seq::BioSequence, s::Integer, kmerhashes::Vector{UInt64})
     # generate first `s` kmers
@@ -51,7 +44,7 @@ function kmerminhash!{k}(::Type{DNAKmer{k}}, seq::BioSequence, s::Integer, kmerh
     state = start(iter)
     while length(kmerhashes) < s && !done(iter, state)
         (_, kmer), state = next(iter, state)
-        h = revcomphash(kmer)
+        h = hash(canonical(kmer)) # hash lexigraphic minimum of kmer and reverse compliment of kmer
         if h ∉ kmerhashes
             push!(kmerhashes, h)
         end
@@ -62,7 +55,7 @@ function kmerminhash!{k}(::Type{DNAKmer{k}}, seq::BioSequence, s::Integer, kmerh
     # scan `seq` to make a minhash
     while !done(iter, state)
         (_, kmer), state = next(iter, state)
-        h = revcomphash(kmer)
+        h = hash(canonical(kmer)) # hash lexigraphic minimum of kmer and reverse compliment of kmer
         if h < kmerhashes[end]
             i = searchsortedlast(kmerhashes, h)
             if i == 0 && h != kmerhashes[1]
--- a/src/seq/minhash.jl
+++ b/src/seq/minhash.jl
@@ -16,7 +16,7 @@ hashes for kmers of a given length. The type contains two parameters:
 * .sketch: a sorted set of hashes
 * .kmersize: the length of kmers used to generate the sketch
 """
-type MinHashSketch
+immutable MinHashSketch
     sketch::Vector{UInt64}
     kmersize::Int
 
@@ -29,7 +29,6 @@ end
 
 Base.getindex(s::MinHashSketch, part) = getindex(s.sketch, part)
 Base.length(s::MinHashSketch) = length(s.sketch)
-Base.size(s::MinHashSketch) = (length(s), s.kmersize)
 Base.start(s::MinHashSketch) = start(s.sketch)
 Base.next(s::MinHashSketch, state) = next(s.sketch, state)
 Base.done(s::MinHashSketch, state) = done(s.sketch, state)
@@ -41,7 +40,7 @@ end
 # A seqence and its reverse complement should be the same, so take the smallest
 # hash of a seq or its reverse complement.
 function revcomphash(kmer::Kmer)
-    k = minimum((kmer, reverse_complement(kmer)))
+    k = min((kmer, reverse_complement(kmer)))
     return hash(k)
 end
 
@@ -66,7 +65,7 @@ function kmerminhash(seq::BioSequence, kmerset, kmerhashes::Vector{UInt64}, k::I
             h = revcomphash(kmer[2])
             if h < kmerhashes[end]
                 i = searchsortedlast(kmerhashes, h)
-                if i == 0 && i != kmerhashes[1]
+                if i == 0 && h != kmerhashes[1]
                     pop!(kmerhashes)
                     unshift!(kmerhashes, h)
                 elseif h != kmerhashes[i]
@@ -106,7 +105,7 @@ function minhash{T<:BioSequence}(seqs::Vector{T}, k::Int, s::Int)
     return MinHashSketch(kmerhashes, k)
 end
 
-function minhash{T<:BioSequence}(seqs::FASTAReader{T}, k::Int, s::Int)
+function minhash{T<:BioSequence}(seqs::FASTA.Reader{T}, k::Int, s::Int)
     kmerset = Set{UInt64}()
     kmerhashes = Vector{UInt64}()
     for seq in seqs
--- a/src/seq/minhash.jl
+++ b/src/seq/minhash.jl
@@ -25,14 +25,14 @@ type MinHashSketch
     end
 end
 
-Base.getindex(s::MASHSketch, part) = getindex(s.sketch, part)
+Base.getindex(s::MinHashSketch, part) = getindex(s.sketch, part)
 Base.length(s::MinHashSketch) = length(s.sketch)
 Base.size(s::MinHashSketch) = (length(s), s.kmersize)
 Base.start(s::MinHashSketch) = start(s.sketch)
 Base.next(s::MinHashSketch, state) = next(s.sketch, state)
 Base.done(s::MinHashSketch, state) = done(s.sketch, state)
 
-function Base.:(==)(a::MASHSketch, b::MASHSketch)
+function Base.:(==)(a::MinHashSketch, b::MinHashSketch)
     a.kmersize == b.kmersize && a.sketch == b.sketch
 end
 
--- a/src/seq/minhash.jl
+++ b/src/seq/minhash.jl
@@ -18,12 +18,12 @@ hashes for kmers of a given length. The type contains two parameters:
 """
 immutable MinHashSketch
     sketch::Vector{UInt64}
-    kmersize::Int
+    kmersize::Integer
 
-    function MinHashSketch(sketch::Vector, kmersize::Int)
+    function MinHashSketch(sketch::Vector, kmersize::Integer)
         length(sketch) > 0 || error("Sketch cannot be empty")
         kmersize > 0 || error("Kmersize must be greater than 0")
-        new(sketch, kmersize)
+        return new(sketch, kmersize)
     end
 end
 
@@ -34,18 +34,18 @@ Base.next(s::MinHashSketch, state) = next(s.sketch, state)
 Base.done(s::MinHashSketch, state) = done(s.sketch, state)
 
 function Base.:(==)(a::MinHashSketch, b::MinHashSketch)
-    a.kmersize == b.kmersize && a.sketch == b.sketch
+    return a.kmersize == b.kmersize && a.sketch == b.sketch
 end
 
 # A seqence and its reverse complement should be the same, so take the smallest
 # hash of a seq or its reverse complement.
 function revcomphash(kmer::Kmer)
-    k = min((kmer, reverse_complement(kmer)))
+    k = min(kmer, reverse_complement(kmer))
     return hash(k)
 end
 
 
-function kmerminhash{k}(::Type{DNAKmer{k}}, seq::BioSequence, s::Int)
+function kmerminhash{k}(::Type{DNAKmer{k}}, seq::BioSequence, s::Integer)
     # generate first `s` kmers
     kmerhashes = UInt64[]
     iter = each(DNAKmer{k}, seq)
@@ -82,16 +82,16 @@ function kmerminhash{k}(::Type{DNAKmer{k}}, seq::BioSequence, s::Int)
 end
 
 """
-    minhash(seq, k::Int, s::Int)
+    minhash(seq, k::Integer, s::Integer)
 
 Generate a MinHash sketch of size `s` for kmers of length `k`.
 """
-function minhash(seq::BioSequence, k::Int, s::Int)
+function minhash(seq::BioSequence, k::Integer, s::Integer)
     kmerhashes = mykmerminhash(DNAKmer{k}, seq, s)
     return MinHashSketch(kmerhashes, k)
 end
 
-function minhash{T<:BioSequence}(seqs::Vector{T}, k::Int, s::Int)
+function minhash{T<:BioSequence}(seqs::Vector{T}, k::Integer, s::Integer)
     kmerset = Set{UInt64}()
     kmerhashes = Vector{UInt64}()
 
@@ -103,7 +103,7 @@ function minhash{T<:BioSequence}(seqs::Vector{T}, k::Int, s::Int)
     return MinHashSketch(kmerhashes, k)
 end
 
-function minhash{T<:BioSequence}(seqs::FASTA.Reader{T}, k::Int, s::Int)
+function minhash{T<:BioSequence}(seqs::FASTA.Reader{T}, k::Integer, s::Integer)
     kmerset = Set{UInt64}()
     kmerhashes = Vector{UInt64}()
     for seq in seqs
--- a/src/seq/minhash.jl
+++ b/src/seq/minhash.jl
@@ -5,6 +5,8 @@
 #
 # This file is a part of BioJulia.
 # License is MIT: https://github.com/BioJulia/Bio.jl/blob/master/LICENSE.md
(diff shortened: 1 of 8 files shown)


Ward9250:
This is great @kescobo!

Ward9250:
I think this probably fits with the methods I've been writing in Var for `distance`, where the first argument is a type which dictates the distance method used.

Ward9250:
@kescobo My question to you and @kmsquire is: This and CountMin.jl both rely on Sketches, so are there commonalities between these two sets of code which can be "merged" within Bio and a consistent module/methods/types for dealing with hashes and sketches implemented that fulfils the needs of the two?

bicycle1885:
MinHash is use in many applications and I'd say +1 for adding this in Bio.jl. I think this will fit in Bio.Seq.

kescobo:
I might be wrong, but I think CountMin.jl is doing something different. If I understand correctly, a Count-Min sketch gives counts of `things`, and basically serves as a frequency table. A MinHash sketch by contrast doesn't care about frequencies, it's just storing a snapshot of all of the `things` that are present, regardless of their frequency. 

I looked at CountMin.jl and didn't see an obvious way that it would be used in this context, but would be happy for @kdmurray91 to take a look and let me know if I missed something.

Ward9250:
I think this definitely should go in Bio.jl, but I think the code may need to be split. There are several aspects of this PR, so I can imagine creating MinHashes of sequences goes in `Seq` as a MinHas is broadly useful, but then computation of the evolutionary distance from said MinHash goes in either `Phylo` or `Var`, as that aspect of the PR is computing measures of genetic variation or relatedness.

@kescobo I think you are right, I've also had a look at CountMin in the past 10 minutes and can see indeed it is different.

kescobo:
@Ward9250 That makes sense. To me, Phylo makes the most sense for MASH distance, but you have a much better sense of that package and Var, so I'll defer to your judgement there.

Ward9250:
In that case, the first step to refinement of organisation is to move purely MinHash code to `Seq`. We can deliberate Var vs Phylo later :)

kescobo:
@Ward9250 upon further reflection and examination of what's in phylo and var, I changed my mind. I've added the MASH distance functions to var, but I'm not 100% sure how to incorporate it into what you've currently got going on in `distances.jl`

I know you're actively working on that, I'm wondering if I should let you handle that integration, and if so, wait on tests/docs.

Ward9250:
I'm happy for you to continue with tests and docs and I can then deal with integration subsequently if you'd like - I'm easy :+1:

kmsquire:
> @kescobo My question to you and @kmsquire is: This and CountMin.jl both rely on Sketches, so are there commonalities between these two sets of code which can be "merged" within Bio and a consistent module/methods/types for dealing with hashes and sketches implemented that fulfils the needs of the two?

@Ward9250, I think you meant to ping @kdmurray91.

Ward9250:
So I did! 😆

src/seq/minhash.jl
```diff
@@ -0,0 +1,118 @@
...
+"""
+MinHash Sketch type
+
+MinHash sketches are a sorted set of kmer hashes, containing the smallest `s`
+hashes for kmers of a given length. The type contains two parameters:
+
+* .sketch: a sorted set of hashes
+* .kmersize: the length of kmers used to generate the sketch
+"""
+type MinHashSketch
```
bicycle1885: This should be `immutable` if possible.

src/seq/minhash.jl
```diff
@@ -0,0 +1,118 @@
...
+    function MinHashSketch(sketch::Vector, kmersize::Int)
+        length(sketch) > 0 || error("Sketch cannot be empty")
+        kmersize > 0 || error("Kmersize must be greater than 0")
+        new(sketch, kmersize)
+    end
+end
+
+Base.getindex(s::MinHashSketch, part) = getindex(s.sketch, part)
+Base.length(s::MinHashSketch) = length(s.sketch)
+Base.size(s::MinHashSketch) = (length(s), s.kmersize)
```
bicycle1885: This looks an abuse of overloading. `Base.size` should return the dimensions (or sizes) of an array-like data structure.

src/seq/minhash.jl
```diff
@@ -0,0 +1,118 @@
...
+Base.done(s::MinHashSketch, state) = done(s.sketch, state)
+
+function Base.:(==)(a::MinHashSketch, b::MinHashSketch)
+    a.kmersize == b.kmersize && a.sketch == b.sketch
+end
+
+# A seqence and its reverse complement should be the same, so take the smallest
+# hash of a seq or its reverse complement.
+function revcomphash(kmer::Kmer)
+    k = minimum((kmer, reverse_complement(kmer)))
```
bicycle1885: `k = min(kmer, reverse_complement(kmer))` would be better.

src/seq/minhash.jl
```diff
@@ -0,0 +1,118 @@
...
+    return hash(k)
+end
+
+
+function kmerminhash(seq::BioSequence, kmerset, kmerhashes::Vector{UInt64}, k::Int, s::Int)
```
bicycle1885: By passing `k` as a usual argument, you cannot enjoy the full optimization of Julia. `kmerminhash{k}(::Type{DNAKmer{k}}, seq, kmerset, kmerhashes, s)` may significantly improve the performance of this function.

bicycle1885: If this function updates arguments, the name should end with `!`.

src/seq/minhash.jl
```diff
@@ -0,0 +1,118 @@
...
+                    push!(kmerhashes, pop!(kmerset))
+                end
+            end
+        end
+
+        if length(kmerhashes) == s
+            h = revcomphash(kmer[2])
+            if h < kmerhashes[end]
+                i = searchsortedlast(kmerhashes, h)
+                if i == 0 && i != kmerhashes[1]
```
bicycle1885: `i != kmerhashes[1]` compares an index and a hash value?

src/var/mash.jl
```diff
@@ -0,0 +1,34 @@
...
+#
+# Functions to MASH distance for MinHash sketches
+#
+# see DOI: 10.1186/s13059-016-0997-x
+#
+# This file is a part of BioJulia.
+# License is MIT: https://github.com/BioJulia/Bio.jl/blob/master/LICENSE.md
+
+function jaccarddistance(sketch1::MinHashSketch, sketch2::MinHashSketch)
+    d = length(setdiff(sketch1.sketch, sketch2.sketch))
```
bicycle1885: Since `sketch`es are sorted, I think `d` can be computed more efficiently (in `O(M+N)`) like the merge step of the merge sort.

src/seq/minhash.jl
```diff
@@ -0,0 +1,118 @@
...
+
+MinHash sketches are a sorted set of kmer hashes, containing the smallest `s`
+hashes for kmers of a given length. The type contains two parameters:
+
+* .sketch: a sorted set of hashes
+* .kmersize: the length of kmers used to generate the sketch
+"""
+type MinHashSketch
+    sketch::Vector{UInt64}
+    kmersize::Int
```
Ward9250: I wonder if having kmer size as a parameter, like array dimensions are a parameter, will afford any benefit here?

src/var/mash.jl
Ward9250: @bicycle1885 I'm not sure about merge sort, but do you mean something like: Iterating through the arrays independently in a loop, and since you know they're sorted you know based greater than, less than and so on, whether elements are in array A and are not in array B?

@kescobo Note that setdiff gets the values that are in sketch1 and not sketch2, but not the other way around, values in sketch2 that are not in sketch1 will not be returned. If this is what you want all is well, but I thought I'd mention it just in case.

kdmurray91:
Hi all,

Sorry, I am just checking in here. I'll review the comments (& code) here properly tomorrow, but the short answer is that @kescobo is right about CountMin.jl. This PR happens to be the essence of my whole PhD, so expect some verbosity from me in the (aussie) morning!

For now, you can take a look at [kWIP](https://github.com/kdmurray91/kwip) for our approach to this problem using count-min sketches. I have a preliminary julia re-write of kWIP too, FWIW.

Cheers,
K

src/seq/minhash.jl
kescobo: Can do.

src/seq/minhash.jl
kescobo: Is there something that would be better (something like `dims()`, or should I just not bother?

src/seq/minhash.jl
kescobo: Can do.

src/seq/minhash.jl
kescobo: I think you're misunderstanding this function or I'm misunderstanding you. The `k` argument here isn't a kmer, it's an argument to specify the length of kmers used to build the minhash sketch. And it doesn't modify the sequence, it generates all possible kmers from a sequence and builds a sketch using the smallest `s` hashes.

src/seq/minhash.jl
kescobo: Good catch! That should be `if i == 0 && h != kmerhashes[1]`.

src/var/mash.jl
kescobo: @bicycle1885 I initially wrote something out to do something like what you're suggesting (you can look [here](https://github.com/kescobo/MashTun/blob/ab044da78095595f8a3b6379cd49903f48ba8d1f/src/MashTun.jl) starting at ln 117 for what I tried), but it didn't seem to be any faster. Probably my lack of knowledge for how to do this efficiently is to blame - happy to take guidance on that.

@Ward9250 the calculation is the length of the intersect over the length of the union. `setdiff` tells me how many items are in one but not the other, and since we know the size of each set, that's enough information to know how many are shared as well (we don't need to know what's shared, just the number).

src/seq/minhash.jl
kescobo: Can you expound on this? I'm not familiar with making something a parameter.

kescobo:
@kdmurray91 Awesome - would be great to have your insight. I'm a total newb at this :-)

src/seq/minhash.jl
Ward9250: I mean doing something like the following gist: https://gist.github.com/Ward9250/1c550e345b742f3287da4f9a2c2fda08

src/var/mash.jl
Ward9250: @kescobo I think the reason it may not have been much faster is the use of `shift!`? I think you can probably do the task you are tying to do in 117 without all the array manipulation, and just keeping track of two indicies.

kescobo: @Ward9250 Yeah, that makes sense. I tried a number of different data structures, and I think at one point I was trying to get a vector of matches out of that function, but later morphed into just a calculation of the number. I can try to re-work that code and see if it improves things.

Ward9250: That would be great if it's possible to eke out a bit more speed. I drew a draft version of a function that does the same job as `setdiff` but is specific to sorted arrays, when I checked out `setdiff` with `@less` the other day, I can make a gist out of it if it helps.

src/seq/minhash.jl
bicycle1885: To make it clear, I wrote a function that is ~10x faster than your implementation: https://gist.github.com/bicycle1885/23f15ac022cff46cb454c9be1a585987. In this function, `k` is passed as a type argument and it gives Julia more opportunities to generate specialized code that is specific to the length of kmers. Also, this function modifies the `kmerset` and `kmerhashes` arguments and hence this should be names like `kmerminhash!`.

kescobo: @bicycle1885 To your second point (`kmerminhash!()`) - I woke up in the middle of the night realizing what you meant! I think my python brain was in force and only thinking about the first argument. 

To your first point, it's no surprise that it was me misunderstanding you. I'm not clear why that provides such a performance boost, but the evidence is clear. And that also makes point 2 moot, since you're not doing the weirdness with passing sets around. Thanks!

kescobo: @bicycle1885 Though now I'm going to modify the code, what about the `minhash()` functions that take groups of sequences rather than just one? Will it still be performant if I do something like:

```
function kmerminhash{k, T<:BioSequence}(::Type{DNAKmer{k}}, seqs::Vector{T}, s::Int)
    # generate first `s` kmers
    kmerhashes = UInt64[]
    iter = [each(DNAKmer{k}, seq) for seq in seqs]
    ...
```

?

Or would it make more sense to make a separate iterator function?

bicycle1885: I'm not sure what the purpose of `minhash` functions taking multiple sequences is. I thought `minhash` is a kind of summary of a sequence.

kescobo: @bicycle1885 That's true, but if the sequence you want a summary of is something like a genome that has multiple contigs, then you need the minhash of all of those contigs together (in principle, you could calculate the minhash of each sequence, and then merge and take the bottom `s` hashes from that, but that seems inefficient).

src/var/mash.jl
kescobo: @Ward9250 @bicycle1885 Alas, this change is only 40x faster o_O

https://gist.github.com/kescobo/7341eed840b56ba90d5661da2d745cac

Ward9250: @kescobo That's it! idxes vs. shifts and slices ftw! 😆

kescobo:
Hey guys, I think this is close - just need some advice on how to move forward with generating sketches that span multiple sequences using @bicycle1885's speedy approach ([see here](https://github.com/BioJulia/Bio.jl/pull/415#discussion_r106399757)). Does it make sense to extend the `each()` function for `DNAKmer` to operate on groups of sequences?

src/seq/minhash.jl
bicycle1885: Ah, it makes sense. So, what about having an update function like `minhash!(sketch::MinHashSketch, seq::BioSequence)` that adds a new sequence to `sketch`?

src/seq/minhash.jl
```diff
@@ -0,0 +1,115 @@
...
+Base.done(s::MinHashSketch, state) = done(s.sketch, state)
+
+function Base.:(==)(a::MinHashSketch, b::MinHashSketch)
+    a.kmersize == b.kmersize && a.sketch == b.sketch
+end
+
+# A seqence and its reverse complement should be the same, so take the smallest
+# hash of a seq or its reverse complement.
+function revcomphash(kmer::Kmer)
+    k = min((kmer, reverse_complement(kmer)))
```
bicycle1885: `k = min(kmer, reverse_complement(kmer))`?

src/seq/minhash.jl
```diff
@@ -0,0 +1,115 @@
...
+* .kmersize: the length of kmers used to generate the sketch
+"""
+immutable MinHashSketch
+    sketch::Vector{UInt64}
+    kmersize::Int
+
+    function MinHashSketch(sketch::Vector, kmersize::Int)
+        length(sketch) > 0 || error("Sketch cannot be empty")
+        kmersize > 0 || error("Kmersize must be greater than 0")
+        new(sketch, kmersize)
```
bicycle1885: Use `return` when returning a value to clarify your intention.

src/seq/minhash.jl
```diff
@@ -0,0 +1,115 @@
...
+end
+
+Base.getindex(s::MinHashSketch, part) = getindex(s.sketch, part)
+Base.length(s::MinHashSketch) = length(s.sketch)
+Base.start(s::MinHashSketch) = start(s.sketch)
+Base.next(s::MinHashSketch, state) = next(s.sketch, state)
+Base.done(s::MinHashSketch, state) = done(s.sketch, state)
+
+function Base.:(==)(a::MinHashSketch, b::MinHashSketch)
+    a.kmersize == b.kmersize && a.sketch == b.sketch
```
bicycle1885: `return`

src/seq/minhash.jl
```diff
@@ -0,0 +1,115 @@
...
+
+    return kmerhashes
+end
+
+"""
+    minhash(seq, k::Int, s::Int)
+
+Generate a MinHash sketch of size `s` for kmers of length `k`.
+"""
+function minhash(seq::BioSequence, k::Int, s::Int)
```
bicycle1885: Using `Integer` for `k` and `s` would be better.

src/var/mash.jl
```diff
@@ -0,0 +1,52 @@
...
+            while sketch1.sketch[i] < sketch2.sketch[j]
+                i += 1
+            end
+        elseif sketch2.sketch[j] < sketch1.sketch[i]
+            while sketch2.sketch[j] < sketch1.sketch[i]
+                j += 1
+            end
+        end
+    end
+    matches == sketchlen ? (return 1.0) : return matches / (2 * sketchlen - matches)
```
bicycle1885: Using the if-else clause would be more readable.

src/seq/minhash.jl
kescobo: Is this also true for other places where `Int` is passed as the type parameter? 

Also, if you don't mind explaining, is `Integer` better because it's more generic?

src/var/mash.jl
kescobo: I agree, though I thought using the ternary operator is more "standard" in julia.

It's worth noting that this syntax is currently in a bunch of places around the Bio.jl code. Admittedly in most cases, they're in quite short and clear places, but in others, not so much eg in banded_needleman_wunch.jl:

```julia
@update2_nextcol (i == m ? end_gap_init_a : middle_gap_init_a) (i == m ? end_gap_extend_a : middle_gap_extend_a)
```

As someone coming from python, I was really confused by this syntax initially, and it's not easy to find info about it. I fully support using if-else in most circumstances, but in that case we should probably be more consistent (and maybe add something to CONTRIBUTING.md if we're breaking from standard julia style).

src/seq/minhash.jl
bicycle1885: Yes, that is the reason. I'm sometimes frustrated by this kind of restrictions. Most integer values can be converted to `Int`, so I think it would be reasonable to make argument types as loose as possible. This is also described in the style guide: http://docs.julialang.org/en/stable/manual/style-guide/#avoid-writing-overly-specific-types.

src/seq/minhash.jl
```diff
@@ -18,12 +18,12 @@ hashes for kmers of a given length. The type contains two parameters:
 """
 immutable MinHashSketch
     sketch::Vector{UInt64}
-    kmersize::Int
+    kmersize::Integer
```
bicycle1885: I think this should be kept as `Int`. Loosening the type of a field will propagate to other variables and result in performance degradation. In Julia, calling `new` automatically converts the types of arguments to those of fields. So, even if the constructor takes an `Integer` value, the value will be converted to `Int` when a sketch object is created.

kescobo: I think I see, but just to be sure, you're saying line 21 there should be `Int`, but the constructor on line 23 should be `Integer`, yes?

So as a general rule, type fields should be specific where possible, but function arguments should be more permissive?

bicycle1885: Yes and yes. Function arguments are for users, but field values are for internal use. Arguments should be generic for usability and fields should be specific (or parameterized by type parameters) for performance.

kescobo:
OK, I've implemented methods for `minhash()` to take multi-sequence (vectors or readers) inputs and generate sketches. This required a slight re-working of `kmerminhash()`, but I benchmarked it and it's the same speed as the old version on a ~4Mb genome making a 12mer sketch of 5000 ([see here](https://gist.github.com/kescobo/8bf9e64f1fecd6b3e456ef4bbeb03e17)). 

Oh, and when I opened this PR I said

>I have an implementation that I've validated and is quite fast - from two FASTA files of ~5 Mbp each to MASH distance takes ~2 sec on my 3 year old macbook pro. Most of this time is spent generating the sketch, but once a sketch is made, the comparison between two genomes takes less than a millisecond (so adding additional genomes scales very well)

I just repeated these procedures, and now going from the FASTA files to a MASH distance is ~10x faster (~250ms) as is getting MASH distance from the sketch (~13 μs). You guys are great! (thanks especially @bicycle1885).

kescobo:
Weee! It works! @bicycle1885 - I think I managed to address all of your concerns, right?

@Ward9250 I played around a bit with your suggested to make `k` a type parameter  for `MinHashSketch`, but the code you put in the gist didn't work (or at least I didn't get it to work).

kdmurray91:
Sorry for the delayed response all...

Regarding the term "sketch": In my mind, this is a generic term for a data structure that holds the essence of some dataset using as few resources as possible. These include bloom filters for presence/absence queries, hyperloglogs for cardinaltiy queries, Count-min sketches (*a la* countmin.jl) for counting queries, and min-hash or bottom sketches for "summary" queries.

So no, the MinHash datastructure is not really a good fit for CountMin.jl, nor is CountMin.jl useful for this PR. However, one thing I've been thinking of is making a SketchingDataStructures.jl package, that would hold a bloom filter, hyperloglog, min-hash and count-min sketch, with a consistent API, and importantly with good serialisation support via HDF5. PhD-related time poverty has prevented this thus far, sorry!

As to how mash fits in to all this, MASH's key (IMHO) innovation is applying min-hash sketches to alignment free seq. comparison. They use a metric derived from [AAF](http://bmcgenomics.biomedcentral.com/articles/10.1186/s12864-015-1647-5), which does a similar thing, but using jellyfish to count the entire dataset, not a subset. For full disclosure, I have a [paper](http://biorxiv.org/content/early/2016/10/04/075481) using a completely different metric derived from the D2 statistic, and using (what amount to) count-min sketches.

As I said, I'm currently writing a thesis on this field, so ping any Q's my way, and I'll try not to take so long to get back to you. I'll take a look through the code now, but it sounds as though @bicycle1885  as already been over it with a sharper eye than I could :)

src/seq/minhash.jl
```diff
@@ -0,0 +1,111 @@
...
+Base.next(s::MinHashSketch, state) = next(s.sketch, state)
+Base.done(s::MinHashSketch, state) = done(s.sketch, state)
+
+function Base.:(==)(a::MinHashSketch, b::MinHashSketch)
+    return a.kmersize == b.kmersize && a.sketch == b.sketch
+end
+
+# A seqence and its reverse complement should be the same, so take the smallest
+# hash of a seq or its reverse complement.
+function revcomphash(kmer::Kmer)
```
kdmurray91: Isn't that implented somewhere as `canonical(kmer::Kmer)`?

src/var/mash.jl
```diff
@@ -0,0 +1,57 @@
...
+# ========
+#
+# Functions to get MASH distance for MinHash sketches
+#
+# see DOI: 10.1186/s13059-016-0997-x
+#
+# This file is a part of BioJulia.
+# License is MIT: https://github.com/BioJulia/Bio.jl/blob/master/LICENSE.md
+
+function jaccarddistance(sketch1::MinHashSketch, sketch2::MinHashSketch)
```
kdmurray91: Coudn't this be obtained from Distances.jl? Then it could be `jaccarddistance(s1, s2) = Distances.jaccard(s1.sketch, s2.sketch)` or whatever

kescobo:
FYI - in light of #425, I've incorporated all of these changes into my original repo [here](https://github.com/kescobo/MashTun).

src/var/mash.jl
kescobo: Looks like the way Distances.jl implements this is `1 - sum(min(x, y)) / sum(max(x, y))`, which I think will be less efficient here. The version written here takes advantage of the fact that we know sketches are (1) sets, (2) the same length and (3) sorted. Besides, probably not worth adding another dependency for this one function right?

Ward9250: @kescobo I expected the Distances methods would be slower since your function is tailored to the type.

src/seq/minhash.jl
kescobo: Possibly, though now that I look at it, I realize that the original MASH paper did the lexicographic minimum - eg the lowest by alphabetical order. I didn't look at the `min()` function as it applies to kmers - might have to convert them to strings?

Ward9250: That would take more memory and add conversion cost though :(

kescobo: @Ward9250 I suspect we can- it would just require a custom sort order to make sure A < C < G < T.

But in any case, this is only necessary to make our sketches compatible with sketches generated by other software, and this is a bigger project that also requires the ability to select the specific hash function (see eg [here](https://github.com/marbl/Mash/issues/27)). I don't think it's necessary to implement in this PR.

kdmurray91: wouldn't it just be `hash(canonical(kmer))`? The canonical kmer *is* the lexographically smaller of the kmer & revcomp(kmer), by definition.

kescobo:
@kdmurray oh - well in that case, yes :-)

I'll take a look at it.

src/seq/minhash.jl
kescobo: @kdmurray91 fixed!

kescobo: @kdmurray oh - well in that case, yes :-)

I'll take a look at it.

EDIT: Fixed!

kescobo:
Just to be clear, I think this PR is ready pending @bicycle1885 checking that I addressed the concerns he raised in his review.

Ward9250:
Looks good @kescobo 👍

kdmurray91:
LGTM :+1:

Ward9250:
The NEWS update would have caused a conflict with the ABIF News update, so I rebased and fixed.

PR #420, opened 2017-03-14
Syntax changes: julia v0.6 Prep.

Ward9250:
This changes some syntax the new julia v0.6 syntax. This will want to be merged (assuming checks pass for 0.6), when julia 0.6 becomes the stable julia release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ os:
   - osx
 julia:
   - 0.5
+  - 0.6
   - nightly
 matrix:
   allow_failures:
--- a/docs/src/man/alignments.md
+++ b/docs/src/man/alignments.md
@@ -76,7 +76,7 @@ julia> Alignment([
 
 Alignment operations follow closely from those used in the [SAM/BAM
 format](https://samtools.github.io/hts-specs/SAMv1.pdf) and are stored in the
-`Operation` bitstype.
+`Operation` primitive type.
 
 | Operation            | Operation Type     | Description                                                                     |
 | :------------------- | :----------------- | :------------------------------------------------------------------------------ |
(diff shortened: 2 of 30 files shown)


sbromberger:
Hi all,

LightGraphs is preparing for a fairly large upgrade with Julia 0.6, and I wanted to make you aware of what we're doing. Very few things are breaking, but there will be some potential performance improvements that you might be able to take advantage of.

In a nutshell: we're abstracting `Graph` and `DiGraph`, renaming them to `SimpleGraph` and `SimpleDiGraph`*, and parameterizing vertex indices to `<: Integer`. This will allow you to create, e.g.,  `SimpleGraph{UInt8}`s that will be much more space-efficient if you have graphs that are < 256 vertices.

*While we plan on more abstractions (e.g., for weighted graphs) in the future, for now `Graph` and `DiGraph` will continue to work and will continue to default to `SimpleGraph`, and the default parameterization for both `Simple(Di)Graph`s and `(Di)Graph`s will be `Int`.

Feel free to review the (very long) WIP at https://github.com/JuliaGraphs/LightGraphs.jl/pull/541 - your feedback would be greatly appreciated.

Ward9250:
@sbromberger This sounds great, I'm particularly looking forward to weighting or abstractions for associating data with graphs. You'll notice phylogeny types in Bio are currently parametrised for the data they will store for a edge or node - currently that includes lengths and and weights as well as absolutely anything else. I can imagine graphs supporting weights and lengths of edges, but then any other data being dealt with by Phylo, as other data is out of the domain of a pure graph library.

Following issue #425, I'm moving Bio.Phylo over to BioJulia/Phylogenetics.jl, so if you could keep us posted there that would be awesome.

Ward9250:
I'm going to close this as I think it is superseded by #425, new syntax and or compat should be added to packages as they are migrated out from Bio.jl.

PR #426, opened 2017-03-20
implement eachoverlap

bicycle1885:
This is basically the same thing as `intersect` but some details are different:
1. More generic: `eachoverlap(intervals_a, intervals_b)` accepts any sequence of elements that are convertible to `Interval`. That means this is applicable to new readers that generate record objects.
2. Different name: The function name is different. There may be a better name than `eachoverlap` but the name `intersect` is already used in Base in a different way.
3. Intuitive iteration order: Both `intersect` and `eachoverlap` iterate over overlapping intervals of `intervals_a` and `intervals_b`. However, the iterator of `intersect` changes the left intervals faster while `eachoverlap` changes the right intervals faster. The iteration order of `eachoverlap` is same as a nested loop like below:
```julia
function alloverlaps(intervals_a, intervals_b)
    ret = Tuple{eltype(intervals_a),eltype(intervals_b)}[]
    for interval_a in intervals_a, interval_b in intervals_b
        if isoverlapping(interval_a, interval_b)
            push!(ret, (interval_a, interval_b))
        end
    end
    return ret
end
```

Also, the implementation is simpler (I think) and the performance is slightly better:
```
length(intervals) = 314925
--- intersect ---
  0.746044 seconds (2.67 M allocations: 104.005 MB, 12.54% gc time)
  0.809606 seconds (2.67 M allocations: 103.998 MB, 12.25% gc time)
  0.706502 seconds (2.67 M allocations: 103.998 MB, 13.47% gc time)
  0.734066 seconds (2.67 M allocations: 103.998 MB, 14.39% gc time)
  0.751998 seconds (2.67 M allocations: 103.998 MB, 13.78% gc time)
--- eachoverlap ---
  0.490069 seconds (3.56 M allocations: 103.990 MB, 20.92% gc time)
  0.602231 seconds (3.56 M allocations: 103.990 MB, 24.75% gc time)
  0.580232 seconds (3.56 M allocations: 103.990 MB, 27.17% gc time)
  0.580102 seconds (3.56 M allocations: 103.990 MB, 25.97% gc time)
  0.655486 seconds (3.56 M allocations: 103.990 MB, 30.58% gc time)

```


src/intervals/queue.jl
```diff
@@ -0,0 +1,72 @@
+# Queue
+# =====
+
+type Queue{T}
```
Ward9250: What is the purpose of Queue? Is this Bio specific? If so is there a better name than Queue, which makes one think of DataStructures.jl

bicycle1885: This is a usual queue data structure not specific to Bio.jl. This is very similar to `Bio.Intervals.StreamBuffer` but internals are slightly different (for example, `StreamBuffer` is index-oriented but `Queue` is iterator-oriented, which makes the algorithm simpler.) I've tried `Queue` of DataStructures.jl but it is not so efficient as I expected and requires a few patches to use.

bicycle1885:
> If this works for absolutely any interval which <: AbstractInterval, then I'm wondering if this belongs in IntervalTrees instead, as it defines abstract interval, and intervals in Bio are Genomic Intervals more specifically.

This includes some Bio.Intervals-specific code (e.g. seqname, metadata) and cannot be included in IntervalTrees.jl.

bicycle1885:
I'm going to merge this since the tests passed. I'm not completely satisfied with the functionality but this is a progress.

PR #432, opened 2017-03-22
fix some Bio.Var tests

bicycle1885:
I didn't notice but large amounts of tests of Bio.Var are inside a block comment. This seems to be introduced in this commit: https://github.com/BioJulia/Bio.jl/commit/3ed91d20d140b2385aa5673bbf949700eb62fec8. Unfortunately, some tests didn't pass and I'm not sure how I should fix them. So, I narrowed the range of the comment and fixed a few test I could.

@Ward9250 Could you take care of that after change?

### DIFF
--- a/test/var/runtests.jl
+++ b/test/var/runtests.jl
@@ -6,6 +6,8 @@ using Bio: Seq, Var
 using TestFunctions
 using PairwiseListMatrices
 using IntervalTrees: IntervalValue
+import BufferedStreams: BufferedInputStream
+import YAML
 
 typealias PWM PairwiseListMatrix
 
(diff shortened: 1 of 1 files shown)


Ward9250:
@bicycle1885 Yes this should relate to distance computation methods which are currently not exported until I deal with them to make them compatible with new site counting api in Var.

The thing is, I've come to see some of the site counting code in Var, should be in Seq, with Var extending it a bit by adding new site types. Additionally, in an older branch in which I was working on bit-flipping methods to count mutations in a bit parallel manner, I see scope for extending the Seq code for doing the bitpar_mismatches. So as my bit_parallel mutation counting methods can slot in there! So there's a few loose ends of this little bit of my postdoc I have to tie together soon.

PR #436, opened 2017-03-29
Site counting framework completion

Ward9250:
This completes a site counting framework for Seq which has both naive methods and faster bit-flipping data-parallel methods. And is extended by Var for sites that constitute mutations.

This will clean up and finish work done in #334 and previous PR's that deal with counting sequence sites.

This should be finished and merged before Bio.jl decomposition occurs - or at least before Seq and Var are decomposed.


Ward9250:
This is getting very close now :) extending bit-parallel counting from Mismatches to Mismatches, Matches, Certain, Ambiguous, and Gap sites is now working.

Ward9250:
Benchmarks are looking impressive:

image

bicycle1885:
What does `dist.dna` stand for?

Ward9250:
`dist.dna` is an algorithm in R's 'APE' package for computing the number of mismatches, or other measures of similarity/dissimilarity, include JC69, K80, logdet and so on. So I benchmarked that with R's microbenchmark package. 

'APE' uses a binary format for nucleotides which makes comparisons between individual nucleotides faster than for if the DNA was encoded in a typical string.

However, their binary sequences use a byte per nucleotide and no succinct encodings. So predictably, it is typically about as fast as BioJulia's naive algorithm (sometimes faster), but much slower than the bit-parallel algorithm(s) I have been working on.

Before these algorithms, dist.dna would be a common go-to for me and my colleagues to calculate genetic distance measures, prior to phylogeny building or any other analysis which takes evolutionary relationship into account.

bicycle1885:
Cool! Bit-parallelism seems to be very important here.

Ward9250:
Thanks :) :+1: I have a few more changes to make, and some cleaning up to do, and this is then ready to go in.

Ward9250:
Ok, back on track with getting this API tidied and merged - I found several errors of the origional bitpar_mismatches function, on which my work is based, which contain serious errors in a few edgecases, some would crash, some would not crash but give the wrong answer, I have fixed these edge cases and can continue finishing this off.

Ward9250:
Ok I think this works now. Bio.Var extended functionality needs cleaning up a bit, but I shall do that separately. I'd like to merge this this evening and break out Bio.Seq to BioSequences.jl, which I've completed and readied for upload except the files modified in this PR.

kdmurray91:
whoohoo!

PR #399, opened 2017-03-01
SAM rewrite

bicycle1885:
This is a rewrite of the SAM reader we already have in `Bio.Align`, which introduces the idea I proposed in #382 and the performance gain is huge as follows:

```
~/.j/v/Bio (sam-rewrite|…) $ julia test.jl
scan("data.sam") = 1000000
  0.975254 seconds (2.79 k allocations: 174.734 KB)
  0.970037 seconds (2.79 k allocations: 174.734 KB)
  1.007919 seconds (2.79 k allocations: 174.734 KB)
~/.j/v/Bio (sam-rewrite|…) $ git co master
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
~/.j/v/Bio (master|…) $ julia test.jl
INFO: compiling FASTA
scan("data.sam") = 1000000
  5.003223 seconds (63.54 M allocations: 3.203 GB, 3.33% gc time)
  4.980728 seconds (63.54 M allocations: 3.203 GB, 4.46% gc time)
  5.058925 seconds (63.54 M allocations: 3.203 GB, 4.37% gc time)
```

Now it gets ~5x faster! What it does is just verifying and indexing each line of a file and actual parsing (e.g. converting an alignment position data into an integer value) is delegated to accessor methods.

Also, I'd like to change the `SAMHeader` type as a sequence of `SAMMetaInfo`s with useful methods. This is much simpler and intuitive to users.

As usual, comments are welcome.

- [x] Rewrite Ragel code with Automa.jl.
- [x] Add some useful methods to `SAMHeaders` (e.g. find a specific tag).
- [x] Fix tests.
- [x] Update docs.
- [x] Update NEWS.md.

PR #440, opened 2017-04-06
add SAM and BAM modules

bicycle1885:
* Define Bio.Align.SAM and Bio.Align.BAM modules.
* Recover old SAM removed in #399 for backward compatibility.
* Revise accessors and docs.


bicycle1885:
I'd like to merge this today if no further comments.

PR #441, opened 2017-04-06
do not print newlines in sequence

bicycle1885:
I found newlines are noise.


kdmurray91:
Couldn't this behaviour be controlled by a keyword argument?

bicycle1885:
@kdmurray91 It makes sense. I've addded `width` keyword to specify the width of a sequence print. So, you can write like `print(STDOUT, dna"ACGT"^100, width=50)`. The default behavior does not print any newlines.

PR #442, opened 2017-04-07
Export the ACGTN constant

kdmurray91:
Since the ACG{T,U} constants are exported, I feel ACGTN should be. That, and I was going mad trying to work out why I coudn't see it in the repl.

K


bicycle1885:
Sounds reasonable. Can you add some tests like [this way](https://github.com/BioJulia/Bio.jl/blob/master/test/seq/runtests.jl#L510-L524)?

Issue #425, opened 2017-03-20
Migrating to modular software organization


BenJWard:
It has been requested by the community that the organisation of software in BioJulia should be made more modular than it currently is. Both for the benefit of both users and developers. We are reaching a point in Bio.jl where Travis does not complete, and anyone wanting one module of Bio.jl has to install and compile/pre-compile the rest of Bio.jl. 

Below I migrate all conversations over from Gitter, and I outline a draft set of guidelines for maintainers and contributors about the organisation of packages.

Ward9250:
@mkborregaard
> @Ward9250 I had another question I'd like to ask you - is Phylo bound to being a submodule of Bio? It seems to me that phylo does not depend on Bio functionality - so a possible model would be to have Phylo live outside but be imported by Bio. Here is the reason I ask - a lot of my ecology code depends on phylogenies, but uses no other Bio functionality. I don't care (I also use Bio things) but I want the SpatialEcology package to use phylogenies but not incur all of Bio as a dependency. I think this is probably not just me - phylogenies are used in much broader fields than DNA stuff.
> I hope it isn't impolite to ask this question here

Ward9250:
@Ward9250

> The same argument can be made for any of the other Bio submodules - for example if one only wanted to use sequences.
> Once you start asking that about the different packages the entire purpose of a Bio.jl package get's called into question. And whether modular vs monolithic design is better, and you can make the case for either design pattern - the most successful piece of monolithic software being the linux kernel.

Ward9250:
@ChrisRackauckas

> @Ward9250 @mkborregaard FWIW, DifferentialEquations.jl went the same direction because of these kinds of requests. But it does have more interdependencies, hence DiffEqBase.jl as a common small dependency. And then DifferentialEquations.jl just reexports the functionality from the various packages.

Ward9250:
@Ward9250

> The main argument for Phylo being in Bio, or indeed all of the modules being in Bio is that we can more easily make sure everything in one module of Bio is compatible with another - One of the reasons development of Phylo has been slow is I'm trying to design it at the same time as designing things to go in Var, because my plans fr storing and efficiently working with Terrabytes worth of trees also feed into y plans for working with both genetic variation, and uncertainties in pan-genomic graphs.
> Bioconductor has a similar design.
> Creating individual packages from Bio.jl submodules wouldn't mean the death of Bio.jl, as it could still exist as a metapackage binding common tools together for installation conveinience.

[...]

PR #443, opened 2017-04-09
Use BioSymbols.jl

bicycle1885:
This makes Bio.jl depend on BioSymbols.jl and minimize tests. I think more tests will be needed to check functions are surely exported as we intend. This is the first step toward #425.


bicycle1885:
Yes, I was wrong. Don't trust my words.

Ward9250:
Resolved conflicts. Ready to merge.

PR #445, opened 2017-04-10
Use Bio.Seq.mismatches in demulitplexer

kdmurray91:
Removes the less optimised hamming_distance function created for this package.

@bicycle1885, there is also a hamming distance implementation in [Bio.Align](https://github.com/BioJulia/Bio.jl/blob/master/src/align/pairwise/algorithms/hamming_distance.jl). Would it be wise to remove some of the duplication of such functions as part of #425?


Ward9250:
For Bio.Align that would depend since it may well make short-cuts using the CIGAR anchors to generate the distance, and so would be quite quick anyway.

bicycle1885:
Hamming distance defined in Bio.Align cannot be replaced with this because it should return mismatching positions as @Ward9250 pointed out.

kdmurray91:
@Ward9250 OK, was unsure about the Bio.Align stuff. Thanks both.

PR #446, opened 2017-04-11
Add @kmer_str macro to allow kmer string literals

kdmurray91:
Also adds documentation and a doctest.

### DIFF
--- a/src/seq/kmer.jl
+++ b/src/seq/kmer.jl
@@ -273,3 +273,21 @@ function swap{T,k}(kmer::Kmer{T,k}, i, j)
     x = ((b >> i) $ (b >> j)) & UInt64(0x03)
     return Kmer{T,k}(b $ ((x << i) | (x << j)))
 end
+
+# String literal
+# --------------
+
+macro kmer_str(seq, flag)
+    if flag == "s"
+        return DNAKmer(remove_newlines(seq))
+    elseif flag == "d"
+        return quote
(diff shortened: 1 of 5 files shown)


Ward9250:
Can the doctest be added as an actual test?

kdmurray91:
Should be good to go now @Ward9250

src/seq/kmer.jl
```diff
@@ -273,3 +273,21 @@ function swap{T,k}(kmer::Kmer{T,k}, i, j)
     x = ((b >> i) $ (b >> j)) & UInt64(0x03)
     return Kmer{T,k}(b $ ((x << i) | (x << j)))
 end
+
+# String literal
+# --------------
+
+macro kmer_str(seq, flag)
+    if flag == "s"
```
bicycle1885: There is no need to have this flag. `Kmer` is always immutable and there is no distinction of 'static' or 'dynamic'.

bicycle1885:
Can you mention this macro in NEWS.md?

src/seq/kmer.jl
kdmurray91: @bicycle1885 so remove the whole macro @ lines 280-289?

bicycle1885: Yes.

bicycle1885:
Thank you, @kdmurray91!

Issue #437, opened 2017-03-30
Lots of allocations when reading FASTQ files


tlnagy:
This is using the supposedly low allocation form from the docs:

```julia
function read_seq_file(filename)
    stream = open(filename)# |> ZlibInflateInputStream

    reader = FASTQReader(stream, quality_encoding=:illumina18)

    record = FASTQSeqRecord{DNASequence}()
    while !eof(reader)
        read!(reader, record)
        # perform some operation on `entry`
    end
    close(reader)
end

results = @benchmark read_seq_file("CTRL_T0_S17_R1_001.fastq")
```

gives

```
BenchmarkTools.Trial: 
  memory estimate:  3.84 GiB
  allocs estimate:  257729754
  --------------
  minimum time:     24.007 s (3.12% GC)
  median time:      24.007 s (3.12% GC)
  mean time:        24.007 s (3.12% GC)
  maximum time:     24.007 s (3.12% GC)
  --------------
  samples:          1
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
```

Versioninfo:
```
julia> Pkg.installed("Bio")
v"0.4.7"

julia> versioninfo()
Julia Version 0.5.0
Commit 3c9d753 (2016-09-19 18:14 UTC)
Platform Info:
  System: Darwin (x86_64-apple-darwin13.4.0)
  CPU: Intel(R) Core(TM) i7-4850HQ CPU @ 2.30GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.7.1 (ORCJIT, haswell)
```

bicycle1885:
IO APIs are under overhaul and restructuring. FASTQ is not yet finished, but the problem will be resolved as [FASTA](https://github.com/BioJulia/Bio.jl/pull/413#issuecomment-286026383).

tlnagy:
Awesome. Excited to see the new developments. Is there a PR for this?

bicycle1885:
Not yet. FASTQ is a little bit more difficult to write than other file formats using the current parser framework. This is because FASTQ is not a regular language. So, I may need to add some special handling for FASTQ.

kdmurray91:
[...]

PR #448, opened 2017-04-14
add FASTQ module

bicycle1885:
This introduces `Bio.Seq.FASTQ` module with the new parser design.

Summary of changes:
* Multiline sequence/quality is no longer supported as per https://github.com/BioJulia/Bio.jl/issues/437#issuecomment-293466753.
* Quality decoding is delegated to the `FASTQ.quality` function.
* I think #437 is solved:

```
using Bio.Seq
using BenchmarkTools

function read_seq_file(filename)
    stream = open(filename)
    reader = FASTQ.Reader(stream)
    record = FASTQ.Record()
    while !eof(reader)
        read!(reader, record)
    end
    close(reader)
end
```

```
julia> @benchmark read_seq_file("./SRR1238088_1.fastq")
BenchmarkTools.Trial:
  memory estimate:  130.28 KiB
  allocs estimate:  20
  --------------
  minimum time:     9.962 s (0.00% GC)
  median time:      9.962 s (0.00% GC)
  mean time:        9.962 s (0.00% GC)
  maximum time:     9.962 s (0.00% GC)
  --------------
  samples:          1
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%

````

### CODE REVIEW

**Comment by bicycle1885** [comment]
`FASTQ.Reader` is ~3.5x faster than [ShortRead](http://bioconductor.org/packages/release/bioc/html/ShortRead.html). Not bad, isn't it?
```
library(ShortRead)

read_seq_file <- function (filename) {
    strm <- FastqStreamer(filename)
    repeat {
        fq <- yield(strm)
        if (length(fq) == 0)
            break
    }
}

system.time(read_seq_file("./SRR1238088_1.fastq"))
system.time(read_seq_file("./SRR1238088_1.fastq"))
system.time(read_seq_file("./SRR1238088_1.fastq"))
```

```
   user  system elapsed
 27.003   9.062  38.226
   user  system elapsed
 25.772   8.359  35.735
Warning message:
closing unused connection 3 (./SRR1238088_1.fastq)
   user  system elapsed
 25.847   8.773  35.872
Warning message:
closing unused connection 3 (./SRR1238088_1.fastq)
```

File: src/seq/Seq.jl [hunk 1]
```diff
@@ -138,8 +138,7 @@ export
     AA_Gap,
     FASTA,
     FASTASeqRecord,
-    FASTQReader,
-    FASTQWriter,
+    FASTQ,
     FASTQSeqRecord,
```
Ward9250: FASTQSeqRecord and FASTASeqRecord to be removed? Depends on whether the old IO system should be included with the final monolithic 0.5 or not.

bicycle1885: I think it would be better to move these record types into deprecated.jl to deprecate them. I will do that in another pull request.

PR #451, opened 2017-04-16
add more FASTQ constructors

bicycle1885:
And move `FASTASeqRecord` and `FASTQSeqRecord` to deprecated.jl.


src/seq/fastq/record.jl
```diff
@@ -50,6 +50,37 @@ function Record(str::AbstractString)
     return convert(Record, convert(Vector{UInt8}, str))
 end
 
+"""
+    FASTQ.Record(identifier, sequence, quality; offset=33)
+
+Create a FASTQ record from `identifier`, `sequence` and `quality`.
+"""
+function Record(identifier::AbstractString, sequence, quality::Vector; offset=33)
+    return Record(identifier, nothing, sequence, quality, offset=33)
```
kdmurray91: Minor pedantry: on L59, shouldn't one use `offset=offset`?

bicycle1885: Oops, my mistake. Thank you, @kdmurray91.

PR #453, opened 2017-04-20
Fix UndefVarError in IntervalCollection(GFFReader)

kdmurray91:
And add a test.


kdmurray91:
Planning on merging this in the Aussie morning if there are no objections...

kdmurray91:
@Ward9250 I didn't know you could do that within the module itself. I'll do that and then merge.

PR #457, opened 2017-05-08
Promotion for DNA/RNA sequences

Ward9250:
This PR will add promotion to biological sequences in order to allow completion of #436.

Essentially the rule will be if you have DNA or RNA sequences and want to promote them to the same type, they will be promoted to the alphabet that is capable of representing all the possible elements, so `promote_rule(DNAAlphabet{2}, DNAAlphabet{4})` will return `DNAAlphabet{4}`, as `DNAAlphabet{4}` is the one that is capable of representing all elements of `DNAAlphabet{2} and DNAAlphabet{4}`.


Ward9250:
Tests are passing, I'd like to merge in the next hour or so if that's ok.

bicycle1885:
👍

PR #459, opened 2017-05-11
add Interval(seqname, range) constructor

bicycle1885:
I always expect this to work but never succeeded. So, I added a new constructor.


bicycle1885:
I think this is not a controversial change.

PR #463, opened 2017-06-09
check gap existence before making a kmer

bicycle1885:
https://github.com/BioJulia/BioSequences.jl/issues/3 cc: @Ward9250


Ward9250:
I think future modifications to Seq should be made over in BioSequences, as it is now current, and finishing the Seq site counting PRs I had open was a natural breakpoint for Seq to be Bio.jl 0.5 ready.

bicycle1885:
It makes sense. I'll reopen this there.

Ward9250:
:+1: Thanks for doing that!

PR #468, opened 2017-06-15
Complete Bio.Seq decomposition

Ward9250:
This PR adds BioSequences 0.6 as a dependency and completes the move of that code out to BioSequences.jl by removing the `seq` folder in `src`. Now Bio.Seq re-exports BioSequences.jl instead.


bicycle1885:
This doesn't pass the CI tests. Please do not merge changes before passing CI.

Ward9250:
As far as I could tell, this was due to tests for another submodule, I think Bio.Align (maybe Bio.Var too), which is being extracted to BioAlignments anyway, so I thought it was of little consequence.
I think it might be tricky to decompose some of the submodules without making the CI fail temporarily, where they depend on each other.

bicycle1885:
The failure you saw may be due to the change you added to test/align/runtests.jl. You removed whitespaces in tests but some tests are actually whitespace-aware. That may solve, at least part of, the CI failure.

PR #472, opened 2017-06-22
findfirst and haskey for IntervalCollection

dcjones:
Adds a `findfirst` and `haskey` function to `IntervalCollection` that both look for exact (start, end, and strand) matches.

### DIFF
--- a/docs/src/man/intervals.md
+++ b/docs/src/man/intervals.md
@@ -154,6 +154,9 @@ for a in intervals_a, b in intervals_b
 end
 ```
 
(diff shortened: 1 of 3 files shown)


dcjones:
Should I be making PRs like this against GenomicFeatures.jl? I kind of lost track of what the plan is.

Ward9250:
@dcjones Yeah GenomicFeatures.jl and IntervalTrees.jl (if applicable). As Bio.jl is being de-monolithed, and GenomicFeatures has had it's first METADATA release :)

dcjones:
Ok, cool. :+1:

PR #474, opened 2017-06-24
Remove Iterators.jl dependency (not used)

iamed2:
Doesn't look like Iterators is ever imported

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -10,7 +10,6 @@ Distributions
 EzXML 0.2
 IndexableBitVectors 0.1.1
 IntervalTrees 0.2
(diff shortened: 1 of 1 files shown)


bicycle1885:
Thank you @iamed2! I haven't noticed it. The CI failure is due to https://github.com/BioJulia/Bio.jl/pull/468#issuecomment-310745734 and would be unrelated to your change. I will merge once the CI failure is fixed.

iamed2:
Cool, thanks :) I am going through packages and renaming Iterators => IterTools and there was absolutely no work to be done here, haha

PR #477, opened 2017-06-28
Pass the CI

### LINKED ISSUE #476: CI failure
See https://github.com/BioJulia/Bio.jl/pull/468. I think we should make it work soon.

Ward9250:
Let's fix #476

### DIFF
--- a/test/align/runtests.jl
+++ b/test/align/runtests.jl
@@ -806,7 +806,7 @@ end
                 """)
 
                 testaln("""
-                 ACGT
+                 ACGT  
                 AACGTTT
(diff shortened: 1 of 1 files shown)


Ward9250:
So it turns out that the Atom editor removes trailing whitespace from lines, which is what caused the line changes in test/align in my previous PR.

bicycle1885:
Perhaps you need to replace helper functions of Bio.jl with those of BioCore.jl. Packages extend the latter but Bio.jl does not.

Ward9250:
This should be moot now.

PR #482, opened 2017-07-20
[WIP] Decompose Bio.Align and Bio.Intervals

Ward9250:
This will decompose Bio.Align and Bio.Intervals now that BioAlignments.jl and GenomicFeatures have release versions.


Ward9250:
This is now passing, also fixes #476.

Ward9250:
I'll merge shortly if no objections.

src/Align.jl
```diff
@@ -0,0 +1,15 @@
...
+# Module for biological sequence alignment.
+#
+# This file is a part of BioJulia.
+# License is MIT: https://github.com/BioJulia/Bio.jl/blob/master/LICENSE.md
+
+module Align
+
+using Reexport
+
+@reexport using BioAlignments
```
bicycle1885: Why not `const Align = BioAlignments`? I mean `Bio.Align` is an alias of `BioAlignments`.

Ward9250: Ok, I'll have to do that for the other modules too in that case, as up until now reexport has been used for them all.

Ward9250: Fixed :+1:

bicycle1885:
BTW, Phylogenies.jl is a blocker to support Julia 0.6 on Bio.jl. Any plan to support Julia 0.6?

Ward9250:
Absolutely, I will update to support 0.6

PR #483, opened 2017-08-03
[WIP] PDB file handling enhancements

joels94:
I have been working on few enhancements over handling PDB files based on BioPython library to be true.
Such as:
- Downloading PDB file to a specific directory instead of only current directory 
- Skip or Overwrite if the PDB file already exists in the directory
- Downloading Entire PDB database
- Downloading multiple PDB files by passing a list of PDB IDs
- Downloading all obsolete PDB entries from RCSB Server
- Updating a PDB directory based on weekly status file from RCSB Server
- Retrieve a PDB file (i.e. it will download the PDB file if it does not exists and read them)
- Obsolete PDB file handling (A directory named "obsolete" is automatically created inside the specified "pdb_dir" to store and handle them)


jgreener64:
Great, thanks for making this PR. I will have a look at this tomorrow.

joels94:
We have few things to discuss. 

**1. Whether the `out_filepath` argument in `downloadpdb()` is required?** 
Because removing it will give an uniformity in the code and easy for users. It would be better to keep the PDB ID as file name so that it will be easy to handle it. The user can specify a different `structure_name` when reading the file to differentiate the file as he wishes.

**2. Should we write test cases?**
The code has grown little complex than before. Especially there are lots of options and different combinations. This is may lead to unexpected bugs. 

**3. Should `getstatuslist()` function be exported in the module?**
As of now its just a helper function for `getrecentchanges()` function.

src/structure/pdb.jl
```diff
@@ -30,26 +39,231 @@ end
...
+
+
+"""
+Downloads a list of PDB files from the RCSB PDB server. 
+By default downloads the PDB files to current working directory; 
+if the keyword argument `pdb_dir` is set the PDB files are downloaded to the specified directory;
+if the keyword argument `obsolete` is set `true`, the PDB files are downloaded into the obsolete directory inside `pdb_dir`;
+if the keyword argument `overwrite` is set `true`, then it will overwrite the PDB file if it exists in the `pdb_dir`;
+"""
+function downloadmultiplepdb(pdbidlist::AbstractArray{String,1}; pdb_dir::AbstractString=pwd(), obsolete::Bool=false, overwrite::Bool=false)
```
jgreener64: This name seems a bit long. We could even make this another method of `downloadpdb`, so that function can take either a string or a list.

I think this should be `pdbidlist::Array{String,1}` for the first argument.

src/structure/pdb.jl
```diff
@@ -30,26 +39,231 @@ end
...
+end
+
+
+"""
+Downloads the entire PDB files available in the RCSB PDB server. 
+By default downloads the PDB files to current working directory; 
+if the keyword argument `pdb_dir` is set the PDB files are downloaded to the specified directory;
+if the keyword argument `overwrite` is set `true`, then it will overwrite the PDB file if exists in the `pdb_dir`
+"""
+function downloadentirepdb(;pdb_dir::AbstractString=pwd(), overwrite::Bool=false)
```
jgreener64: Perhaps an info line telling the user that this will take a long time and use a lot of disk space?

src/structure/pdb.jl
```diff
@@ -30,26 +39,231 @@ end
...
+    downloadmultiplepdb(getallpdbentries(), pdb_dir=pdb_dir, overwrite=overwrite)
+end
+
+
+"""
+Updates your local copy of the PDB files. It gets the weekly lists of new, 
+modified and obsolete pdb entries and automatically downloads those PDB files to the `pdb_dir`;
+if the keyword argument `pdb_dir` is set the PDB files in the specified directory are updated;
+"""
+function updatepdb(;pdb_dir::AbstractString=pwd())
```
jgreener64: Perhaps `updatelocalpdb`?

src/structure/pdb.jl
```diff
@@ -107,6 +321,17 @@ function Base.read(filepath::AbstractString,
     end
 end
 
+# Read PDB file based on PDB ID and pdb_dir. 
+function Base.read(pdbid::AbstractString,
```
jgreener64: This method clashes with the above `read` method. I wonder whether if it is necessary, but if it goes in the directory should be a full argument.

src/structure/pdb.jl
```diff
@@ -30,26 +39,231 @@ end
...
+            end
+        end
+    end
+    rm("obsolete.dat")
+    return obsoletelist
+end
+
+
+"""
+Download a PDB file or biological assembly from the RCSB PDB server.
```
jgreener64: Use line breaks in the doc strings to keep line length to 80.

src/structure/pdb.jl
```diff
@@ -30,26 +39,231 @@ end
...
+Returns a list of all PDB entries from RCSB PDB server
 """
-function downloadpdb(pdbid::AbstractString,
-                    out_filepath::AbstractString="$pdbid.pdb";
-                    ba_number::Integer=0)
+function getallpdbentries()
+    pdbidlist = Array{String,1}()
+    info("Fetching list of all PDB Entries from RCSB PDB Server...")
+    download("ftp://ftp.wwpdb.org/pub/pdb/derived_data/index/entries.idx","entries.idx")
+    open("entries.idx") do input
```
jgreener64: Given that we make then delete the file here, I wonder if we should use a temporary filepath? If the user has `entries.idx` in their directory here it is getting overwritten without warning. You can use `tempname()` to save a name to a variable, write to this then delete it (I think there are examples of this in the PDB tests).

src/structure/pdb.jl
```diff
@@ -30,26 +39,231 @@ end
...
+            linecount +=1
+        end
+    end
+    rm("entries.idx")
+    return pdbidlist
+end
+
+
+"""
+Returns a list of pdb codes in the weekly pdb status file from the given URL.
```
jgreener64: Pedantry - PDB should be capital in docstrings.

src/structure/pdb.jl
```diff
@@ -30,26 +39,231 @@ end
...
+    end
+    rm("entries.idx")
+    return pdbidlist
+end
+
+
+"""
+Returns a list of pdb codes in the weekly pdb status file from the given URL. 
+"""
+function getstatuslist(url::AbstractString)
```
jgreener64: What URL is the user supposed to use here? If there is a certain one perhaps put it in the docstring?

src/structure/pdb.jl
```diff
@@ -30,26 +39,231 @@ end
...
+
+"""
+Download a PDB file or biological assembly from the RCSB PDB server and reads it. 
+By default downloads the PDB file to current working directory; 
+if the keyword argument `ba_number` is set the biological assembly with that number will be downloaded; 
+if the keyword argument `pdb_dir` is set the PDB file is downloaded to the specified directory; 
+if the keyword argument `obsolete` is set `true`, the PDB file is downloaded into the obsolete directory inside `pdb_dir`;
+if the keyword argument `overwrite` is set `true`, then it will overwrite the PDB file if it exists in the `pdb_dir`;
+"""
+function retrievepdb(pdbid::AbstractString;
```
jgreener64: This is a nice function to have, difficult to find a descriptive name for it though. A descriptive one would be `downloadpdbandread` but that is way too long. Maybe this name is okay.

jgreener64:
So overall looks good, thanks for adding this useful feature. There are a couple more things but the above is most of my issues. For reference the Biopython implementation is here: https://github.com/biopython/biopython/blob/master/Bio/PDB/PDBList.py

In answer to your specific questions:

src/structure/pdb.jl
jgreener64: Actually see comment below.

jgreener64:
1. `out_filepath` in my initial implementation was generally meant for the user to specify a directory rather than filename. Since you are adding the extra directory argument you can remove `out_filepath` for uniformity of filenames.

2. Yes, there should be some tests. These will rely on an internet connection but I think that is okay. Obviously we cannot download the whole PDB in the tests but certainly on the other functions.

3. Maybe don't export it, if we do then there should be information on its use in the docstring.

jgreener64:
There is one more thing that could be added now you are looking at this code, namely download of mmCIF and MMTF files from the PDB. Since mmCIF is now the standard, this is an important feature (I am writing an mmCIF parser for `Bio.Structure` now too). Would you be okay to implement this as a `file_format` (or similar) keyword argument to `downloadpdb`?

src/structure/pdb.jl
```diff
@@ -30,26 +39,231 @@ end
...
 """
-Download a Protein Data Bank (PDB) file or biological assembly from the RCSB
-PDB. By default downloads the PDB file; if `ba_number` is set the biological
-assembly with that number will be downloaded.
+Returns a list of all PDB entries from RCSB PDB server
 """
-function downloadpdb(pdbid::AbstractString,
-                    out_filepath::AbstractString="$pdbid.pdb";
-                    ba_number::Integer=0)
+function getallpdbentries()
```
joels94: That should be fine. I will change it accordingly.

src/structure/pdb.jl
joels94: Nice. I didn't know about it. thank you.

src/structure/pdb.jl
joels94: Yes. K.

src/structure/pdb.jl
joels94: The URL is supposed be weekly status file from RCSB Server. I will add it in the docstrings.

src/structure/pdb.jl
```diff
@@ -30,26 +39,231 @@ end
...
+    end
+    rm(filename)
+    return statuslist
+end
+
+
+"""
+Returns three lists of the newest weekly files (added,modified,obsolete) from RCSB PDB Server
+"""
+function getrecentchanges()
```
joels94: Yes. K

src/structure/pdb.jl
```diff
@@ -30,26 +39,231 @@ end
...
+    modifiedlist = getstatuslist("ftp://ftp.wwpdb.org/pub/pdb/data/status/latest/modified.pdb")  
+    obsoletelist = getstatuslist("ftp://ftp.wwpdb.org/pub/pdb/data/status/latest/obsolete.pdb")  
+    return addedlist, modifiedlist, obsoletelist
+end
+
+
+"""
+Returns a list of all obsolete entries ever in the RCSB PDB server
+"""
+function getallobsolete()
```
joels94: Yes. K

src/structure/pdb.jl
joels94: K. I had this doubt. I will keep the line length to 80 in docstrings.

src/structure/pdb.jl
joels94: Nice. I will override the `downloadpdb` for downloading multiple PDB files. Regarding `pdbidlist::Array{String,1}` I was getting an error, defining it as `AbstractArray` was only working. I will look into it and update accordingly.

src/structure/pdb.jl
joels94: That would be useful. I will add it.

src/structure/pdb.jl
joels94: That would be clear. I will update accordingly.

src/structure/pdb.jl
joels94: Ya. I was so confused. Then finally decided keep it as in BioPython. So if BioPython users are using this, they may find it easy. As mentioned `downloadpdbandread` will be too long. We will keep it as `retrievepdb` as of now. Any other good function name is welcomed.

src/structure/pdb.jl
joels94: Actually I just added it because that will be in a uniform format as the rest. If users wants to download first and then read it later, they will have uniform way of calling the functions. We may change the arguments as you mentioned.

joels94:
@jgreener64 Thank you so much for your review. I will update the code accordingly in the future commits. Adding option `file_format` in `downloadpdb` will be useful. Nice that you are working on mmCIF parser. We can discuss further changes as the code grows.

src/structure/pdb.jl
jgreener64: One possibility is `readfrompdb`.

src/structure/pdb.jl
```diff
@@ -30,26 +39,231 @@ end
...
+    rm("entries.idx")
+    return pdbidlist
+end
+
+
+"""
+Returns a list of pdb codes in the weekly pdb status file from the given URL. 
+"""
+function getstatuslist(url::AbstractString)
+    statuslist = Array{String,1}()
```
jgreener64: I think `statuslist = String[]` is better, it seems to allocate less (this applies to all similar ones below too).

src/structure/pdb.jl
```diff
@@ -30,26 +39,231 @@ end
...
+    failedlist = Array{String,1}()
+    for pdbid in pdbidlist
+        try
+            downloadpdb(pdbid, pdb_dir=pdb_dir, obsolete=obsolete, overwrite=overwrite)
+        catch
+            warn("Error downloading PDB file : $pdbid")
+            push!(failedlist,pdbid)
+        end
+    end
+    warn(length(failedlist)," PDB file failed to download : ", failedlist)
```
jgreener64: Only do this if `length(failedlist) > 0`.

src/structure/pdb.jl
joels94: K. I didnt know that

src/structure/pdb.jl
joels94: Yes. I missed it

joels94:
As discussed I have made all the recommended changes. Also added option to download and update files in PDB, XML and mmCIF formats. Let me know if there is something I missed or to be modified or added. I didn't still write test cases. I thought of refining the code before righting test cases.

jgreener64:
Thanks @joels94 , I will look at this tomorrow or Monday.

joels94:
- PDB, XML, mmCIF files are now downloaded in compressed format to reduce internet usage.
- PDB files can now be downloaded and updated in MMTF format also.

src/structure/pdb.jl
```diff
@@ -30,26 +46,314 @@ end
...
+if the keyword argument `read_std_atoms` is set false, then standard 
+ATOM records wont be parsed
+if the keyword argument `read_het_atoms` is set false, then hetro 
+HETATOM records wont be parsed
+"""
+function readpdb(pdbid::AbstractString;
+            pdb_dir::AbstractString=pwd(),
+            ba_number::Integer=0,
+            structure_name::AbstractString="$pdbid.pdb",
+            remove_disorder::Bool=false,
```
jgreener64: These 3 arguments can be replaced by `kwargs...` and then pass that to the inner read function (this means the defaults for `remove_disorder` etc. are defined in one place only).

src/structure/pdb.jl
```diff
@@ -30,26 +46,314 @@ end
...
+            read_std_atoms::Bool=true,
+            read_het_atoms::Bool=true)
+    filepath = joinpath(pdb_dir,"$pdbid.pdb")
+    pdbpath = ba_number == 0 ? filepath : filepath*"$ba_number"
+    open(pdbpath, "r") do input
```
jgreener64: This `open` is not actually required as calling `read` with the filepath is already defined below.

src/structure/pdb.jl
```diff
@@ -30,26 +46,314 @@ end
...
+    downloadpdb(pdbid, pdb_dir=pdb_dir, obsolete=obsolete, overwrite=overwrite, ba_number=ba_number)
+    info("Parsing the PDB file...")
+    if obsolete
+        # if obsolete is set true, the PDB file will downloaded to obsolete directory inside "pdb_dir"
+        filepath = joinpath(pdb_dir,"obsolete","$pdbid.pdb")
+    else    
+        filepath = joinpath(pdb_dir,"$pdbid.pdb")
+    end
+    pdbpath = ba_number == 0 ? filepath : filepath*"$ba_number"
+    open(pdbpath, "r") do input
```
jgreener64: This `open` is not actually required as calling `read` with the filepath is already defined below.

joels94:
As discussed I have made the recommended code fixes.  

**TO DO**

- [ ] Write test cases

- [ ] Add documentation

Meanwhile, as mentioned by @jgreener64 , it would be nice to get feedback from someone else.

src/structure/pdb.jl
```diff
@@ -30,26 +49,315 @@ end
...
+Downloads a list of PDB files from the RCSB PDB server. 
+By default downloads the PDB files to current working directory; 
+if the keyword argument `pdb_dir` is set the PDB files are downloaded to the
+specified directory;
+if the keyword argument `obsolete` is set `true`, the PDB files are downloaded
+into the obsolete directory inside `pdb_dir`;
+if the keyword argument `overwrite` is set `true`, then it will overwrite the
+PDB file if it exists in the `pdb_dir`.
+"""
+function downloadpdb(pdbidlist::Array{String,1}; pdb_dir::AbstractString=pwd(), file_format::Type=PDB, obsolete::Bool=false, overwrite::Bool=false)
```
jgreener64: The last 4 arguments here could just be `kwargs...` so their defaults are defined in one place above.

src/structure/pdb.jl
```diff
@@ -30,26 +49,315 @@ end
...
+                download("http://files.rcsb.org/download/$pdbid-assembly$ba_number"*pdbextension[file_format]*".gz", archivefilepath)
+            else
+                warn("Biological Assembly is available only in PDB and mmCIF formats!")
+            end
+        end
+        # Verify if the compressed PDB file is downloaded properly and extract it. For MMTF no extraction is needed
+        if ispath(archivefilepath) && filesize(archivefilepath) > 0 && file_format != MMTF           
+            input = open(archivefilepath) |> ZlibInflateInputStream
+            open(pdbpath,"w") do output
+                for line in eachline(input)
```
jgreener64: This prints each newline character twice so the files alternate line/blank line. Could use `print` rather than `println`.

src/structure/pdb.jl
```diff
@@ -30,26 +49,315 @@ end
...
+                end
+            end
+            close(input)
+        end
+        # Verify if the PDB file is downloaded without any error
+        if !ispath(pdbpath) || filesize(pdbpath)==0
+            throw(ErrorException("Error downloading PDB : $pdbid"))
+        end
+    end
+    rm(archivefilepath)
```
jgreener64: If the condition is satisfied on line 180 then this will error as the file is not created. Either check the file exists here or move it up in the logic to where you know it exists.

src/structure/pdb.jl
```diff
@@ -30,26 +49,315 @@ end
...
-    if ba_number == 0
-        download("http://www.rcsb.org/pdb/files/$pdbid.pdb", out_filepath)
+    archivefilepath = tempname()
+    # The PDB file name will end with the ba_number when downloading biological assembly
+    if ba_number==0
+        pdbpath = joinpath(pdb_dir,"$pdbid"*pdbextension[file_format])
     else
-        # Will download error page if ba_number is too high
-        download("http://www.rcsb.org/pdb/files/$pdbid.pdb$ba_number", out_filepath)
+        pdbpath = joinpath(pdb_dir,"$pdbid"*pdbextension[file_format]*"$ba_number")
```
jgreener64: Thinking a bit more about it, the biological number should not change the extension of the file. Ideally it would change the filename, e.g. `1ABC_ba1.pdb` or something like that rather than `1ABC.pdb1`.

src/structure/pdb.jl
joels94: I don't think so. The `archivefilepath` exists independent of the if condition in line 180. Its because, `archivefilepath=tempname()` in line 172 actually creates an empty temporary file instead of just the path.

jgreener64: Interesting, it doesn't do that on my machine. I wonder if that is system specific. For example if I attempt to overwrite when overwrite is false I get

```
INFO: PDB Exists : 1AKE
ERROR: unlink: no such file or directory (ENOENT)
 in unlink(::String) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in #rm#7(::Bool, ::Bool, ::Function, ::String) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in #downloadpdb#1(::String, ::Type{T}, ::Bool, ::Bool, ::Int64, ::Function, ::String) at ./REPL[1]:61
 in (::#kw##downloadpdb)(::Array{Any,1}, ::#downloadpdb, ::String) at ./<missing>:0
```

joels94: Ok. Can you check by trying `ispath(tempname())`? For me i m getting true in Windows 10

joels94: Julia 0.6

jgreener64: I get `false` in Mac OSX, Julia 0.6. Searching this on Julia turns up https://github.com/JuliaLang/julia/issues/9053 which addresses this.

I guess check it exists and if so remove, which will work either way.

joels94: Yes. Ok.

src/structure/pdb.jl
joels94: I think this is also a system specific issue? Because when using `println()` the file is generated properly and I can read them. But if I use `print()`, all lines are getting concatenated and I cannot read the file. I have attached the two files for your reference. (Renamed the extensions as .txt as I was not able upload in .pdb format). Let me know what you find.

[1ENT_print().txt](https://github.com/BioJulia/Bio.jl/files/1209563/1ENT_print.txt)
[1ENT_println().txt](https://github.com/BioJulia/Bio.jl/files/1209564/1ENT_println.txt)

jgreener64: Thanks for this. It seems to be a Julia 0.5/0.6 issue, see the first breaking change for Julia v0.6.0 here: https://github.com/JuliaLang/julia/blob/master/NEWS.md .

In 0.5 the line break is not removed by eachline, in 0.6 it is. In order to be compatible with both I think we can use `eachline(..., chomp=false)` and use `print` inside.

joels94: Ok. Guess, I should keep an eye on the release notes as Julia changes a lot in each version. Thank you. I will update accordingly.

src/structure/pdb.jl
```diff
@@ -30,26 +49,326 @@ end
...
+    download("ftp://ftp.wwpdb.org/pub/pdb/derived_data/index/entries.idx",tempfilepath)
+    open(tempfilepath) do input
+        # Skips the first two lines as it contains headers
+        linecount = 1
+        for line in eachline(input)
+            if linecount > 2
+                # The first 4 characters in the line is the PDB ID
+                pdbid = uppercase(line[1:4])
+                # Check PDB ID is 4 characters long and only consits of alphanumeric characters
+                if length(pdbid) != 4 || ismatch(r"[^a-zA-Z0-9]", pdbid)
```
bicycle1885: `r"^[a-zA-Z0-9]{4}$"` will check the length as well.

src/structure/pdb.jl
```diff
@@ -30,26 +49,326 @@ end
...
+                    throw(ArgumentError("Not a valid PDB ID: \"$pdbid\""))
+                end
+                push!(pdbidlist,pdbid)
+            end
+            linecount +=1
+        end
+    end
+    rm(tempfilepath)
```
bicycle1885: `rm(tempfilepath, force=true)` would be better because it works even when the file does not exist.

bicycle1885: Also, to make sure that the temporary file will be deleted, you need to use the try-catch-finally statement.

src/structure/pdb.jl
```diff
@@ -30,26 +49,326 @@ end
...
+        # Download the compressed PDB file to the temporary location
+        info("Downloading PDB : $pdbid")
+        if ba_number == 0            
+            if file_format == PDB || file_format == PDBXML || file_format == mmCIF
+                download("http://files.rcsb.org/download/$pdbid"*pdbextension[file_format]*".gz", archivefilepath)
+            elseif file_format == MMTF
+                # MMTF is downloaded in uncompressed form, thus directly stored in pdbpath
+                download("http://mmtf.rcsb.org/v1.0/full/$pdbid", pdbpath)
+            else
+                warn("Invalid PDB file format!")
```
bicycle1885: This should throw an `ArgumentError` exception.

src/structure/pdb.jl
```diff
@@ -30,26 +49,326 @@ end
...
+            end
+        else            
+            if file_format == PDB
+                download("http://files.rcsb.org/download/$pdbid"*pdbextension[file_format]*"$ba_number.gz",archivefilepath)
+            elseif file_format == mmCIF
+                download("http://files.rcsb.org/download/$pdbid-assembly$ba_number"*pdbextension[file_format]*".gz", archivefilepath)
+            else
+                warn("Biological Assembly is available only in PDB and mmCIF formats!")
```
bicycle1885: `ArgumentError`.

src/structure/pdb.jl
```diff
@@ -30,26 +49,326 @@ end
...
+            end
+        end
+    end
+    rm(tempfilepath)
+    return obsoletelist
+end
+
+
+"""
+Download a PDB file or biological assembly from the RCSB PDB server.
```
bicycle1885: The first line should be the signature of the function (see: https://docs.julialang.org/en/stable/manual/documentation).

src/structure/pdb.jl
```diff
@@ -30,26 +49,326 @@ end
...
+            if file_format == PDB
+                download("http://files.rcsb.org/download/$pdbid"*pdbextension[file_format]*"$ba_number.gz",archivefilepath)
+            elseif file_format == mmCIF
+                download("http://files.rcsb.org/download/$pdbid-assembly$ba_number"*pdbextension[file_format]*".gz", archivefilepath)
+            else
+                warn("Biological Assembly is available only in PDB and mmCIF formats!")
+            end
+        end
+        # Verify if the compressed PDB file is downloaded properly and extract it. For MMTF no extraction is needed
+        if ispath(archivefilepath) && filesize(archivefilepath) > 0 && file_format != MMTF
```
bicycle1885: Use `isfile`, not `ispath`.

src/structure/pdb.jl
```diff
@@ -30,26 +49,326 @@ end
...
+By default downloads the PDB files to current working directory; 
+if the keyword argument `pdb_dir` is set the PDB files are downloaded to the
+specified directory;
+if the keyword argument `overwrite` is set `true`, then it will overwrite the
+PDB file if exists in the `pdb_dir`.
+"""
+function downloadentirepdb(;pdb_dir::AbstractString=pwd(), file_format::Type=PDB, overwrite::Bool=false)
+    # Get the list of all pdb entries from RCSB PDB Server using getallpdbentries() and downloads them
+    pdblist = pdbentrylist()
+    info("About to download "*string(length(pdblist))*" PDB files. Make sure to have enough disk space and time!")
```
bicycle1885: Why don't you use string interpolation? `"... download $(length(pdblist)) PDB files..."`

src/structure/pdb.jl
joels94: Wouldn't `try-finally` be sufficient? That way we may be able to know what error occurs. Which might be helpful in the initial stages to debug the code.

joels94:
As discussed I have made the recommended code fixes.  

**TO DO**
- [ ] Update docstrings

- [ ] Write test cases

- [ ] Add documentation

src/structure/pdb.jl
jgreener64: Actually, I just tested that on 0.5 and `eachline` can't take the `chomp` argument. It should be
```julia
for line in eachline(input)
    println(output, chomp(line))
end
```
Sorry about that.

joels94:
I have written the test cases. Kindly take a look at it and let me know if any changes are required.

joels94:
**TO DO**
- [x] ~Update docstrings~

- [x] ~Write test cases~

- [x] ~Add documentation~

Finally, completed writing the documentation. Let me if any changes are required before merging the code.

jgreener64:
On further thought, should the `mmCIF` type be called `MMCIF`? I think the Julia convention of capital types might supersede the technical name, and also this would bring it in line with `MMTF` where the MMs mean the same thing.

joels94:
Updated docs and code as discussed.

docs/src/man/structure.md
```diff
@@ -244,6 +232,210 @@ julia> rad2deg(psiangle(struc['A'][50], struc['A'][51]))
...
+
+- To parse a PDB file by specifying the PDB ID and PDB directory into a Structure-Model-Chain-Residue-Atom framework:
+
+    ```julia
+    struc = readpdb("1EN2", pdb_dir="/path/to/pdb/directory")
+    ```
+
+    **Note:** This requires the PDB file name to be uppercase PDB ID. Example : "1EN2.pdb"
+
+    The function `readpdb` provides an uniform way to download and read PDB files. For example:
```
jgreener64: I'm not sure lines 309-337 are required. Perhaps we could just say "The function `readpdb` provides an uniform way to download and read PDB files, for example `readpdb("1EN2",pdb_dir="/path/to/pdb/directory")`. The same keyword arguments are taken as `read` above, plus `pdb_dir` and `ba_number`."

docs/src/man/structure.md
```diff
@@ -244,6 +232,210 @@ julia> rad2deg(psiangle(struc['A'][50], struc['A'][51]))
...
+
+| Function           | Returns                                                                         | Return type                                              |
+| :----------------- | :------------------------------------------------------------------------------ | :------------------------------------------------------- |
+| `pdbentrylist`     | List of all PDB entries from RCSB Server                                        | `Array{String,1}`                                        |
+| `pdbstatuslist`    | List of PDB entries from specified RCSB weekly status list URL                  | `Array{String,1}`                                        |
+| `pdbrecentchanges` | Added, modified and obsolete PDB lists from the recent RCSB weekly status files | `Tuple{Array{String,1},Array{String,1},Array{String,1}}` |
+| `pdbobsoletelist`  | List of all obsolete PDB entries in the RCSB server                             | `Array{String,1}`                                        |
+
+
+## Maintaining a local copy of the entire RCSB PDB Database
```
jgreener64: This section title is quite long, and the section mainly refers to the above section. Perhaps remove this section and put the sentence "Run the `downloadentirepdb` function..." into the above section.

docs/src/man/structure.md
```diff
@@ -20,13 +20,14 @@ The `Bio.Structure` module provides functionality to manipulate macromolecular s
 To download a PDB file:
```
jgreener64: Perhaps change this section title from "Parsing PDB files" to "Basics" as this makes more sense in the new arrangement.

docs/src/man/structure.md
```diff
@@ -20,13 +20,14 @@ The `Bio.Structure` module provides functionality to manipulate macromolecular s
...
 ```julia
+# Stored in the current working directory by default
 downloadpdb("1EN2")
 ```
 
 To parse a PDB file into a Structure-Model-Chain-Residue-Atom framework:
 
 ```julia
-julia> struc = read(filepath_1EN2, PDB)
+julia> struc = read("/path/to/pdb/file.pdb", PDB)
```
jgreener64: For reference, this was initially set as `filepath_1EN2` to pass the doctests but I don't think they are being tested here so this can stay as `"/path/to/pdb/file.pdb"`.

docs/src/man/structure.md
```diff
@@ -244,6 +232,210 @@ julia> rad2deg(psiangle(struc['A'][50], struc['A'][51]))
...
+    Number of residues          -  85
+    Number of point mutations   -  5
+    Number of other molecules   -  5
+    Number of water molecules   -  76
+    Number of atoms             -  614
+    Number of hydrogens         -  0
+    Number of disordered atoms  -  27
+    ```
+
+    Various options can be set through optional keyword arguments when parsing a PDB file as follows:
```
jgreener64: This applies to other lines too.

joels94:
I have updated the docs. Let me know if any changes are required.

jgreener64:
Great, I'm happy with this to go in. Thanks for all the work. I will wait until tomorrow in case anyone else has any comments, then I'll merge.

PR #484, opened 2017-08-11
Indexed versions of {omega,phi,psi}angle

h3nnn4n:
While using Bio.Structure for a project I found a use case for an indexed version for the `{omega,phi,psi}angle` functions.

### DIFF
--- a/docs/src/man/structure.md
+++ b/docs/src/man/structure.md
@@ -227,6 +227,10 @@ Various functions are provided to calculate spatial quantities for proteins:
 | `rmsd`               | RMSD between two elements of the same size - assumes they are superimposed                      |
 | `displacements`      | `Vector` of displacements between two elements of the same size - assumes they are superimposed |
 
+The omega, phi and psi angle functions can take either a pair of residues or a chain and a position. Omega and phi will
+measure the angle between the residue at the given index and the one before. Psiangle measures between the given index
+and the one after.
+
 For example:
 
 ```julia
@@ -240,6 +244,9 @@ julia> rad2deg(dihedralangle(struc['A'][50]["N"], struc['A'][50]["CA"], struc['A
 -177.38288114072924
 
 julia> rad2deg(psiangle(struc['A'][50], struc['A'][51]))
+-177.38288114072924
+
+julia> rad2deg(psiangle(struc['A'], 50))
 -177.38288114072924
 ```
 
--- a/docs/src/man/structure.md
+++ b/docs/src/man/structure.md
@@ -219,17 +219,20 @@ Various functions are provided to calculate spatial quantities for proteins:
 | `sqdistance`         | Minimum square distance between two elements                                                    |
 | `bondangle`          | Angle between three atoms                                                                       |
 | `dihedralangle`      | Dihedral angle defined by four atoms                                                            |
-| `omegaangle`         | Omega angle between a residue and the previous residue                                          |
-| `phiangle`           | Phi angle between a residue and the previous residue                                            |
-| `psiangle`           | Psi angle between a residue and the next residue                                                |
+| `omegaangle`         | Omega dihedral angle between a residue and the previous residue                                 |
+| `phiangle`           | Phi dihedral angle between a residue and the previous residue                                   |
+| `psiangle`           | Psi dihedral angle between a residue and the next residue                                       |
+| `omegaangles`        | `Vector` of omega dihedral angles of an element                                                 |
+| `phiangles`          | `Vector` of phi dihedral angles of an element                                                   |
+| `psiangles`          | `Vector` of psi dihedral angles of an element                                                   |
 | `ramachandranangles` | `Vector`s of phi and psi angles of an element                                                   |
 | `contactmap`         | Contact map of two elements, or one element with itself                                         |
 | `rmsd`               | RMSD between two elements of the same size - assumes they are superimposed                      |
 | `displacements`      | `Vector` of displacements between two elements of the same size - assumes they are superimposed |
 
-The omega, phi and psi angle functions can take either a pair of residues or a chain and a position. Omega and phi will
-measure the angle between the residue at the given index and the one before. Psiangle measures between the given index
-and the one after.
+The `omegaangle`, `phiangle` and `psiangle` functions can take either a pair of residues or a chain and a position.
+The `omegaangle` and `phiangle` functions measure the angle between the residue at the given index and the one before.
+The `psiangle` function measures between the given index and the one after.
 
 For example:
 
--- a/src/structure/spatial.jl
+++ b/src/structure/spatial.jl
@@ -173,7 +173,7 @@ end
 """
 Calculate the omega angle in radians for an `AbstractResidue`. Arguments can
 either be a residue and the previous residue or a chain and the position as
-an integer" The first residue (or one at the given index) requires the
+a residue ID. The first residue (or one at the given index) requires the
 atoms "N" and "CA" and the previous residue requires the atoms "CA" and "C".
 """
 function omegaangle(res::AbstractResidue, res_prev::AbstractResidue)
@@ -189,23 +189,24 @@ function omegaangle(res::AbstractResidue, res_prev::AbstractResidue)
     return dihedralangle(res_prev["CA"], res_prev["C"], res["N"], res["CA"])
 end
 
-function omegaangle(chain::Chain, res_id::Union{Integer, String})
-    inds = find(r -> resid(r)==string(res_id), chain)
+function omegaangle(chain::Chain, res_id::Union{Integer, AbstractString})
+    inds = find(r -> r == string(res_id), resids(chain))
     if length(inds) != 1
         throw(ArgumentError("$res_id is an invalid residue ID"))
     end
-    res_list = collect(chain)
     i = inds[1]
-    if i == 1 || !sequentialresidues(res_list[i - 1], res_list[i])
+    res = chain[resids(chain)[i]]
+    res_prev = chain[resids(chain)[i-1]]
+    if i == 1 || !sequentialresidues(res_prev, res)
         throw(ArgumentError("$res_id is an invalid residue ID"))
     end
-    return omegaangle(res_list[i], res_list[i - 1])
+    return omegaangle(res, res_prev)
 end
 
 """
 Calculate the phi angle in radians for an `AbstractResidue`. Arguments can
 either be a residue and the previous residue or a chain and the position as
-an integer" The first residue (or one at the given index) requires the
+a residue ID. The first residue (or one at the given index) requires the
 atoms "N", "CA" and "C" and the previous residue requires the atom "C".
 """
 function phiangle(res::AbstractResidue, res_prev::AbstractResidue)
@@ -221,23 +222,24 @@ function phiangle(res::AbstractResidue, res_prev::AbstractResidue)
     return dihedralangle(res_prev["C"], res["N"], res["CA"], res["C"])
 end
 
-function phiangle(chain::Chain, res_id::Union{Integer, String})
-    inds = find(r -> resid(r)==string(res_id), chain)
+function phiangle(chain::Chain, res_id::Union{Integer, AbstractString})
+    inds = find(r -> r == string(res_id), resids(chain))
     if length(inds) != 1
         throw(ArgumentError("$res_id is an invalid residue ID"))
     end
-    res_list = collect(chain)
     i = inds[1]
-    if i == 1 || !sequentialresidues(res_list[i - 1], res_list[i])
+    res = chain[resids(chain)[i]]
+    res_prev = chain[resids(chain)[i-1]]
+    if i == 1 || !sequentialresidues(res_prev, res)
         throw(ArgumentError("$res_id is an invalid residue ID"))
     end
-    return phiangle(res_list[i], res_list[i - 1])
+    return phiangle(res, res_prev)
 end
 
 """
 Calculate the psi angle in radians for an `AbstractResidue`. Arguments can
 either be a residue and the previous residue or a chain and the position as
-an integer" The first residue (or one at the given index) requires the
+a residue ID. The first residue (or one at the given index) requires the
 atoms "N", "CA" and "C" and the next residue requires the atom "N".
 """
 function psiangle(res::AbstractResidue, res_next::AbstractResidue)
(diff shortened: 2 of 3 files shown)


h3nnn4n:
Since the original functions accept the residues in either order I was thinking about adding this possibility for the indexed version. To my eyes there are 3 possibilities:

 - Using a negative index to denote reverse order (seems weird imho)
 - Using a separate function for it (meh)
 - Using a second optional parameter (the best imho)

Comments are much appreciated.

jgreener64:
Thanks for making this PR @h3nnn4n, this is a really nice addition! I'll add some line comments to the code.

Also would be good to add something to the docs at https://github.com/BioJulia/Bio.jl/blob/master/docs/src/man/structure.md, e.g. a line after the spatial table saying something like "`omegaangle`, `phiangle` and `psiangle` can take as input either two residues or a chain and a position as an integer".

jgreener64:
I don't quite understand what you mean by your second comment. The original functions accept two `AbstractResidue`s but they have to be in the correct order - if they are not it will return a value but that won't correspond to a meaningful dihedral.

test/structure/runtests.jl
```diff
@@ -1470,6 +1470,13 @@ end
     @test_throws ArgumentError phiangle(struc_1AKE['A'][7], Residue("ALA", 6, ' ', false, Chain('A')))
     @test_throws ArgumentError psiangle(struc_1AKE['A'][8], Residue("ALA", 9, ' ', false, Chain('A')))
 
+    @test isapprox(omegaangle(struc_1AKE['A'], 19), omegaangle(struc_1AKE['A'][19], struc_1AKE['A'][20]), atol=1e-5)
```
jgreener64: The indices here will change based on comments above.

src/structure/spatial.jl
```diff
@@ -187,6 +187,23 @@ function omegaangle(res::AbstractResidue, res_prev::AbstractResidue)
 end
 
 """
+Calculate the omega angle in radians for a residue in a `Chain`. Arguments are
```
jgreener64: I think the docstrings can go as an extension to the docstrings in the above function. So change the above docstrings to be something like "Arguments can either be a residue and the previous residue or a chain and the position as an integer".

src/structure/spatial.jl
```diff
@@ -187,6 +187,23 @@ function omegaangle(res::AbstractResidue, res_prev::AbstractResidue)
...
+the chain and the first residue index (where atoms "N" and "CA" are taken
+from). The next residue is at pos + 1 (where atom "CA" and "C" iares required).
+"""
+function omegaangle(chain::Chain, pos::Int64)
```
jgreener64: `pos::Integer` will accept all integer types.

src/structure/spatial.jl
```diff
@@ -187,6 +187,23 @@ function omegaangle(res::AbstractResidue, res_prev::AbstractResidue)
...
+    if !(string(pos) in keys(chain.residues))
```
jgreener64: Can use `haskey` and the defined residues accessor: `if !haskey(residues(chain), string(pos))`.

src/structure/spatial.jl
```diff
@@ -187,6 +187,23 @@ function omegaangle(res::AbstractResidue, res_prev::AbstractResidue)
...
+        throw(ArgumentError("$pos is an invalid index"))
+    elseif !(string(pos + 1) in keys(chain.residues))
```
jgreener64: This should be `pos - 1` to preserve directionality. The final call should be `omegaangle(res, res_prev)`. The same is true for `phiangle` (but not `psiangle`).

test/structure/runtests.jl
jgreener64: Otherwise these tests are good.

jgreener64:
There is one more complexity which is that integers are not enough to define the order of residues in a chain. This is due to insertion codes and heteroatoms.

h3nnn4n:
Makes sense. It is not as easy as I thought, hehe. I will see if I can find a simple solution.

jgreener64:
I will have a further look and try and make suggestions over the weekend. The changes you just pushed look good.

jgreener64:
I think you could do something like this:
```julia
function psiangle(chain::Chain, res_id::Union{Integer, String})
    inds = find(r -> resid(r)==string(res_id), chain)
    if length(inds) != 1
        throw(ArgumentError("$res_id is an invalid residue ID"))
    end
    res_list = collect(chain)
    i = inds[1]
    if i == length(res_list) || !sequentialresidues(res_list[i], res_list[i+1])
        throw(ArgumentError("$res_id is an invalid residue ID"))
    end
    return psiangle(res_list[i], res_list[i+1])
end
```
So you find the list index for the residue index given (e.g. 10, 20A or H_30), check one is returned, check the next item is a sequential residue (it may not necessarily be) and if so make the `psiangle` call.

You are right, this is a lot of overhead, but we can see how fast it goes.

Note for omega and phi the second condition changes to `if i == 1 || !sequentialresidues(res_list[i-1], res_list[i])`.

jgreener64:
Yes, `Bio.Strucure` seems to work okay with DNA/RNA though these functions refer to the protein Ramachandran angles.

h3nnn4n:
What I have in mind is more or less this:
```
function psiangle(chain::Chain, pos::Integer)
    res_list = collectresidues(chain, standardselector)
    if pos == length(res_list) || pos < 1 || !sequentialresidues(res_list[pos], res_list[pos+1])
        throw(ArgumentError("$res_id is an invalid residue ID"))
    end
    return psiangle(res_list[pos], res_list[pos+1])
end
```

jgreener64:
Ah interesting, yes I see. I think this would make an inconsistent API though because `chain[6]` would return the residue with resid 6 but `psiangle(chain, 6)` would get the psi angle for the 6th residue in the sequence (resid 11 in your example).

To get the behaviour you want you could use `ramachandranangles`:

```julia
phi, psi = ramachandranangles(chain, standardselector)
psi[1]
```

We could think about a `psiangles` function to just get the psi angles from this.

h3nnn4n:
I implemented your suggestions to the original functions that I submitted.

I also implement the functions to return the array of angles for omega, psi and phi. Those functions are pretty much the same as `ramachandranangles`. They are exported as {phi,psi,omega}angle**s**.

docs/src/man/structure.md
```diff
@@ -227,6 +227,10 @@ Various functions are provided to calculate spatial quantities for proteins:
 | `rmsd`               | RMSD between two elements of the same size - assumes they are superimposed                      |
 | `displacements`      | `Vector` of displacements between two elements of the same size - assumes they are superimposed |
 
+The omega, phi and psi angle functions can take either a pair of residues or a chain and a position. Omega and phi will
```
jgreener64: Can state explicitly the function names e.g. "`omegaangle`, `phiangle` and `psiangle`  can take either..." and "`omegaangle` and `phiangle` will measure...".

jgreener64: Also given that you are saying this here, perhaps change the entries in the table above to be e.g. "Omega dihedral angle" for `omegaangle`.

jgreener64: Can also add `omegaangles` etc. in the table above `ramachandranangles`.

src/structure/spatial.jl
```diff
@@ -167,11 +170,11 @@ function dihedralangle(vec_a::Vector{Float64},
...
 end
 
-
 """
-Calculate the omega angle in radians for an `AbstractResidue`. Arguments are the
-residue (atoms "N" and "CA" required) and the previous residue (atoms "CA" and
-"C" required).
+Calculate the omega angle in radians for an `AbstractResidue`. Arguments can
+either be a residue and the previous residue or a chain and the position as
+an integer" The first residue (or one at the given index) requires the
```
jgreener64: Should be "Arguments can either be a residue and the previous residue or a chain and the position as a residue ID. The first..."

jgreener64: Also for other docstrings.

src/structure/spatial.jl
```diff
@@ -186,10 +189,24 @@ function omegaangle(res::AbstractResidue, res_prev::AbstractResidue)
     return dihedralangle(res_prev["CA"], res_prev["C"], res["N"], res["CA"])
 end
 
+function omegaangle(chain::Chain, res_id::Union{Integer, String})
```
jgreener64: In my example I was too specific, this can be `Union{Integer, AbstractString}`.

src/structure/spatial.jl
```diff
@@ -222,6 +253,103 @@ function psiangle(res::AbstractResidue, res_next::AbstractResidue)
...
+    res_list = collect(chain)
+    i = inds[1]
+    if i == length(res_list) || !sequentialresidues(res_list[i], res_list[i + 1])
+        throw(ArgumentError("$res_id is an invalid residue ID"))
+    end
+    return psiangle(res_list[i], res_list[i + 1])
+end
+
+"""
+Calculate vector of omega, psi or phi angles for a `StructuralElementOrList`.
```
jgreener64: Can copy the docstring from `ramachandranangles` here.

src/structure/spatial.jl
```diff
@@ -222,6 +253,103 @@ function psiangle(res::AbstractResidue, res_next::AbstractResidue)
...
+"""
+function omegaangles(el::StructuralElementOrList)
```
jgreener64: I think to be more flexible the user should be able to pass any selector. So the arguments and `collectresidues` line should be the same as `ramachandranangles`.

src/structure/spatial.jl
```diff
@@ -222,6 +253,103 @@ function psiangle(res::AbstractResidue, res_next::AbstractResidue)
...
+    res_list = collectresidues(el, standardselector)
+    if length(res_list) < 2
+        throw(ArgumentError("At least 2 residues required to calculate dihedral angles"))
+    end
+
+    omega_angles = Float64[]
```
jgreener64: This should be initiated as `Float64[NaN]` or `omegaangles[1]` will refer to the second residue in the list.

src/structure/spatial.jl
jgreener64: And also to `phiangles` and `psiangles`.

src/structure/spatial.jl
```diff
@@ -222,6 +253,103 @@ function psiangle(res::AbstractResidue, res_next::AbstractResidue)
...
+    return omega_angles
+end
+
+function phiangles(el::StructuralElementOrList)
+    res_list = collectresidues(el, standardselector)
+    if length(res_list) < 2
+        throw(ArgumentError("At least 2 residues required to calculate dihedral angles"))
+    end
+
+    phi_angles = Float64[]
```
jgreener64: `Float[NaN]`

src/structure/spatial.jl
```diff
@@ -222,6 +253,103 @@ function psiangle(res::AbstractResidue, res_next::AbstractResidue)
...
+                push!(psi_angles, psi_angle)
+            catch ex
+                isa(ex, ArgumentError) || rethrow()
+                push!(psi_angles, NaN)
+            end
+        else
+            push!(psi_angles, NaN)
+        end
+    end
+
```
jgreener64: Need `push!(psi_angles, NaN)` here to add the value for the last residue.

src/structure/spatial.jl
```diff
@@ -222,6 +253,103 @@ function psiangle(res::AbstractResidue, res_next::AbstractResidue)
...
+    return psi_angles
+end
 
 """
 Calculate the `Vector`s of phi and psi angles of a `StructuralElementOrList`.
```
jgreener64: Now, actually we can replace the `ramachandranangles` function to avoid code duplication. So it can change to:

```julia
function ramachandranangles(el::StructuralElementOrList, residue_selectors::Function...)
    return phiangles(el, residue_selectors...), psiangles(el, residue_selectors...)
end
```

test/structure/runtests.jl
```diff
@@ -1481,6 +1488,15 @@ end
     @test sum(map(x -> Int(isnan(x)), psis)) == 243
     @test_throws ArgumentError ramachandranangles(struc_1AKE['A'][10]["CA"])
 
+    phis = phiangles(struc_1AKE['A'])
+    psis = psiangles(struc_1AKE['A'])
+    omegas = omegaangles(struc_1AKE['A'])
+    @test length(phis) == countresidues(struc_1AKE['A'], standardselector) - 1
```
jgreener64: This test will change given above comments.

src/structure/spatial.jl
```diff
@@ -222,6 +253,103 @@ function psiangle(res::AbstractResidue, res_next::AbstractResidue)
     return dihedralangle(res["N"], res["CA"], res["C"], res_next["N"])
 end
 
+function psiangle(chain::Chain, res_id::Union{Integer, String})
+    inds = find(r -> resid(r)==string(res_id), chain)
```
jgreener64: So I played around with the code here and this is much faster:

```julia
function psiangle(chain::Chain, res_id::Union{Integer, AbstractString})
    inds = find(r -> r==string(res_id), resids(chain))
    if length(inds) != 1
        throw(ArgumentError("$res_id is an invalid residue ID"))
    end
    i = inds[1]
    res = chain[resids(chain)[i]]
    res_next = chain[resids(chain)[i+1]]
    if i == length(chain) || !sequentialresidues(res, res_next)
        throw(ArgumentError("$res_id is an invalid residue ID"))
    end
    return psiangle(res, res_next)
end
```

We use the pre-calculated lists to avoid having to `collect`. Sorry for changing my mind but I think this can just be dropped in (and edited as appropriate for the other two functions).

src/structure/spatial.jl
```diff
@@ -186,10 +189,25 @@ function omegaangle(res::AbstractResidue, res_prev::AbstractResidue)
     return dihedralangle(res_prev["CA"], res_prev["C"], res["N"], res["CA"])
 end
 
+function omegaangle(chain::Chain, res_id::Union{Integer, AbstractString})
+    inds = find(r -> r == string(res_id), resids(chain))
+    if length(inds) != 1
+        throw(ArgumentError("$res_id is an invalid residue ID"))
+    end
+    i = inds[1]
+    res = chain[resids(chain)[i]]
```
jgreener64: I just realised that this needs re-ordering to catch the error properly, at the minute if i is 1 it fails on line 199. Probably easiest to remove lines 198 and 199 and switch lines 200-203 to:
```julia
    if i == 1 || !sequentialresidues(chain[resids(chain)[i-1]], chain[resids(chain)[i]])
        throw(ArgumentError("$res_id is an invalid residue ID"))
    end
    return omegaangle(chain[resids(chain)[i]], chain[resids(chain)[i-1]])
```
and similar for the other 2 functions.

jgreener64:
I'll give a day or so more for comments and then merge this.

jgreener64:
Thanks for all the work @h3nnn4n .

PR #485, opened 2017-08-17
Decompose Bio.Structure

jgreener64:
Once the BioStructures.jl package (https://github.com/BioJulia/BioStructures.jl) is tagged and released this removes the code from Bio.jl.


Ward9250:
This looks ok to me at first glance, but we should perhaps wait for BioStructures release, so as then the CI can actually run with the BioStructures dependency available, just to be safe.

jgreener64:
Yes agreed.

jgreener64:
So having re-triggered the build it fails because there are no tagged versions of BioStructures.jl for Julia v0.5. The first release is tagged for v0.6 minimum.

So I guess this PR should wait until Bio.jl is ready to transfer to Julia v0.6 generally?

Ward9250:
Well since all the old syntax actually works with 0.6 - just a few warnings here and there, we can try and just bump up the CI to 0.6, change the bare minimum syntax, and continue with the decompose :+1:

jgreener64:
Sounds good. I see you've started this process, thanks.

Ward9250:
Ok I've resolved the merge commits, now we can see how the tests go this time, there should be no issues.

Ward9250:
Ok this can merge now, thanks for all the hard work @jgreener64, fantastic stuff!

jgreener64:
Great, thanks for the merge @Ward9250 .

PR #486, opened 2017-08-23
Remove PDB update test that may fail

jgreener64:
This test of the PDB can actually fail if no entries were obsoleted recently.


jgreener64:
Will check this builds okay, then merge.

jgreener64:
Merging this as it fixes the build.

PR #487, opened 2017-09-01
Update the CI

Ward9250:
Don't test on julia 0.5 anymore on master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,14 +3,9 @@ os:
   - linux
   - osx
 julia:
-  - 0.5
-  - nightly
-matrix:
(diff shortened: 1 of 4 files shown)


NEWS.md
```diff
@@ -1,46 +1,2 @@
-Bio.jl v0.5.0 Release Notes
+Bio.jl v0.6.0 Release Notes
```
jgreener64: Shouldn't this be v0.5.0 as it refers to Bio.jl not Julia?

Ward9250: Oh yes, that was accidental!

